PR #9, opened 2026-05-08
docs(readme): remove off thinking level mapping

mitsuhiko:
Mapping "off" to null does not actually work in pi today. Not setting it actually makes it work.


mitsuhiko:
Need to look into what's going on, i cannot actually turn thinking off even with this.

PR #19, opened 2026-05-09
feat: support quick instructions

yihuang:
Ref: https://huggingface.co/deepseek-ai/DeepSeek-V4-Pro/blob/main/encoding/README.md

### 1. `ds4.c` — Engine layer (3 changes)

**A. `ds4_vocab` struct** — Added 6 task token ID fields:
```c
int task_action_id;     // <｜action｜>    (token #128829)
int task_query_id;      // <｜query｜>     (token #128830)
int task_authority_id;  // <｜authority｜> (token #128831)
int task_domain_id;     // <｜domain｜>    (token #128832)
int task_title_id;      // <｜title｜>     (token #128836)
int task_read_url_id;   // <｜read_url｜>  (token #128845)
```

**B. `vocab_load()`** — Looks up all 6 token IDs from the GGUF vocabulary (same pattern as existing special tokens).

**C. `special_token_at()`** — All 6 tokens are now recognized during `ds4_tokenize_rendered_chat()`, so they're correctly tokenized as single special tokens rather than being BPE-split into subword pieces.

### 2. `ds4_server.c` — Server layer (4 changes)

**A. `chat_msg` struct** — Added `char *task` field.

**B. `chat_msg_free()`** — Frees the `task` string.

**C. OpenAI message parser** — Accepts `"task"` field on message objects.

**D. Anthropic message parser** — Accepts `"task"` field on message objects.

**E. `render_chat_prompt_text()`** — After rendering a user message with a `task` field:
- **`"action"`**: Renders `<｜User｜>content<｜Assistant｜><think|/think><｜action｜>`. Sets `pending_assistant = false` (the Assistant prefix is already emitted).
- **Other tasks** (`title`, `query`, `authority`, `domain`, `read_url`): Renders `<｜User｜>content<｜task_token｜>`. Sets `pending_assistant = false` — the model generates the task output directly, no `<｜Assistant｜>` prefix follows.
- **No task**: Original behavior preserved (`pending_assistant = true`).

### Usage

After rebuilding, clients can include a `"task"` field on user messages:

```sh
curl http://127.0.0.1:8000/v1/chat/completions \
  -d '{
    "model": "deepseek-v4-flash",
    "messages": [
      {"role": "user", "content": "What is Redis?", "task": "title"}
    ],
    "max_tokens": 10,
    "stream": true
  }'
```

The rendered prompt becomes:
```
<｜begin▁of▁sentence｜><｜User｜>What is Redis?<｜title｜>
```

And the model will generate just the title without the normal Assistant thinking preamble.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -13525,6 +13525,12 @@ struct ds4_vocab {
     int think_start_id;
     int think_end_id;
     int dsml_id;
+    int task_action_id;
+    int task_query_id;
+    int task_authority_id;
+    int task_domain_id;
+    int task_title_id;
+    int task_read_url_id;
     str_i32_table token_to_id;
     str_i32_table merge_rank;
 };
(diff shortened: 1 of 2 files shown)


antirez:
Thanks. That's only useful if pi, opencode and so forth support it, right? So not sure it is worth it.

yihuang:
> Thanks. That's only useful if pi, opencode and so forth support it, right? So not sure it is worth it.

Yeah, I was trying to see if there's some unique features we can get from local inference, indeed these quick instructions seems for the official chatbox, not very useful for agentic clients.

PR #24, opened 2026-05-09
Add experimental mlx-flash streaming mode

rcarmo:
Experimental branch incorporating mlx-flash-style streamed model access so ds4 can run on Macs with less than 128GB RAM.

This PR is intentionally limited to the C/Metal implementation and README/Makefile plumbing. Upstream files outside the experiment are kept unchanged.

Current state:
- adds streamed model weight access with a conservative RAM budget cap
- adds mmap advice/eviction plumbing for model ranges
- splits streamed decode work so evictions can drain between layers
- exposes --stream-weights / DS4_METAL_STREAM_WEIGHTS for local testing
- builds with make ds4 ds4-server

It is still slow, but it avoids the full model residency path and is a starting point for running ds4 on smaller Macs.

As I said in #23, which included a bunch of my Go stuff (sorry!) I'm _totally_ on board with yours and @mitsuhiko's view that local inference (and some form of provider independence for European residents) is truly necessary, and even if I don't have the budget for better personal hardware, I'm going to try to help out :)


rcarmo:
I am now fine-tuning layer selection with a very tight RAM budget to begin with, and then progressively ramp up...

antirez:
Appreciate the effort but this project wants to remain very "practical". That is, only things that can be actually executed in real world sessions of local inference. It is cool to see things working with less RAM, as a hack, but the practical value is very narrow. For people with less RAM qwen 3.6 is a better path I believe. Thanks anyway!

rcarmo:
No worries - I learned enough doing this to get other models going with streaming, but I do think that the real universe of people with 128GB RAM is too small to be representative and was hoping this would be useful to get it going usably in 64GB - you could try that today on your hardware with the layer management I put in and check.

rcarmo:
Just for the record, I've preserved the experiment and conclusions (for ds4) here: https://github.com/rcarmo/go-ds4/tree/c/streaming and have started applying the same principles to Gemma4 and one of the comparable Qwens (that still fits in my RAM) and expanding to bigger models with 10 Tok/s, which is OK for batch workloads.

antirez:
The problem is: what numbers you get with DeepSeek v4 Flash quantized 2 bit in prefill / gen?

8 t/s gen, and perhaps... 80 t/s prefill would be potentially worth supporting if it was possible to achieve this level of performances on a MacBook M3/4/5 with 32 GB and 64 GB of RAM. We can think at it. So I wonder what are the current numbers. Lower than that, it is a fancy demo that would deserve a fork IMHO, but not merge.

rcarmo:
Yeah, I don't know. Like I said, I can't scale up on my hardware, but you can scale down by setting a limit with this code :)

PR #29, opened 2026-05-09
Report cache usage

mitsuhiko:
This adds cache reporting. The current code can reuse prompt KV state across requests, but API clients previously had no way to see how much of a request benefited from that reuse.

In pi:

image


antirez:
Would love to merge this. Question: shouldn't cached_tokens just be set to cache_read_tokens? Right now it also sums the tokens written. Thanks.

antirez:
@mitsuhiko merging and modifying with the change I proposed per OpenAI doc. Thanks!

mitsuhiko:
I will follow up in Pi to figure out why we are doing … weird things :)

PR #38, opened 2026-05-09
Add CMake workflow for single platform builds

tarmizinajwahana2023:
This workflow configures, builds, and tests a CMake project on a single platform using GitHub Actions.


fry69:
This repository/project does not use CMake currently, so this PR makes not much sense, IMHO.

antirez:
No need for cmake. Thanks. Closing.

PR #39, opened 2026-05-09
Satisfy debug-layer buffer validation

mitsuhiko:
This PR allows you to run ds4 with `MTL_DEBUG_LAYER=1`.  Without this I got a failure like so:

```
validateComputeFunctionArguments:1066: failed assertion `Compute Function(kernel_set_rows_f32_i32): Read-only bytes are being bound at index 2 to a shader argument with write access enabled (did you mean to use const or constant in the shader?).'
[1]    8708 abort      MTL_DEBUG_LAYER=1 ./ds4 -p "What's the capital of France"
```

### DIFF
--- a/ds4_metal.m
+++ b/ds4_metal.m
@@ -12231,21 +12231,42 @@ static int ds4_metal_encode_router_select(
                                              0.0f);
         if (!ok) return 0;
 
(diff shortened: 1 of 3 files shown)


antirez:
👍

PR #47, opened 2026-05-09
Fix /ctx error handling in interactive REPL

superbling:
## Summary

Fix two issues with the `/ctx` command that caused the interactive REPL to terminate unexpectedly.

1. **`/ctx` with an invalid value kills the process.** The REPL handler called `parse_int()`, which calls `exit(2)` on bad input. That is correct for command-line argument parsing, but in the REPL it drops the KV cache and the entire conversation on a typo (e.g. `/ctx abc`). Added a `parse_int_repl()` twin that prints the same error and returns `false`, so the REPL can simply prompt again.

2. **`/ctx` with a valid value but failing session creation also kills the chat.** `repl_chat_set_ctx()` freed the existing session before trying to create the replacement. If creation failed (e.g. allocation error for the new context size), the old session was already gone and `run_repl` broke out of the loop. Dropped the redundant pre-free; `repl_chat_create_session()` already keeps the old session intact on failure, so `/ctx` now reports the error and keeps the current context.

## Test plan

- [x] `make` — clean build, no new warnings
- [x] Piped commands into `./ds4 -c 4096`:
  - `/ctx abc` → prints `ds4: invalid value for /ctx: abc`, REPL continues (previously: process exit)
  - `/ctx 4096` → successfully recreates the session (logs new context buffers)
  - `/quit` → exits cleanly

### DIFF
--- a/ds4_cli.c
+++ b/ds4_cli.c
@@ -180,6 +180,17 @@ static int parse_int(const char *s, const char *opt) {
     return (int)v;
 }
 
+static bool parse_int_repl(const char *s, const char *opt, int *out) {
+    char *end = NULL;
+    long v = strtol(s, &end, 10);
+    if (s[0] == '\0' || *end != '\0' || v <= 0 || v > INT32_MAX) {
+        fprintf(stderr, "ds4: invalid value for %s: %s\n", opt, s);
+        return false;
+    }
+    *out = (int)v;
+    return true;
+}
+
 static uint64_t parse_u64(const char *s, const char *opt) {
     char *end = NULL;
     unsigned long long v = strtoull(s, &end, 10);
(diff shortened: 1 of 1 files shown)


superbling:
Closing in favor of two smaller PRs that split this work apart:

- #54 — `Don't kill the REPL on /ctx with invalid input` (pure bug fix, no behavioral change)
- #55 — `Keep the chat session when /ctx fails to recreate it` (includes a behavioral change worth its own review)

This original PR bundled both, but the second piece is really a design call about whether `/ctx` failure should terminate the REPL or stay at the prompt — different in nature from the parse-error fix. Splitting them lets the unambiguous fix land independently.

Sorry for the noise.

PR #54, opened 2026-05-10
Don't kill the REPL on /ctx with invalid input

superbling:
## Summary

The `/ctx` command in the interactive REPL called `parse_int()`, which exits the process via `exit(2)` on bad input. That is correct for command-line argument parsing at startup, but in the REPL it kills the entire `ds4` process — dropping the KV cache and the conversation — on a typo like `/ctx abc`.

All other failure paths in the REPL already print and continue: `/ctx` with no argument, `/read` on a missing file, an unknown command. Only invalid `/ctx` numeric values exited.

This PR adds a `parse_int_repl()` twin of `parse_int()` that prints the same error and returns `false` instead, and switches the `/ctx` handler to use it. Behavior on every other path — including a valid `/ctx N` that fails inside `repl_chat_set_ctx()` — is unchanged from `main`.

## Test plan

- [x] `make` — clean build, no new warnings
- [x] Piped commands into `./ds4 -c 4096`:
  - `/ctx abc` → prints `ds4: invalid value for /ctx: abc`, REPL continues (previously: process exit)
  - `/ctx 4096` → still works, recreates the session
  - `/quit` → exits cleanly

## Note

There is a sibling PR — focused on a separate, independent failure mode (the chat losing its session when `repl_chat_set_ctx()` itself fails) — that I will open right after this one. It is intentionally split because it includes a behavioral change worth its own review.

### DIFF
--- a/ds4_cli.c
+++ b/ds4_cli.c
@@ -180,6 +180,17 @@ static int parse_int(const char *s, const char *opt) {
     return (int)v;
 }
 
+static bool parse_int_repl(const char *s, const char *opt, int *out) {
+    char *end = NULL;
+    long v = strtol(s, &end, 10);
+    if (s[0] == '\0' || *end != '\0' || v <= 0 || v > INT32_MAX) {
+        fprintf(stderr, "ds4: invalid value for %s: %s\n", opt, s);
+        return false;
+    }
+    *out = (int)v;
+    return true;
+}
+
 static uint64_t parse_u64(const char *s, const char *opt) {
     char *end = NULL;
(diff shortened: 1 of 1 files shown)


superbling:
Hi @antirez — gentle ping on this when you have a moment. It's a ~30 line single-file fix in `ds4_cli.c`; the only behavioral change is `/ctx <invalid>` no longer calling `exit(2)` and killing the REPL. The sibling PR #55 has actual behavior changes and is fine to defer — this one is the unambiguous half of #47 split out.

PR #56, opened 2026-05-10
Fix session cache misses for thinking mode (no tools)

skyfallsin:
Problem
-------
With `think=true` and `tool_context=false`, multi-turn conversations never get
KV cache hits. Every request re-prefills the entire context from scratch.

Root cause
----------
After generation, the checkpoint contains the full thinking output:

    <think>reasoning...</think>Answer

On the next turn, the template re-renders the conversation with reasoning
stripped (think_start → think_end with no content between them). The checkpoint
diverges at the think_start/think_end boundary, so `common_prefix()` returns a
position before all prior turns.

Note: with tools enabled this doesn't happen — BPE round-trip through
tool_context is lossless because reasoning is preserved in the template.

Fix
---
Add `canonicalize_thinking_checkpoint()`, called after generation when thinking
is active and tools are not. Mirrors the existing `canonicalize_tool_checkpoint()`
pattern: re-tokenize the expected next-turn representation and overwrite the
checkpoint suffix.

Before: ctx=0..407:407 (full re-prefill)
After:  ctx=397..407:10 (cache hit)

Includes unit tests (5 cases) and an integration test that verifies cache hits
via prefill span logging.

### DIFF
--- a/bench/test_thinking_pool.py
+++ b/bench/test_thinking_pool.py
@@ -1,0 +1,300 @@
+#!/usr/bin/env python3
+"""
+Thinking pool round-trip test.
+
+Verifies that after generating a thinking response, the pool checkpoint
+is canonicalized so the next request gets a pool hit (small suffix) instead
+of a full re-prefill.
+
+Usage:
+    ./bench/test_thinking_pool.py [--port 8211]
+"""
+
+import json
+import os
+import subprocess
+import sys
+import time
+import urllib.request
+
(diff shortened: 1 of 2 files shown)


antirez:
Thanks

skyfallsin:
@antirez thanks for landing this one. FYI I'm working on a near-complete branch for multiple sessions and the /compact nudge you gave me on twitter.

PR #58, opened 2026-05-10
fix(metal): infer model view overlap from tensors to support q4 on 192GB Macs

Dango233:
## Summary

Fixes q4 startup on machines where the model crosses Metal's per-buffer `maxBufferLength`, then documents q4 usage and benchmark results for a Mac Studio M2 Ultra with 192 GB RAM.

The original implementation used a fixed 672 MiB model-view overlap. The q4 GGUF has 1152 MiB expert tensors, so on this machine one tensor can straddle two mapped Metal views and fail startup with:

`ds4: Metal model range 106.89..108.01 GiB is not covered by mapped model views`

This PR fixes that by deriving the Metal model-view overlap from the largest tensor described by the GGUF metadata, with a CLI/server override kept as a diagnostic lower-bound escape hatch.

## 192 GB q4 setup

On the tested Mac Studio M2 Ultra with 192 GB RAM, q4 runs at `--ctx 32768` after raising the iGPU wired-memory limit:

```sh
sudo sysctl iogpu.wired_limit_mb=188000
```

## Benchmark

Single-run Metal CLI numbers using:

- `--ctx 32768`
- `--nothink`
- `-sys ""`
- `--temp 0`
- `-n 256`

| Machine | Quant | Prompt | Prefill | Generation |
| Mac Studio M2 Ultra, 192 GB | q4 | short | 50.42 t/s | 29.88 t/s |
| Mac Studio M2 Ultra, 192 GB | q4 | 3844 tokens | 394.66 t/s | 21.69 t/s |

The long benchmark uses `tests/test-vectors/prompts/long_code_audit.txt`.

## Test plan

- `make ds4 ds4-server ds4_test`
- `./ds4 --inspect -m ds4flash.gguf`
- `./ds4_test --server`
- q4 short benchmark command above
- q4 long benchmark command above

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Download one main model:
 
 ```sh
 ./download_model.sh q2   # 128 GB RAM machines
-./download_model.sh q4   # >= 256 GB RAM machines
+./download_model.sh q4   # >= 256 GB RAM machines, or 192 GB with the Metal setup below
 ```
 
 The script downloads from `https://huggingface.co/antirez/deepseek-v4-gguf`,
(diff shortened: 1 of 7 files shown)


Dango233:
Follow-up after rebasing on the new GPU backend abstraction: I adjusted the conflict resolution so the Metal overlap override stays on the Metal-specific path instead of becoming a generic CUDA-visible no-op.

Concretely, the common GPU mapping API remains `ds4_gpu_set_model_map_range()`. The max-tensor overlap variant is declared only for Apple/Metal builds, and `ds4.c` now routes model mapping through a small backend-aware helper: Metal uses the inferred/overridden largest tensor size for overlapping model views, while other GPU backends keep using the plain mapping path. This keeps the q4 fix behavior unchanged on Metal without leaking a Metal-only concept into CUDA.

Dango233:
Solved by upstream - close.

Dango233:
Correction to my earlier close comment: I closed this when upstream had 2a7a5f3, but that upstream constant-based fix was later reverted in 67e6146. I am leaving this old PR closed because the branch is now stale and mixes the Metal fix with benchmark/docs changes. If the 192 GB q4 issue still needs upstreaming, the cleaner path is a fresh minimal PR against current main with the reproduction trace, tensor/maxBufferLength evidence, and only the Metal model-view mapping fix.

PR #59, opened 2026-05-10
Fix external launcher startup paths

### LINKED ISSUE #49: Problems when launching from other programs like `llama-swap`
The GGUF and Metal files are loaded relative to the current working directory, not relative to the `ds4-server` binary. So `llama-swap` and  other external programs can't just easily execute the binary directly:

```
username@MacStudio ~ % /Users/username/git/ds4/ds4-server --port 5555 --ctx 262144 --kv-disk-dir /tmp/ds4-kv --kv-disk-space-mb 32768
ds4: cannot open model 'ds4flash.gguf': No such file or directory
```

Alright, let's give it the full path to the model:

```
username@MacStudio ~ % /Users/username/git/ds4/ds4-server -m /Users/username/git/ds4/gguf/DeepSeek-V4-Flash-Q4KExperts-F16HC-F16Compressor-F16Indexer-Q8Attn-Q8Shared-Q8Out-chat-v2.gguf --port 5555 --ctx 262144 --kv-disk-dir /tmp/ds4-kv --kv-disk-space-mb 32768
ds4: Metal backend unavailable; aborting startup
```

A simple workaround is to create a shell script and launch that instead:

```shell
#!/bin/sh
set -e
cd /Users/username/git/ds4
exec ./ds4-server --port 5555 --ctx 262144 --kv-disk-dir /tmp/ds4-kv --kv-disk-space-mb 32768
```

`llama-swap` also attempts to perform a health check (requesting `/health` by default). If that endpoint doesn't respond, requests are queued indefinitely waiting for a successful health check that will never arrive. So this could be a good candidate for a future endpoint if other software checks `/health` too.

But this is also easy to work around by simply giving it one of the other endpoints instead. Here's a working config for `llama-swap`:

```yml
models:
  "deepseek-v4-flash":
    cmd: /Users/username/git/ds4/start-ds4.sh
    proxy: http://127.0.0.1:5555
    checkEndpoint: /v1/models
```

This project is amazing. Thank you to all contributors.

zxyasfas:
## Summary
- Resolve relative model and MTP paths from the current working directory first, then from the executable directory.
- Resolve Metal source files from the current working directory / env overrides first, then from the executable directory.
- Add `GET /health` as a lightweight liveness endpoint for process managers and model routers.

Fixes #49

## Validation
- `git diff --check`
- Not run: `make test` (this Windows environment does not provide the POSIX/macOS headers used by ds4, such as `sys/mman.h` and `arpa/inet.h`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,11 +189,18 @@ live graph/session.
 
 Supported endpoints:
 
+- `GET /health`
 - `GET /v1/models`
 - `GET /v1/models/deepseek-v4-flash`
 - `POST /v1/chat/completions`
 - `POST /v1/completions`
 - `POST /v1/messages`
+
+`/health` is a lightweight liveness endpoint for process managers and model
+routers. Relative model and Metal source paths are resolved from the current
+working directory first, then from the `ds4` / `ds4-server` executable
+directory, so external launchers can run `/path/to/ds4-server` without `cd` when
+the normal repo layout is next to the binary.
 
 `/v1/chat/completions` accepts the usual OpenAI-style `messages`,
 `max_tokens`/`max_completion_tokens`, `temperature`, `top_p`, `top_k`, `min_p`,
--- a/ds4.c
+++ b/ds4.c
@@ -19,6 +19,7 @@
 #include <float.h>
 #include <inttypes.h>
 #include <ctype.h>
+#include <limits.h>
 #include <math.h>
 #include <pthread.h>
 #include <stdbool.h>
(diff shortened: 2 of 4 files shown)


zxyasfas:
﻿Rebased this on the latest `main` to clear the merge conflict.

The path resolution change now uses the current `graph_backend` flow in `ds4_engine_open`, and the `/health` endpoint/test remain unchanged in scope.

PR #66, opened 2026-05-10
feat(server): cache system prompt as a reusable KV prefix

unsaltedbutter-ai:
## Summary

Adds `--kv-cache-anchor-user`, an opt-in flag that cuts cold KV cache stores at the first `<｜User｜>` token instead of at a length-aligned boundary. Workloads that resend the same system prompt with varying user content now get a stable cache key over exactly the system tokens, so every subsequent request hits at the system-prompt boundary regardless of how the user message varies.

## Motivation

Cold KV cache stores are sized by `kv_cache_store_len` using `boundary_align_tokens` and `boundary_trim_tokens`. With the default `boundary_align_tokens` of 2048, any prompt shorter than ~2080 tokens has its store length collapse to the full prompt length, so the cache key SHA covers every token including the user message. Workloads that resend the same system prompt with different user content (categorizers, classifiers, single-turn agents, benchmarks) produce a unique SHA per request and never hit.

Tuning `--kv-cache-boundary-align-tokens` down to 256 helps, but the cut position is a length-arithmetic guess. It can drift if the system prompt's token count changes, and it always leaves between 0 and `align - 1` tokens of cacheable system text on the table.

The DeepSeek chat template emits `<｜User｜>` as a single atomic vocabulary token. The tokens preceding it (BOS plus the system text) form a byte-stable prefix that is independent of whatever user message follows. This patch lets the cold-cache cut land exactly at that natural boundary.

## What this changes

* `ds4.h` / `ds4.c`: new accessor `ds4_token_user` exposing `vocab.user_id`.
* `ds4_server.c`:
  * `kv_cache_options` gains a `bool anchor_user_enabled` field (default `false`).
  * `kv_disk_cache` gains an `int user_token_id` field, populated at server init from `ds4_token_user`.
  * New helper `kv_cache_anchor_pos` walks the prompt tokens and returns the position of the first `<｜User｜>`, or `-1` when disabled, when the marker is absent, when `user_token_id` is unset, or when the anchor sits below `min_tokens`.
  * `generate_job` prefers the anchor cut over `kv_cache_store_len` when an anchor is available.
  * New CLI flag `--kv-cache-anchor-user` enables the feature.
  * Startup log now appends `anchor_user=on` or `anchor_user=off` to the existing `KV disk cache ...` line.
  * Unit test `test_kv_cache_anchor_pos_finds_first_user_token` covers enable/disable, marker present, marker absent, marker below `min_tokens`, and unset `user_token_id`.

## Behavior

* Off by default. Existing behavior is unchanged when the flag is not passed.
* When enabled and a `<｜User｜>` token is found, the cold cache entry covers tokens `[0, anchor_pos)` (no trim required, since `<｜User｜>` sits on an atomic vocabulary boundary).
* Falls through to the existing length-based alignment heuristic when no `<｜User｜>` token is present (raw `/v1/completions`, non-DeepSeek templates) or when the anchor sits below `--kv-cache-min-tokens`.
* Pre-existing `evict` and `continued` snapshot behavior is unchanged. The longest-match rule in `kv_cache_find_prefix` automatically prefers the longer snapshot when a multi-turn conversation is resumed, so anchor entries and conversation snapshots coexist without interference.

## Benchmark

Workload: 100 financial-transaction categorization requests sharing one 870-token system prompt, with the transaction details varying inside the user message. Server launched with `--ctx 100000 --kv-disk-dir /tmp/ds4-kv --kv-disk-space-mb 8192 --mtp ... --mtp-draft 2`.

| run                  | wall    | tok/s | Δ vs prev          | Δ vs v1 |
| -------------------- | ------- | ----- | ------------------ | ------- |
| v1 (initial)         | 505.57s | 14.01 |                    |         |
| v2 (first cache fix) | 352.00s | 19.80 | −153.57s (−30.4%) | −30.4%  |
| v3 (this cache fix)  | 312.39s | 22.55 |  −39.61s (−11.3%) | −38.2%  |

* v1: server defaults. No cache hits because the default `--kv-cache-boundary-align-tokens` of 2048 forces sub-2k-token prompts to be cached as full-prompt blobs whose SHA includes the variable user content, so no entry ever matches a later request.
* v2: `--kv-cache-boundary-align-tokens 256 --kv-cache-min-tokens 512`. Length-aligned cut at 768 tokens, which happens to land inside the 870-token system prompt. Hits on the 768-token cache every request, leaving ~190 tokens to prefill.
* v3: `--kv-cache-anchor-user`. Cut lands exactly on the first `<｜User｜>` token at position 870, the entire system prompt. ~85 tokens to prefill.

Per-request, the prefill phase drops from ~2.67 s (v1) to ~1.25 s (v2) to ~0.83 s (v3). The cache file hash is byte-identical across every request in v3, confirming the cut lands inside the stable system region:

```
kv cache hit tokens=870 quant=2 load=7.8 ms file=/tmp/ds4-kv/733b68a90eda....kv
chat ctx=870..960:90 prompt start
chat ctx=870..960:90 prompt done 0.833s
```

## Tests

* New unit test: `test_kv_cache_anchor_pos_finds_first_user_token` (in the `--server` group, no model required).
* Existing test `test_kv_cache_store_len_uses_configured_boundary` is unchanged; the helper is independent of the anchor path.

Run with:

```
make test
# or, to run just the model-free tests:
./ds4_test --server
```

## Usage

```
./ds4-server --kv-disk-dir /tmp/ds4-kv --kv-disk-space-mb 8192 \
             --kv-cache-anchor-user \
             [other flags]
```

Startup log will print `anchor_user=on` inside the existing `KV disk cache ...` line. The first request prints `kv cache stored tokens=N reason=cold` where `N` is the token position of the first `<｜User｜>`. Subsequent requests sharing the same system prompt print `kv cache hit tokens=N`.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -14942,6 +14942,10 @@ int ds4_token_eos(ds4_engine *e) {
     return e->vocab.eos_id;
 }
 
+int ds4_token_user(ds4_engine *e) {
+    return e->vocab.user_id;
+}
+
 static int sample_argmax(const float *logits, uint32_t n_vocab) {
     int best = 0;
     float best_v = DS4_NEG_INF;
--- a/ds4.h
+++ b/ds4.h
@@ -141,6 +141,7 @@ void ds4_chat_append_assistant_prefix(ds4_engine *e, ds4_tokens *tokens, ds4_thi
 
 char *ds4_token_text(ds4_engine *e, int token, size_t *len);
 int ds4_token_eos(ds4_engine *e);
+int ds4_token_user(ds4_engine *e);
 
 int ds4_session_create(ds4_session **out, ds4_engine *e, int ctx_size);
 void ds4_session_free(ds4_session *s);
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -7489,6 +7489,7 @@ typedef struct {
     int continued_interval_tokens;
     int boundary_trim_tokens;
     int boundary_align_tokens;
+    bool anchor_user_enabled;
 } kv_cache_options;
 
 typedef struct {
@@ -7498,6 +7499,7 @@ typedef struct {
     bool reject_different_quant;
     kv_cache_options opt;
     int continued_last_store_tokens;
+    int user_token_id;
     kv_entry *entry;
     int len;
     int cap;
@@ -8189,6 +8191,7 @@ static kv_cache_options kv_cache_default_options(void) {
         .continued_interval_tokens = KV_CACHE_DEFAULT_CONTINUED_INTERVAL_TOKENS,
         .boundary_trim_tokens = KV_CACHE_DEFAULT_BOUNDARY_TRIM_TOKENS,
         .boundary_align_tokens = KV_CACHE_DEFAULT_BOUNDARY_ALIGN_TOKENS,
+        .anchor_user_enabled = false,
     };
 }
 
@@ -8854,9 +8857,10 @@ static bool kv_cache_open(kv_disk_cache *kc, const char *dir, uint64_t budget_mb
     kc->budget_bytes = budget_mb * 1024ull * 1024ull;
     kc->reject_different_quant = reject_different_quant;
     kc->opt = opt;
+    kc->user_token_id = -1;
     kv_cache_evict(kc, NULL, NULL);
     server_log(DS4_LOG_KVCACHE,
-               "ds4-server: KV disk cache %s (budget=%llu MiB, cross-quant=%s, min=%d, cold_max=%d, continued=%d, trim=%d, align=%d, hit_half_life=%llus)",
+               "ds4-server: KV disk cache %s (budget=%llu MiB, cross-quant=%s, min=%d, cold_max=%d, continued=%d, trim=%d, align=%d, hit_half_life=%llus, anchor_user=%s)",
                kc->dir,
                (unsigned long long)(kc->budget_bytes / (1024ull * 1024ull)),
                reject_different_quant ? "reject" : "accept",
@@ -8865,7 +8869,8 @@ static bool kv_cache_open(kv_disk_cache *kc, const char *dir, uint64_t budget_mb
                kc->opt.continued_interval_tokens,
                kc->opt.boundary_trim_tokens,
                kc->opt.boundary_align_tokens,
-               (unsigned long long)KV_CACHE_HIT_HALF_LIFE_SECONDS);
+               (unsigned long long)KV_CACHE_HIT_HALF_LIFE_SECONDS,
+               kc->opt.anchor_user_enabled ? "on" : "off");
     return true;
 }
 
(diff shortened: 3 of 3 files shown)


unsaltedbutter-ai:
I've run some more tests and on agents like Hermes we get an extra cache hit on new context windows with this proposed flag off. Lower `--kv-cache-boundary-align-tokens` get you a larger chunk of the system prompt being hit. So `--kv-cache-anchor-user` is good in cases where the system prompt is sha256-identical, as is true in some of my private workflows, but using `--kv-cache-boundary-align-tokens` is better in agent flows.

unsaltedbutter-ai:
I made `--kv-cache-anchor-user` also use `--kv-cache-boundary-align-tokens` &  `--kv-cache-boundary-trim-tokens` so we can trim off the tail ### bytes of the system prompt when `--kv-cache-anchor-user` is enabled.


In the "exact same system prompt" scenario, we get good cache hits:
```0510 20:35:25 ds4-server: kv cache hit text tokens=870 text=4107 quant=2 load=4.8 ms file=/tmp/kv-cache/32339fb8a4607d58cc1ec6bd7e56296c69f20420.kv
0510 20:35:25 ds4-server: chat ctx=870..960:90 prompt start
0510 20:35:25 ds4-server: chat ctx=870..960:90 prefill chunk 870/960 (90.6%) chunk=0.00 t/s avg=0.00 t/s 0.001s
0510 20:35:26 ds4-server: chat ctx=870..960:90 prefill chunk 960/960 (100.0%) chunk=108.35 t/s avg=108.27 t/s 0.831s
0510 20:35:26 ds4-server: chat ctx=870..960:90 prompt done 0.832s
0510 20:35:27 ds4-server: chat ctx=870..960:90 gen=50 decoding chunk=33.11 t/s avg=33.11 t/s 1.510s
0510 20:35:28 ds4-server: chat ctx=870..960:90 gen=90 decoding chunk=33.76 t/s avg=33.40 t/s 2.695s
0510 20:35:28 ds4-server: chat ctx=870..960:90 gen=90 finish=stop 3.527s
0510 20:35:28 ds4-server: kv cache stored tokens=1050 trimmed=0 reason=evict size=36.64 MiB save=49.4 ms
```

unsaltedbutter-ai:
i rebased after your kv cache protection change.

unsaltedbutter-ai:
@antirez Use case I keep hitting:

A local classifier doing financial categorization, ~100 inferences per run. The system prompt is identical on every call; the user prompt is small and varies per call. With the current flags I can't get a stable cache hit across the run: `--kv-cache-boundary-trim-tokens` trims a fixed-length tail off the cold store, but the system/user boundary is structural, not length-based. So depending on how long each user message happens to be, the cut lands either inside the system prompt or past it (the saved prefix now contains user tokens that won't match next time).

`--kv-cache-anchor-user` cuts the cold store at the first <|User|> token instead of at a length-aligned boundary, so the saved prefix is exactly the system tokens. Every subsequent request hits at the system-prompt boundary regardless of user-message length. Off by default. The existing trim/align path is unchanged and still handles raw /v1/completions, where there's no <|User|> token to anchor on.

antirez:
@unsaltedbutter-ai thanks, I'll arrive to this PR soon :)

antirez:
Indeed this is a very good change @unsaltedbutter-ai, thank you. I fixed it in a different way after testing four different coding agents. Even in that case, if we stick to the rule of trim at the most advanced  < | User | > that is before the assistant mark, we get the best of all the worlds, without any configuration setting, and a single fixed behavior.

unsaltedbutter-ai:
thank you very much. I'm loving ds4

PR #67, opened 2026-05-11
feat: multi-session pool with zero-wait switching between agents and context windows

skyfallsin:
Right now, a single `ds4-server` process has one live KV state. That works well for one linear conversation, but it hurts multi-agent workflows: switching between agents or context windows can force a long re-prefill even when each conversation was already warm recently.

This PR adds a multi-session pool for zero-wait switching between active context windows, plus prompt-rendering and cache-safety changes that make reuse stable and correct.

### Multi-session pool

`--sessions N` keeps N backend sessions resident in memory. The worker still runs one inference job at a time, but it can switch which live session it uses without rebuilding that session's KV state.

Routing order:

1. `X-Session-Id` exact match
2. Best live prefix match when no session id is provided
3. Empty slot
4. LRU eviction

If `--kv-disk-dir` is configured, evicted pool slots are persisted by session id and can be restored later. This gives the in-memory pool an L2 cache for returning agents.

### Prompt layout for KV prefix stability

The renderer now splits system messages into:

- leading system messages: before the first non-system message
- trailing system messages: after the first non-system message

The final layout is:

```text
BOS + [leading system] + [tools] + [trailing system] + [conversation]
```

This keeps the base prompt + tools byte-stable across sessions, while still allowing dynamic per-session context such as memory, daily logs, or project state. The goal is better cross-session KV reuse without changing the model-visible conversation semantics.

### Warmup endpoint

`POST /v1/warmup` accepts the same request shape as chat completions, requires `X-Session-Id`, queues a prefill-only job, and returns `202 Accepted` immediately.

This is useful when a client has just compacted or prepared a context window and wants the server to rebuild that session before the next user-visible request arrives.

### Prompt tokenization cache

Tool schemas are rendered and tokenized once per unique raw tools JSON. On cache hits, the server attempts to splice cached tool tokens into the full prompt token stream.

Because BPE tokenization is not generally compositional across arbitrary text boundaries, the splice is validated against full tokenization before it is used. If the token streams differ, the server falls back to full tokenization.

## Changes to existing server/session code

Beyond adding the new pool module, this PR adjusts existing server/session paths so the pool fits the current architecture safely:

- `ds4_session_progress_fn` now returns `bool`, where `true` means “abort this prefill.” Existing CLI/server/test callbacks were updated to return `false` during normal progress. This gives long chunked prefill a clean cancellation hook without polling global state inside the graph code.

- Chunked prefill now accepts an optional abort callback and checks it between layers. This is used by `DELETE /v1/pool/:slot` to stop long active jobs at safe boundaries.

- `ds4-server` no longer assumes one permanent session owns all live KV. In pool mode, the worker selects the appropriate pool slot before syncing/generating, and `s.session` points at that selected slot for the duration of the job.

- Server shutdown now treats single-session mode and pool mode separately. Single-session mode persists/frees the legacy session; pool mode persists/frees active pool slots.

- Queue/pool diagnostic state is now tracked in the server and exposed through `/v1/pool`: queued job count, queued session ids, active slot metadata, last acquire stats, and prompt-cache stats.

- Existing disk KV cache behavior was adjusted so continued checkpoints are reusable. Continued frontiers are aligned prefix checkpoints for future reuse, so they should not be consumed/deleted after one load like one-shot cold checkpoints.

- Chat prompt rendering now places tools before trailing system messages. That gives stable `leading system + tools` prefixes for cross-session KV reuse while preserving dynamic context inside the rendered prompt.

- HTTP response handling now includes `202 Accepted` for the async warmup endpoint.

## New endpoints

- `GET /v1/pool` — pool status, active slots, queue snapshot, prompt-cache stats
- `DELETE /v1/pool/:slot` — invalidate an idle slot or abort the active job on that slot
- `POST /v1/warmup` — async prefill for a session id

## Benchmarks

DeepSeek V4 Flash IQ2_XXS, M4 Max 128GB, `--ctx 100000 --sessions 4`:

| System prompt | Cold prefill | Pool switch TTFT | Speedup |
|---|---:|---:|---:|
| 1K tokens | 3.34s | 0.59s | 5.7x |
| 10K tokens | 42.9s | 0.64s | 67x |
| 25K tokens | 117.8s | 0.83s | 142x |
| 50K tokens | 263.7s | 0.88s | 298x |
| 60K tokens | 487.8s | 1.02s | 477x |

## Validation

- `make ds4-server ds4_test`
- `make test`
- `git diff --check`

Added coverage for:

- session-id affinity
- LRU eviction
- empty-slot preference
- prefix-match routing
- prompt-cache hit/miss behavior
- prompt-cache overflow
- cached tool tokenization matching full tokenization

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,9 +55,9 @@ ds4-server: ds4_server.o rax.o $(CORE_OBJS) $(POOL_OBJS)
 ds4-bench: ds4_bench.o $(CORE_OBJS)
 	$(CC) $(CFLAGS) -o $@ ds4_bench.o $(CORE_OBJS) $(METAL_LDLIBS)
 
-cpu: ds4_cli_cpu.o ds4_server_cpu.o ds4_bench_cpu.o linenoise.o rax.o $(CPU_CORE_OBJS)
+cpu: ds4_cli_cpu.o ds4_server_cpu.o ds4_bench_cpu.o linenoise.o rax.o $(CPU_CORE_OBJS) $(POOL_OBJS)
 	$(CC) $(CFLAGS) -o ds4 ds4_cli_cpu.o linenoise.o $(CPU_CORE_OBJS) $(LDLIBS)
-	$(CC) $(CFLAGS) -o ds4-server ds4_server_cpu.o rax.o $(CPU_CORE_OBJS) $(LDLIBS)
+	$(CC) $(CFLAGS) -o ds4-server ds4_server_cpu.o rax.o $(CPU_CORE_OBJS) $(POOL_OBJS) $(LDLIBS)
 	$(CC) $(CFLAGS) -o ds4-bench ds4_bench_cpu.o $(CPU_CORE_OBJS) $(LDLIBS)
 
 cuda-regression:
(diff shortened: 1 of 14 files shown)


skyfallsin:
Merged main into the branch to resolve conflicts with the thinking checkpoint refactor (the `ds4_session_rewrite_from_common` + rebuild-via-disk changes from #56's follow-up). I took the upstream canonicalization logic wholesale — it's strictly better than what I had — and kept the pool code alongside it. The only real decision point was conflict 8 (post-prefill hook): I inserted the pool template save call right before the existing `cold_store_len` check, using your new `prompt_for_sync` pointer. Also fixed the progress callback signature change in `ds4_test.c` (`void` → `bool`).

skyfallsin:
Removed the template system and background flusher. Both were redundant with the KV disk cache's SHA1-based continued checkpoints, which already store aligned frontiers (10K, 20K, 30K...) during prefill for cross-session reuse.

Pool L2 (session-id addressed) now only saves after generation completes and on eviction/shutdown. Fixed two bugs: continued frontier files were being deleted after first load (`cold_max_tokens` consume logic), and partial pool prefix matches were being discarded in favor of worse KV cache hits.

skyfallsin:
@antirez simplified substantially. no extra template calls, PR description has been updated with the latest.

antirez:
Will check, thank you

skyfallsin:
@antirez any notes? Would love to update if so and resubmit

skyfallsin:
closing!

PR #72, opened 2026-05-11
Sub 2bit iq1s down

adis-b:
64 GB Target (experimental) version added


antirez:
In theory interesting, in practice is not a matter of code but of how it performs at 1 bit quant. Only if it is more useful than a smaller model it makes sense. It needs to bet qwen 3.6 35B at least in some way, otherwise it is not useful, and I bet this can be achieved with 1 bit quants... Did you have any test prompt output?

adis-b:
I will have the details tomorrow, I've added another feature as well (https://github.com/adis-b/ds4-64gb/tree/sparse-residency-64gb).

Issue #13, opened 2026-05-08
/v1/chat/completions streams thinking into content instead of reasoning_content when no tools are requested


luisi-tech:
May not be an issue for most and I’m not savvy enough to know if this would be the same for other chat hosts, but experienced this and asked Opus to summarize what it found when using ds4-server with openai:

Streaming chat requests without a tools field bypass the structured OpenAI streamer and fall through to the plain SSE path, which emits every token (including the literal </think> closer) as content deltas. Reasoning never lands in reasoning_content, and clients like Jan that look for balanced <think>...</think> tags see only the closer.

Repro:

curl -N http://127.0.0.1:8000/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{"model":"deepseek-v4-flash","messages":[{"role":"user","content":"What is 17 * 23?"}],"stream":true}'

Observed: every delta is {"content":"..."}, including one {"content":"</think>"} mid-stream.
Expected: thinking tokens in reasoning_content deltas, no </think> literal on the wire.
Cause is in ds4_server.c around line 4964: structured_stream and openai_live_tools both require j->req.has_tools, so non-tool chats skip openai_sse_stream_update (which already handles thinking correctly) and use the plain sse_chunk path at line 5083. Dropping the has_tools requirement so the OpenAI structured streamer runs for all OpenAI streaming chats fixes it; verified locally against Jan.

fry69:
I do not see this behavior, on my machine running current [main](https://github.com/antirez/ds4/commit/d615ab08c8bce9b8242963ecece5aed6b5a79367), the response gets correctly split into `reasoning_content` and `content` delta parts.

Note that there were some recent fixes regarding `ds4-server` and reasoning token encapsulation, so make sure you run the latest commit.

Command:
```shell
curl -N http://127.0.0.1:8000/v1/chat/completions -H 'Content-Type: application/json' -d '{"model":"deepseek-v4-flash","messages":[{"role":"user","content":"What is 17 * 23?"}],"stream":true}'
```

Output:
```shell
data: {"id":"chatcmpl-3","object":"chat.completion.chunk","created":1778258878,"model":"deepseek-v4-flash","choices":[{"index":0,"delta":{"role":"assistant"},"finish_reason":null}]}

data: {"id":"chatcmpl-3","object":"chat.completion.chunk","created":1778258878,"model":"deepseek-v4-flash","choices":[{"index":0,"delta":{"reasoning_content":"We "},"finish_reason":null}]}

[...]

Issue #11, opened 2026-05-08
Hallucinated tool call end tokens on 2-bit


mitsuhiko:
I saw this a few times now but I'm not sure what to make of it.  Basically at one point where reasoning was supposed to end I saw this:

```
    if name == "main":
        run()
    </｜DSML｜parameter>
    </｜DSML｜invoke>
    </｜DSML｜tool_calls>
```

You can see a pi session where this behavior showed up here: https://pi.dev/session/#37a4801fd89fb9ee20f525825ef634f1

I'm not sure yet (don't have enough debug output) if the model ended up hallucinating bad tokens or if the parser state ended up corrupt.

antirez:
Thank you! Checking.

antirez:
In theory the issue is that when a tool call end tag does not match the start, there is a bug (easy to fix), but in practice the question is to check if it is legit for the model to allucinate a closing tag in this moment. Likely yep if it believe it is creating a new file, but I'm investigating a bit more. There is always the option of preventing tool calls tags if no matching opening tag is found or to even just force the grammar of everything is tool calling.

antirez:
What I want to do here is this: if we are into thinking mode, if token 128825 is emitted  (| DSML |), we don't allow it, and sample from the remaining tokens masking it. The model will have to recover in some other way if it was about to emit tool opening/closing whatever in the wrong context.

antirez:
@mitsuhiko I zoomed a lot into the problem, and used a script that continuously try to do a similar chat, but it never happens that DS4 hallucinates the tool calling (or closing) inside the thinking stanza. When it happens (I can force it using a session similar to yours) what works to avoid the problem is substituting the user-defined | DSML | token with </think>, and the model recover, however I'm not sure if actually the problem is really of the model, because of lack of replication. So please if you could run the server with --trace /tmp/trace.txt when you do tests, in case this happens again I would love to obtain the trace: this would really tell us if there was no tool call opening tag, and if the tool call happened inside thinking or not. In the meantime I'll try harder :) Thanks.

antirez:
[...]

Issue #10, opened 2026-05-08
Would this work on Ubuntu 24.04 + NVIDIA RTX 5060(8GB)?


TerryLiu:
I have a laptop with an Intel Core i7-13645HX processor ‌ and ‌NVIDIA GeForce RTX 5060 graphics card (8GB of video memory)‌,  16GB RAM + 512GB SSD, and I want to run a ds4 on it.  please!

xaionaro:
TLDR: It won't run.

I was just wondering by (never used this project, never read it carefully), but:
* I think at the very least you need 128GB RAM. Moreover, if you want to compute it on GPU, then you essentially need GPU-accessible 128GB (VRAM or unified RAM).
* Moreover it mentions Metal, which in these contexts usually implies MacBooks/Macs with MacOS.

I guess you can make 128GB of swap, and make the project compute on CPU (instead of Metal). But would be extremely slow.

PR #73, opened 2026-05-11
fix(ds4): Implement MoE low-memory streaming to work around macOS's kernel bug.

HomeroRR:
# CPU `--low-mem` Streaming Implementation Summary

## What Was Implemented

CPU streaming feature for DeepSeek V4 inference on 8GB RAM systems, complementing the already-working Metal GPU streaming path.

### Architecture

**Two-Phase Per-Layer Processing:**
- **Phase 1**: Load non-expert layer weights from disk, compute attention + routing, save expert selection to scratch
- **Phase 2**: Load selected 6 experts from disk, compute MoE + output using streamed experts

**Memory Model:**
- Layer buffer: 256 MB (one layer at a time)
- Expert buffer: 64 MB (six experts at a time)
- KV cache: ~100 MB (compressed, 8K context)
- Scratch: ~100 MB (activation buffers)
- **Peak**: ~1 GB (vs 10+ GB with full model in memory)

### Code Changes

| File | Lines | Change |
|------|-------|--------|
| ds4.c | 16419 | Removed `--cpu --low-mem` rejection guard |
| ds4.c | 220-223 | Added `stream_selected[]`, `stream_router_weights[]`, `layer_buf`, `expert_buf` to struct |
| ds4.c | 5367-5430 | Added `streaming_phase`, `expert_model`, `scratch` params to `layer_routed_moe_one_prealloc()` with phase logic |
| ds4.c | 5683-5697 | Added params to `layer_ffn_one_decode_scratch()` |
| ds4.c | 7369-7381 | Added params to `layer_forward_raw_swa_one()` |
| ds4.c | 7574-7642 | Created `forward_token_raw_swa_cpu_streaming()` function (new) |
| ds4.c | 15182-15318 | Integrated streaming into `generate_raw_swa_cpu()` with buffer allocation/deallocation |
| ds4.c | All callers | Updated 7 call sites with `0, NULL` (non-streaming) or appropriate phase values |

### Reused Infrastructure

- `stream_load_layer()`: Loads all 32 non-expert tensors per layer
- `stream_layer_build_temp()`: Creates temporary model pointing to stream buffer
- `stream_load_experts()`: Packs 6 selected experts into contiguous layout
- All three functions were implemented in previous sessions for Metal GPU path

### Two-Phase Logic

```c
/* Phase 1: Attention + Routing */
layer_forward_raw_swa_one(..., streaming_phase=1, expert_model=NULL);
// Computes routing, saves to scratch->stream_selected[], returns early

/* Phase 2: MoE + Output */
layer_forward_raw_swa_one(..., streaming_phase=2, expert_model=expert_model);
// Skips routing (already computed), uses streamed experts, completes output
```

### Key Design Decisions

1. **Phase 1 Returns Early**: Avoids duplicate computation of routing; can be optimized later
2. **Stack Variables for Non-Streaming**: Batch prefill paths don't need scratch, use stack arrays
3. **Expert Remapping**: Phase 2 remaps expert IDs {N, M, K, ...} to {0,1,2,3,4,5} for contiguous buffer access
4. **Backward Compatibility**: All existing code paths work unchanged with `streaming_phase=0, expert_model=NULL`

---

## Testing

A comprehensive test plan has been created: **[CPU_STREAMING_TEST_PLAN.md](CPU_STREAMING_TEST_PLAN.md)**

### Quick Verification (Non-Building)

```bash
# Check structure definitions
grep -n "stream_selected\|stream_router_weights\|layer_buf\|expert_buf" ds4.c | head -5

# Check Bug #14 guard removed
grep -A 5 "low_mem && opt->backend == DS4_BACKEND_CPU" ds4.c | wc -l  # Should be 0

# Check streaming function exists
grep -n "forward_token_raw_swa_cpu_streaming" ds4.c | head -2

# Check phase logic
grep -n "streaming_phase == 1" ds4.c | wc -l  # Should be > 0
```

### Build & Test

```bash
# Build
cd /workspaces/ds4
make clean && make

# Test 1: Streaming mode
./ds4 --cpu --low-mem -p "Hello" -n 10

# Test 2: Regression (normal mode)
./ds4 --cpu -p "Hello" -n 10

# Test 3: Output comparison
./ds4 --cpu -p "Test" -n 5 > /tmp/normal.txt
./ds4 --cpu --low-mem -p "Test" -n 5 > /tmp/streaming.txt
diff /tmp/normal.txt /tmp/streaming.txt  # Should be identical
```

---

## Known Limitations

1. **lm_head Not Streamed** (~536 MB always resident)
   - Not required for 8GB systems (only ~1 GB peak with lm_head)
   - Future optimization if needed

2. **Token-by-Token Prefill** (not batched)
   - Slower than layer-major (necessary for architecture)
   - Typical trade-off for memory efficiency

3. **macOS Metal Requires macOS**
   - CPU path works on all platforms
   - Metal testing requires macOS with Metal support

---

## Related Work

This implementation completes a 15-bug fix series:

| Bug # | Issue | Status | Session |
|-------|-------|--------|---------|
| #1-5 | Forward decl, tensor offset, shadowing, mmap, array size | ✅ FIXED | S1 |
| #6-7 | Startup abort, use-after-free | ✅ FIXED | S2 |
| #8-9 | Diagnostic test functions, warm-weights crash | ✅ FIXED | S3 |
| #10 | Metal API return value checks inverted | ✅ FIXED | S4 |
| #11-12 | Routing staleness, expert remapping OOB | ✅ FIXED | S5 |
| #13-15 | CPU path bugs (compile, crash, progress) + **STREAMING** | ✅ FIXED | **S6** |

All 15 bugs are now fixed across both Metal and CPU paths.

---

## What's NOT Included (Deferred)

- lm_head streaming (optimization, not required for 8GB)
- Batched CPU streaming prefill (would require significant refactoring)
- INT8 KV cache compression (KV cache is already only ~100 MB)

---

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,6 @@
 MIT License
 
+Copyright (c) 2026-2036 Homero Roman
 Copyright (c) 2026 The ds4.c authors
 Copyright (c) 2023-2026 The ggml authors
 
--- a/ds4.c
+++ b/ds4.c
@@ -14,6 +14,7 @@
  * no-copy MTLBuffers.
  */
 
+#define _XOPEN_SOURCE 600
 #include <errno.h>
 #include <fcntl.h>
 #include <float.h>
@@ -52,6 +53,9 @@
 #define DS4_RMS_EPS ( 1.0e-6f)
 #define DS4_HC_EPS  ( 1.0e-6f)
 #define DS4_EXPERT_WEIGHT_SCALE (1.5f)
+
+/* Engine flags for low-memory streaming mode */
+#define DS4_FLAG_LOW_MEM (1 << 0)
 #define DS4_SWIGLU_CLAMP_EXP    (10.0f)
 #define DS4_ROPE_FREQ_BASE      (10000.0f)
 #define DS4_ROPE_SCALE_FACTOR   (16.0f)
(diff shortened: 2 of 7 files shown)


antirez:
Hi @HomeroRR thank you for the PR. I would have /v1/models as it is without additional fields other than openrouter / openai stuff, and would instead add our ds4 specific things in /props. So we overload the llama.cpp convention with our things in a similar way to llama.cpp, but we can leave the "standard" endpoint as clean as possible.

PR #82, opened 2026-05-11
Fuse block-sum accumulation into `ds4_quantize_row_q8_K pass`

GGraziadei:
This PR optimizes the ds4_quantize_row_q8_K hot path by merging the quantization step and the block-sum calculation into a single memory pass.
Previousl y, the function performed two separate iterations over the data one to quantize and one to accumulate sums which resulted in redundant memory traversals.  By initializing the bsums array and accumulating values inline during the primary quantization loop, we can improve cache locality and reduce the overall instruction count.


antirez:
This is likely slower. To provide speed improvements without benchmarks... does not work ;)

Issue #22, opened 2026-05-09
Question: any interest in OpenAI Responses API support for Codex compatibility?


alessandrobologna:
Hi,

First, thanks for publishing ds4. I have been experimenting with it in the last couple of days, both on an M3 Ultra 512 and an M5 128, with both quants, and it's really working well for me (at least for what I have tested it with: opencode and pi).

I wanted to try it as well with Codex, but [now it requires using the Responses API](https://github.com/openai/codex/discussions/7782), which is not yet supported here, and so I had Codex with GPT-5.5 build support for it... Modern times dogfooding ;)

For reference, here is the merged PR in my fork:
https://github.com/alessandrobologna/ds4/pull/1

The implementation currently:

- adds `/v1/responses` endpoint with Responses-style request parsing and Responses SSE streaming events
- extends tool-schema handling to support Responses tools, tool search tools, and namespaced tools, preserving wire name + namespace
- updates README to document the new endpoint and Codex usage via `wire_api = "responses"`

I'd be happy to contribute it, or maybe you can just use it as a starting point to look at. Either way works. So no worries at all if this is outside the direction you have in mind. I just wanted to ask first before sending a larger compatibility PR.

Cheers, and thanks again for working on this.

alessandrobologna:
Hi,
FYI, I rebased the Responses API branch on the latest upstream `main`, so it should be easier to look at without also having to account for old conflicts.

Branch:
https://github.com/alessandrobologna/ds4/tree/feature/responses-api

Current branch tip:
`0544533dd52b7f6847f6264c7ec2d9f19c9db91c`

Upstream commit it is based on:
`e88a51fdac110ca5c0e0da06f1a27d4c0313b563`

This branch includes the original Responses endpoint work plus the follow-up fixes needed after the newer DSML / KV replay changes landed upstream. Just leaving the current pointer here in case it becomes useful at some point.

fry69:
FYI: The `Responses` API PR just got merged into `main` -> https://github.com/antirez/ds4/commit/e88a71eb2ee48af2c67cf42b0c1cd83683646928

PR #91, opened 2026-05-12
feat(server): add /v1/responses (OpenAI Responses API) for Codex CLI

audreyt:
Implements the Responses API endpoint that Codex CLI (and other modern OpenAI tooling) speaks instead of /v1/chat/completions. The wire format is documented in OpenAI's Responses API; this implementation has been iterated against the Codex CLI binary's SSE parser shape until no remaining schema gaps were found.

Request parsing (parse_responses_request, parse_responses_input):
- Accepts the typed input array (message, function_call, function_call_output, reasoning, custom_tool_call(_output), local_shell_call(_output), web_search_call(_output), tool_search_call(_output), image_generation_call(_output), compaction, context_compaction).
- Maps hosted-tool history to function_call/function_call_output so prior actions survive across turns; rejects unknown item types and non-completed status with 400 to avoid silent context loss.
- Strict content-array parsing: only string|null|array of recognized text blocks (input_text/output_text/text/summary_text/ reasoning_text); rejects non-text modalities (input_image/file/ audio) instead of accepting an empty prompt.
- Merges adjacent function_call items into the preceding assistant message so text + tool-call turns render as a single assistant block.
- Honors reasoning.effort (incl. "minimal"/"none") and gates reasoning summary surface on reasoning.summary opt-in.
- Rejects previous_response_id, conversation, and forced tool_choice explicitly (constrained decoding / persisted state not supported).

Output (responses_sse_*, responses_final_response):
- Emits the full streaming lifecycle: response.created, output_item.added/.done, reasoning_summary_part.added/.done, reasoning_summary_text.delta/.done, content_part.added/.done, output_text.delta/.done, function_call_arguments.delta/.done, response.completed.
- Branches the terminal event by finish reason: response.failed for errors and response.incomplete with reason "max_tokens" for length.
- Every event carries sequence_number; every output_text part carries annotations:[]; function_call output_item.added ships with an empty arguments string (full args arrive via function_call_arguments.done and output_item.done), and item ids are stable across added/done.
- Tracks whether </think> was actually observed so a truncated stream marks the reasoning item incomplete instead of "completed".
- Recovers gracefully when the DSML tool parse fails after the model was suppressed at the tool marker: the suppressed tail is flushed as additional output_text deltas so the streamed message matches output_item.done.

Tested by 25 rounds of /codex:adversarial-review against the same client this is meant to feed.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -475,6 +475,7 @@ typedef enum {
 typedef enum {
     API_OPENAI,
     API_ANTHROPIC,
+    API_RESPONSES,
 } api_style;
 
 static void random_tool_id(char *dst, size_t dstlen, api_style api) {
@@ -580,6 +581,10 @@ typedef struct {
     ds4_think_mode think_mode;
     bool has_tools;
     bool prompt_preserves_reasoning;
+    /* For /v1/responses: emit reasoning_summary_* events / fields only when the
+     * client opted in via reasoning.summary. Other APIs leave this false; the
+     * field is ignored on those code paths. */
+    bool reasoning_summary_emit;
     tool_replay_stats tool_replay;
 } request;
 
@@ -703,7 +708,7 @@ static void request_free(request *r) {
 }
 
 static ds4_think_mode think_mode_from_enabled(bool enabled, ds4_think_mode effort) {
-    if (!enabled) return DS4_THINK_NONE;
+    if (!enabled || effort == DS4_THINK_NONE) return DS4_THINK_NONE;
     return effort == DS4_THINK_MAX ? DS4_THINK_MAX : DS4_THINK_HIGH;
 }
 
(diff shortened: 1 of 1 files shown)


simonw:
Just came here to feature-request this because I wanted to try Codex CLI.

With this patch applied I think we should be able to add this to `~/.codex/config.toml`:
```toml
[model_providers.ds4]
name = "DS4"
base_url = "http://127.0.0.1:8000/v1"
wire_api = "responses"
stream_idle_timeout_ms = 1000000
```
And then run:

    codex --model deepseek-v4-flash -c model_provider=ds4

fry69:
FYI: There is the issue #22 which also contains a branch with Responses API implementation (but no PR).

mitsuhiko:
I think it's a bit unfortunate if this ends up with many different protocol versions in the core server because it will be tricky to support them all at the same level of quality without blowing up the complexity. I wonder if it would not be better to have a separate proxy that translates to one common protocol.

antirez:
@mitsuhiko I understand the complexity limitation argument, but I have the feeling the API side is the less fragile part of all this, if we have good tests in place. The tool calling, which is the nightmare part, should be in theory shared across the paths. I'll investigate this PR soon and report back.

antirez:
Handling this, news soon.

antirez:
You can find the candiate for merging here: [responses-api](https://github.com/antirez/ds4/tree/responses-api), it implements continuing from the current KV memory checkpoint matching IDs without the need for prefix matching.

antirez:
Merged :) Thanks.

davidgasquez:
Just wanted to say what an awesome bunch of humans you all are! :raised_hands:

PR #93, opened 2026-05-12
README.md: add usage instructions for swival.dev

jedisct1:
Add basic instructions to use DS4 with Swival.dev.

### DIFF
--- a/README.md
+++ b/README.md
@@ -384,6 +384,31 @@ Optionally make it the default Pi model in `~/.pi/agent/settings.json`:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


fry69:
Why does a client that uses the generic OpenAI interface need a shoutout in the README?

There seems to be nothing special about this client. This feels a bit too much like name dropping for my taste.

Update: The force pushed changes still look either too generic or are client specific quirks. Those should best served from the homepage of said client. Otherwise the README becomes a bloated collection of all quirks of all possible clients. (personal opinion, I am not the maintainer)

PR #70, opened 2026-05-11
support browser CROS

donaldxu:
to support browser cross error, I add some header meta

PR #94, opened 2026-05-12
[proposal] - Allow CORS Headers for intranet / VPN purposes

WebReflection:
This MR has been successfully tested in my local network with a DGX Spark around the WiFi and I need this variant to be able to query via my `localhost` or any other connected device that Spark so that we can all benefit from this project within my house.

Thanks for considering this change/update.

### To be discussed

Ideally there should be a `--cors` flag when starting the server but I'd like to start with this implementation that "*just works*" :tm: and hear out from others/maintainer if there's anything else I can improve/change but trust me it works already and I am playing around a tiny library that would let me lurk *ds4* from anywhere I am in my own apartment, as long as the DGX is up and running.

P.S. thanks a lot for this project, I will inevitably try to bring it to ROCm once I have my machine around but so far with Spark it's working wonderfully!

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -4644,10 +4644,21 @@ static bool http_response(int fd, int code, const char *type, const char *body) 
         "HTTP/1.1 %d %s\r\n"
         "Content-Type: %s\r\n"
         "Content-Length: %zu\r\n"
+        "Access-Control-Allow-Origin: *\r\n"
+        "Access-Control-Allow-Methods: GET,HEAD,OPTIONS,POST,PUT\r\n"
+        "Access-Control-Allow-Headers: Access-Control-Allow-Headers, Origin, Accept, Content-Type, Access-Control-Request-Method, Access-Control-Request-Headers\r\n"
         "Connection: close\r\n\r\n",
         code, reason, type, strlen(body));
     bool ok = send_all(fd, h.ptr, h.len) && send_all(fd, body, strlen(body));
     buf_free(&h);
+    return ok;
+}
+
+static bool http_ok(int fd) {
+    buf b = {0};
+    buf_puts(&b, "OK");
+    bool ok = http_response(fd, 200, "text/plain", b.ptr);
+    buf_free(&b);
     return ok;
 }
 
@@ -4668,6 +4679,9 @@ static bool sse_headers(int fd) {
     const char *h =
         "HTTP/1.1 200 OK\r\n"
         "Content-Type: text/event-stream\r\n"
+        "Access-Control-Allow-Origin: *\r\n"
+        "Access-Control-Allow-Methods: GET,HEAD,OPTIONS,POST,PUT\r\n"
+        "Access-Control-Allow-Headers: Access-Control-Allow-Headers, Origin, Accept, Content-Type, Access-Control-Request-Method, Access-Control-Request-Headers\r\n"
         "Cache-Control: no-cache\r\n"
         "Connection: close\r\n\r\n";
     return send_all(fd, h, strlen(h));
@@ -11116,6 +11130,11 @@ static void *client_main(void *arg) {
         goto done;
     }
 
+    if (!strcmp(hr.method, "OPTIONS")) {
+        http_ok(fd);
+        http_request_free(&hr);
+        goto done;
+    }
     if (!strcmp(hr.method, "GET") && !strcmp(hr.path, "/v1/models")) {
         send_models(s, fd);
         http_request_free(&hr);
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -4633,7 +4633,14 @@ static void append_tool_call_deltas_json(buf *b, const tool_calls *calls, const 
     buf_putc(b, ']');
 }
 
-static bool http_response(int fd, int code, const char *type, const char *body) {
+static void append_cors_headers(buf *h) {
+    buf_puts(h,
+        "Access-Control-Allow-Origin: *\r\n"
+        "Access-Control-Allow-Methods: GET, POST, OPTIONS\r\n"
+        "Access-Control-Allow-Headers: *\r\n");
+}
+
+static bool http_response(int fd, bool enable_cors, int code, const char *type, const char *body) {
     const char *reason = code == 200 ? "OK" :
                          code == 400 ? "Bad Request" :
                          code == 404 ? "Not Found" :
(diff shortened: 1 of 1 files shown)


WebReflection:
If anyone is interested around "*how can I test this*" ?

**test.js**
```js
import Queue from 'https://esm.run/gen-q';

const { parse, stringify } = JSON;
const decoder = new TextDecoder;

const chatOptions = {
  stream: true,
  role: 'user',
};

export default class DS4 {
  #model;
  #url;

  constructor({
    url = 'http://YOUR_MACHINE_IP:8000',
    model = 'deepseek-v4-flash',
    version = 'v1',
  }) {
    this.#model = model;
    this.#url = new URL(`${url}/${version}`);
  }

  async *chat(content, { stream = true, role = 'user' } = chatOptions) {
    const items = new Queue;

    const { body } = await fetch(`${this.#url}/chat/completions`, {
      method: 'POST',
      headers: {
        'Content-Type': 'application/json',
      },
      body: stringify({
        model: this.#model,
        messages: [
          { role, content }
        ],
        stream,
      }),
    });

    const reader = body.getReader();

    new ReadableStream({
      async start(controller) {
        (function next() {
          reader.read().then(({ done, value }) => {
            if (done) {
              items.splice(0);
              controller.close();
              return;
            }
            const text = decoder.decode(value);
            if (/^\s*data:\s*(\{[\s\S]+\})\s*$/.test(text)) {
              const { $1: json } = RegExp;
              items.push(...parse(json.trim()).choices);
            }
            next();
          });
        }());
      },
    });

    for await (const item of items) {
      const { finish_reason, delta } = item;

      if (finish_reason != null) break;

      const { content, reasoning_content } = delta;
      if (content || reasoning_content)
        yield { content, reasoning: reasoning_content };
    }
  }

}
```

**index.html**
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>DS4</title>
    <style>
        body {
          font-family: sans-serif;
          &.thinking {
            opacity: 0.5;
          }
        }
    </style>
    <script type="module">
        import DS4 from './test.js';

        const ds4 = new DS4({
          url: 'http://YOUR_MACHINE_IP:8000',
          model: 'deepseek-v4-flash',
          version: 'v1',
        });

        let thinking = true;
        document.body.classList.add('thinking');
        for await (const { content, reasoning } of ds4.chat('List three Redis design principles.')) {
          if (thinking && reasoning && !content) {
            document.body.textContent += reasoning;
          }
          else if (thinking && content) {
            thinking = false;
            document.body.classList.remove('thinking');
            document.body.textContent = content;
          }
          else if (content) {
            document.body.textContent += content;
          }
        }

        console.log('done');
    </script>
</head>
<body>
    
</body>
</html>
```

The testing library is a *WIP* and it will be able to consume all channels and do more but with that, and this patch, you'll see results from a `localhost` without a sweat 🥳

WebReflection:
this might be a duplicate of https://github.com/antirez/ds4/pull/70 which I've just realized was in already ... my thoughts:

  * that MR is way more permissive **but** ...
  * one needs to explicit use `--host 0.0.0.0` instead of default `127.0.0.1` to make the machine reachable out there
  * accordingly, I am not super sure the `--cors` flag is even needed ... it might be better, as explicit intent, yet it's impractical if the host is not `0.0.0.0` or bound to something reachable from the intranet, so that we could eventually just merge 70 and call it a day, I wouldn't be upset at all as long as CORS in intranet is possible

thank you!

calvinrp:
Alternatively, #44

WebReflection:
@calvinrp answered in here https://github.com/antirez/ds4/pull/70#issuecomment-4433233946

d3y4n:
Why not add a proxy on top for all the shenanigans? IMHO this should be kept as simple as possible.

WebReflection:
@d3y4n having CORS options backed in **is** the "*as simple as possible*" idea indeed ... anything else is not simple anymore.

d3y4n:
@WebReflection not my call, just saying you're hardcoding values and tomorrow someone might need different ones (even you).
What I suggest is add `caddy` on top as this is anyways not "real" production server.

```
api.example.com {

    @preflight method OPTIONS

    handle @preflight {
        header {
            Access-Control-Allow-Origin   "{http.request.header.Origin}"
            Access-Control-Allow-Methods  "GET, POST, PUT, PATCH, DELETE, OPTIONS"
            Access-Control-Allow-Headers  "Authorization, Content-Type, Accept, X-Requested-With, X-CSRF-Token"
            Access-Control-Max-Age        "3600"
            Vary                          "Origin"
        }
        respond "" 204
    }

    header {
        Access-Control-Allow-Origin   "{http.request.header.Origin}"
        Access-Control-Expose-Headers "Content-Length, Content-Range"
        Vary                          "Origin"
    }

    reverse_proxy localhost:8080
}
```

Cheers

WebReflection:
@d3y4n PR #70 is easier and simpler, I’ll try that and amend this one.

the issue is that requiring any other tool just to have CORS (3 related PRs already, it’s not something a few needs, it’s something everyone expects as possibility at some point) means goodbye ease of portability and basic flag in docs to add such feature (if we want the flag at all, otherwise it’s about enabling CORS on 0.0.0.0 ‘cause that’s already an explicit intent).

Both devices and automated tests done remotely will need that, if browsers are used instead of curls, so yeah, everyone *could* fix it in a way or another but changes are so tiny/simple I am not sure why making it any more complicated for users would be desirable

fry69:
FYI/related: There is also the idea to create a separate web proxy that then talks to the inference engine with a generalized protocol:

There are already two competing protocols, classic stateless chat completion and the more modern, stateful "Responses". It might make sense to separate the protocol frontend from the engine, as suggested here -> https://github.com/antirez/ds4/pull/91#issuecomment-4435105068

CORS would then only be needed to get implemented in this combined frontend proxy and the engine does not have to worry about all that.

Otherwise I am a bit skeptical about making the current `ds4-server` more friendly to be used from remote places, as it is extremely bare bones and likely contains a gazillion of security related problems (personal opinion).

WebReflection:
To whom it might concern, the latest edit of this file does the following:

  * it uses minimal simplified headers from #70
  * it adds a `--cors` or `--CORS` argument like in #44
  * it uses `0.0.0.0` instead of `127.0.0.1` if the flag is passed and the `host` is `localhost`
  * it doesn't change anything else if the flag is not passed

Tested via:

```sh
# example
./ds4-server --cors --host localhost --ctx 100000 --kv-disk-dir /tmp/ds4-kv --kv-disk-space-mb 8192
```

It is working like a charm, glad to hear where/how I could improve this MR, thank you.

WebReflection:
FYI much cleaner MR, I can't find anything particularly cumbersome + tests pass but I can test only on CUDA

WebReflection:
To whom it might concern, I've rebased from main and the status is till OK. Please let me know if there's any intention to merge this otherwise every single other change will inevitably conflict due `enable_cors` boolean propagated internally, thank you.

WebReflection:
two rebases and running ... @antirez can you or anyone else involved in this project **pleaase** provide an *interest* or a *never going to happen* so that I stop rebasing this clean MR and just fork the project and deliver/use it by any other mean? Thank you 🙏

WebReflection:
If anyone is interested, I have improved the test/demo/example and it's here:
https://github.com/WebReflection/ds4-chat

Works locally in both laptops and phones, I love it already!

image

antirez:
@WebReflection sorry, I'm only one.

WebReflection:
@antirez then ask *ds4* thoughts about this 😁

jokes a part, I understand that but this feature has 3 or more opened PRs and if there's not a clear way forward or answer about "*not planned*" or "*not interested because*" probably more PRs will land 'cause CORS is a curse that doesn't need to exist within an intranet and having reverse proxies for 3 LOC extra in headers when a flag is passed feels like picking a hammer to cut a boiled egg in two so I hope we can find a way forward or, if not, we can fork and apply patches automatically every single release/change but at least I won't bother anyone following this project anymore.

Thanks in advance for any possible hint/way forward.

d3y4n:
@WebReflection IMO if you're going to add this then do it properly, just for "VPN/intranet" is niche, you should make the headers configurable (especially Origin).

WebReflection:
@d3y4n that’s a slippery slope. For anything **more complex** there are solutions, this one is rather a *dev* quick fix/hook to enable CORS, it’s not meant to replace nginx or other fully configurable products. I’m a KISS kind of person and I try to avoid feature creep any time I can

antirez:
@WebReflection fixed with a different commit and implementation, credits in the commit message. Thanks.

WebReflection:
@antirez thanos a lot, really appreciated 🙏

PR #98, opened 2026-05-12
Fix file paths in README for benchmark prompt and SVG generation

Chida82:
The rename of the folder was not reported in the readme. Although easy to understand, in this way it is immediate to launch the commands


antimatter96:
Can copy these changes as well if you want : https://github.com/antirez/ds4/pull/102

Chida82:
> Can copy these changes as well if you want : #102

Done

PR #106, opened 2026-05-12
server: resolve asset paths relative to executable

### LINKED ISSUE #49: Problems when launching from other programs like `llama-swap`

gmontana:
## Summary

Fixes #49 by resolving relative asset paths from the executable directory when they are not found from the current working directory.

This covers:

- relative GGUF model paths, including the default `ds4flash.gguf`
- relative MTP model paths
- bundled Metal source files such as `metal/flash_attn.metal`

Current-directory paths still take precedence, and the existing Metal source environment overrides are still tried first.

## Validation

- `git diff --check`
- `make ds4_test`
- `./ds4_test --paths --server`
- `make ds4-server`
- `make ds4`
- launched `/Users/u1770777/repos/ds4/ds4-server --ctx 2048 --port 5556` from `/tmp` and confirmed it reached `listening on http://127.0.0.1:5556`

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -19,6 +19,7 @@
 #include <float.h>
 #include <inttypes.h>
 #include <ctype.h>
+#include <limits.h>
 #include <math.h>
 #include <pthread.h>
 #include <stdbool.h>
@@ -35,6 +36,10 @@
 #include <unistd.h>
 
 #include "ds4.h"
+
+#ifdef __APPLE__
+#include <mach-o/dyld.h>
+#endif
 
 #ifndef DS4_NO_GPU
 #include "ds4_gpu.h"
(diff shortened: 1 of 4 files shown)


gmontana:
I saw the note on #49 about possibly inlining the Metal sources at build time. This PR is meant as a smaller runtime lookup fix for the current layout: it keeps CWD precedence and existing Metal source env overrides, but lets launchers start `ds4-server` by absolute path from another working directory.

It also covers the default model path / MTP path case, which inlining Metal sources would not address. Happy to adjust the scope if you prefer a different split.

gmontana:
Saw the `--chdir` decision in #49. That's a cleaner solution than my path-resolution approach. Closing this in favor of that direction.

PR #109, opened 2026-05-12
docs: add Jetson Thor CUDA notes

amarrmb:
Adds documentation for building CUDA binaries for Jetson Thor with CUDA_ARCH=sm_110, exposing ds4-server on a LAN, and setting Jetson power/clocks before benchmarking.


antirez:
Thanks. I'm changing the build system so that when you type "make", for Linux CUDA you are offered a list of targets, like make cuda-spark, cuda-generic, or make cuda CUDA_ARCH=..., so it is probably better to modify this PR in this way once the commits land the repo today.

amarrmb:
Sounds good, will rework and submit. Thanks!

PR #111, opened 2026-05-13
cancel prefill on client disconnect

n30g30:
problem: ds4-server only notices client disconnects after prompt prefill finishes, because prefill performs no socket writes. Long-context requests can therefore keep running for minutes after the client has cancelled.

solution: Check the client socket from the prefill progress callback and cancel the active session when the connection is gone. `ds4_session_sync()` now observes that cancellation between prefill chunks and returns early.


n30g30:
closing since the upstream now has cleaner solution

PR #128, opened 2026-05-14
cuda: prepare model-map ownership for MTP weight caching

Entrpi:
## Draft PR stack

1. #130 Generalized proof harness and VMM weight server (independent infrastructure; merge first)
2. #128 CUDA model-map/fd-cache ownership prep
3. #129 CUDA MTP draft support and proof matrix
4. #132 CUDA MTP hot-path optimization iteration
5. #131 Purpose-built MTP verifier and exactness workflows

All PRs are drafts against `antirez/ds4:main`. GitHub cannot retarget these upstream PRs to branch bases that exist only in the fork, so this page may temporarily show #130 commits in its diff. The branch is rebuilt on top of #130 and the full merge chain has been checked clean.

## Purpose

Prepare CUDA model-cache ownership for MTP/shared-weight work. The CUDA fd-backed cache is bound to the model map that owns the fd so later auxiliary model maps, such as MTP draft GGUFs, cannot read bytes from the wrong file offsets.

This branch also composes the MTP startup pre-cache behavior with the shared-weight import hook from #130: shared ranges are imported first when a weight-server manifest is present, then normal startup caching/pre-caching proceeds.

## Dependency

Review after #130. #129 builds on this behavior.

## Validation

- Full-stack disposable merge check passed through #130 -> #128 -> #129 -> #132 -> #131.
- Local `make cpu` passed at the final stacked state.
- Earlier DGX Spark `make cuda-spark` passed.
- Earlier DGX Spark `make cuda-regression` passed on immediate rerun (`0.009s`). Notable: the first cold run exceeded the strict top-k timing guard once (`2.087s` vs `2.000s`), then the same binary passed on rerun.
- Earlier DGX Spark `make cpu` passed.
- Earlier `ds4-bench` smoke versus `origin/main` showed no speed-regression signal for this prep change: prefill `398.15 t/s` vs `399.08 t/s`, generation `13.87 t/s` vs `13.88 t/s`.


Entrpi:
Closing this draft in favor of a cleaner rebuilt PR series with the proof harness first and the downstream MTP work ordered incrementally.

PR #134, opened 2026-05-14
Guard KV cache against page-cache pressure

Ghatage:
The disk KV cache uses plain read/write to avoid mapping more VM into a process that already maps a large GGUF, but the bytes still land in the Linux page cache where they compete with the mmapped weights for resident memory. A 30k-token cold save leaves hundreds of MiB sitting in Cached: after the save returns, exactly the RAM pressure the no-mmap decision was meant to avoid.

Hint the kernel to invalidate those pages with
posix_fadvise(POSIX_FADV_DONTNEED) right after each full payload write and read. Header-only scans are deliberately untouched, since repeated small header reads still benefit from page-cache reuse. Expose DS4_KV_KEEP_PAGES=1 as an escape hatch for diagnostic comparisons, mirroring the existing cuda_model_drop_file_pages and its DS4_CUDA_KEEP_MODEL_PAGES toggle.

Correctness: make test; ./ds4_test --server.
Two new mincore-based tests assert that resident pages of a 4 MiB temp file drop from ~all to under 25% after the hint, and that DS4_KV_KEEP_PAGES=1 keeps them in place.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -5557,6 +5557,23 @@ static void kv_fill_header(uint8_t h[KV_CACHE_FIXED_HEADER], uint8_t quant_bits,
     le_put64(h + 40, payload_bytes);
 }
 
+/* Hint the kernel to drop clean page-cache pages for an open .kv file.
+ * The KV cache deliberately avoids mmap so cache restore does not add VM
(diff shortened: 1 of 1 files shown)


antirez:
Like this. Will take care.

PR #140, opened 2026-05-14
reference: full CUDA MTP proof harness stack

Entrpi:
## Stack Story

This stack is about making CUDA MTP work reviewable, measurable, and safe to optimize on DGX Spark / GB10.

The first problem is iteration speed. Engine optimization tests need process isolation so crashes, environment changes, and experimental CUDA paths do not contaminate later runs. But isolated tests are painfully slow if every worker reloads and re-uploads the same 80+ GiB model weights. The proof harness and VMM weight server solve that by keeping workers isolated while preserving immutable CUDA weight ranges in a dedicated owner process that can re-serve them to each test worker.

The second problem is enabling CUDA MTP without corrupting the base-model loading assumptions. MTP introduces a second GGUF map in the same engine process, so CUDA fd-backed weight caching must know which mmap owns the fd. The prep layer makes that ownership explicit before CUDA MTP support starts using both base and draft model maps.

The third problem is making CUDA MTP useful without hiding correctness or performance costs. The stack adds CUDA MTP draft support, proof coverage, verifier-sized hot-path optimizations, and exact verifier workflows that compare MTP behavior against no-MTP and fast-MTP baselines. The later PRs focus on making the verifier path observable enough to continue optimizing it without guessing.

The final proof/reporting layer turns those checks into repeatable review artifacts: named proof budgets, normalized steady-state timing fields, and direct speed-bench comparison of no-MTP against exact MTP across long context frontiers.

In short: this stack builds the testing infrastructure first, then makes CUDA MTP safe to run, then adds the measurement and verifier machinery needed to keep optimizing it responsibly.

## Intended Review Path

1. #135 Generalized proof harness and VMM weight server
2. #136 CUDA model-map/fd-cache ownership prep
3. #137 CUDA MTP draft support and proof matrix
4. #138 CUDA MTP hot-path optimization iteration
5. #139 Purpose-built MTP verifier and exactness workflows
6. #141 Proof budgets, steady-state timing reports, and exact MTP speed-bench comparison

## Why This PR Exists

This is a reference PR showing the full final stack as one diff against `main`. It is not meant to replace the split review path unless maintainers prefer landing the whole stack as one PR.

GitHub cannot base upstream PRs on intermediate branches that only exist in a fork. Until prior PRs land, the split PR pages may include earlier stack commits in their diffs. This reference PR gives reviewers and CI one clean full-stack view of the final merged shape produced by the split PR sequence.

## Speed-Bench Result
download

#141 adds a DGX Spark / GB10 speed-bench artifact comparing no-MTP against current exact MTP over 2048..65536 context frontiers with 128 greedy generation tokens per point.

| Run | Prefill mean | Generation mean | Generation first | Generation last |
| no-MTP | 342.20 t/s | 12.68 t/s | 14.00 t/s | 11.63 t/s |
| exact MTP | 341.02 t/s | 12.62 t/s | 12.53 t/s | 11.68 t/s |

The curves are effectively tied: exact MTP trails on the cold first row, then tracks no-MTP closely and is slightly ahead at the largest contexts.

## Validation

- Local published-branch merge-chain check passed cleanly through #135 -> #136 -> #137 -> #138 -> #139.
- GitHub reports #135-#140 as mergeable/clean.
- Local `make cpu` passed at this final stacked state.
- `python3 -m py_compile tests/ds4_proof.py tests/ds4_weight_server_harness_smoke.py tests/cuda_mtp_proof_matrix.py` passed.
- DGX Spark validation across the stack included `make cuda-spark`, `make cuda-regression`, `make cpu`, `./ds4_test --server`, proof-harness smoke tests, and real VMM shared-weight proof runs.
- The #137 prefill regression was fixed before this reference branch: patched default `ds4-bench` smoke measured `401.51 t/s` prefill and `14.05 t/s` generation on DGX Spark.
- #141 validation added proof-harness budget smoke checks, `ds4-bench` MTP support validation, and full no-MTP vs exact-MTP GB10 speed-bench sweeps.


Entrpi:
## Current Experimental State

The stack is reviewable as infrastructure plus CUDA MTP enablement, but the MTP performance work should still be treated as experimental.

The best validated result so far is that the harness and verifier tooling now let us measure MTP against both fast-MTP and no-MTP baselines without repeatedly paying full model load cost. The VMM weight server path was validated on DGX Spark with full base+MTP raw tensor sharing, and the proof runner can now report exact byte comparisons plus steady-state throughput.

For runtime MTP itself, the main conclusion is that exact MTP is no longer dominated by startup or measurement artifacts. It is dominated by the target verifier pass.

download

Results summary across 2k-64k context, 128 greedy generation tokens per point:

| Run | Prefill mean | Generation mean | Generation first | Generation last |
| no-MTP | 342.20 t/s | 12.68 t/s | 14.00 t/s | 11.63 t/s |
| exact MTP | 341.02 t/s | 12.62 t/s | 12.53 t/s | 11.68 t/s |

Currently MTP is basically a wash; the curves are effectively tied. Exact MTP trails on the first cold row, then tracks no-MTP closely and is slightly ahead at the largest contexts.

Validated wins:

- Q8 pair batching is useful for verifier-sized batches, but must stay limited to `n_tokens <= 2`. Applying it to normal long prefill regressed prefill from about `402 t/s` to `261 t/s`; after gating it, default prefill returned to `401.51 t/s`.
- The exact decode2 top2 output-head path is the cleanest exact CUDA-side win so far. It reduced forced decode2 verifier averages from roughly `145-148 ms` to `130-134 ms` while preserving exact byte parity against no-MTP.
- The acceptance gate helps low-acceptance prompts by falling back toward ordinary decode. On prose, exact MTP steady-state became nearly equivalent to no-MTP after warmup.
- The VMM weight server materially improves proof iteration time by keeping raw weights loaded across isolated workers.

What remains experimental:

- Non-strict / fast MTP can show speedups on favorable prompts, but outputs should be compared before treating a run as an exact speedup.
- Decode3, batch-first verifier, and candidate-certified logits are promising research paths, but not yet proven default wins.
- The reverted attention-output-B N2 Q8 path did not meet the proof bar as a default optimization; shadow infrastructure remains useful for further localization.

Best current leads:

1. **Purpose-built exact decode2 layer kernels.** The current verifier still runs a target-model layer pass over the drafted suffix. Profiling shows about `103 ms` of synchronized verifier layer work per cycle on GB10.
2. **Attention output projection.** This is one of the largest remaining single costs, about `27 ms` per verifier cycle.
3. **Routed MoE gate/up.** Routed MoE is about `30 ms` per verifier cycle, and most of that is gate/up rather than the already-optimized down/sum side.
4. **Q path folding.** The Q path is about `15 ms` per verifier cycle. There may still be savings by folding projection/RoPE/bookkeeping work for the fixed decode2 verifier shape.
5. **Candidate-certified logits.** This could be a major win if it can prove acceptance without full vocabulary logits in common cases, but it needs a higher proof bar before becoming default.

The practical reading is: this stack makes CUDA MTP safe and measurable, fixes the known prefill regression, and lands a few exact verifier improvements. The next performance frontier is reducing exact target verifier layer cost, especially attention output projection and MoE gate/up.

Issue #126, opened 2026-05-14
speed benchmark on M3 Max Macbook Pro with different fan control.


taozhiyuai:
Based on the 6 benchmark runs of deepseek-v4-flash on the M3 Max 128GB, the comparison between the two cooling strategies is unequivocal:

---

### 📊 Core Conclusion: **Maximum Fan Speed Dominates Auto Fan Control**

| Metric | Auto Fan (Runs 1–3) | Max Fan (Runs 4–6) | Improvement |
|--------|---------------------|---------------------|-------------|
| **Avg Prefill Speed** | 157.0 TPS | 178.2 TPS | **+13.5%** |
| **Avg Generation Speed** | 14.74 TPS | 16.73 TPS | **+13.5%** |
| **Prefill Variability (CV)** | 28.5% | 26.1% | More Stable |
| **Generation Variability (CV)** | 21.1% | 15.3% | More Stable |

---

### 🔍 Key Findings

**1. The Auto Fan Strategy Suffers from Severe Performance Jitter**
- In the **20K–40K context window range**, Runs 2 and 3 experienced a catastrophic performance cliff.
- The worst data point occurs at **ctx=28672**:
  - Run 2: Prefill crashed to **124.6 TPS** (down **40%** from Run 1’s 207.8), Gen crashed to **10.8 TPS** (down **44%**).
  - Run 3: Prefill 142.7 TPS, Gen 12.1 TPS.
- This indicates the auto fan fails to dissipate heat under sustained load, triggering aggressive thermal throttling.

**2. Max Fan Speed is Remarkably Consistent**
- Runs 4, 5, and 6 trace nearly identical curves for both prefill and generation speeds, degrading smoothly as context length increases.
- At the same critical point (ctx=28672), the three runs recorded 215.4, 216.0, and 224.0 TPS respectively—minimal variance.

**3. The "Sweet Spot" for Gains is the 25K–35K Context Range**
- This is where the auto fan’s thermal throttling is most severe, making the max fan advantage most pronounced:
  - Prefill peak gain: **+60.1 TPS (+38%)** at ctx=28672
  - Generation peak gain: **+5.5 TPS (+39%)** at ctx=28672

**4. Long-Context Gains Persist (70K+)**
- Even at 102K context length, max fan maintains a stable **11–15%** speed advantage over auto fan.

---

### 💡 Recommendation

For sustained inference workloads like deepseek-v4-flash on the M3 Max 128GB:

[...]

PR #103, opened 2026-05-12
Add MacBook Pro M5 Max (128 GB) Metal throughput report

lixiangnlp:
## Summary

Adds a five-run `ds4-bench` performance report for **MacBook Pro Apple M5 Max,
128 GB** (Metal, IQ2 weight-mostly), using the exact command from the README:

```sh
./ds4-bench -m ds4flash.gguf \
  --prompt-file bench/promessi_sposi.txt \
  --ctx-start 2048 --ctx-max 65536 \
  --step-incr 2048 --gen-tokens 128
```

Five back-to-back sweeps with a passive cool-down between runs to keep one
hot sweep from dominating the headline numbers. The PR ships:

- `bench/m5_max_128gb_report.md` — host/model setup, per-run summary, and a
  per-frontier table with mean ± σ across the five runs.
- `bench/results/run{1..5}.csv` + `bench/results/run{1..5}.log` — raw outputs
  so any aggregate in the report can be regenerated.

## Headline numbers (n = 5, Metal)

| ctx | prefill mean (σ) | gen mean (σ) |
|----:|-----------------:|-------------:|
| 2048 | 369.7 t/s (1.7) | 31.46 t/s (0.07) |
| 16384 | 261.9 t/s (6.8) | 28.50 t/s (0.80) |
| 32768 | 222.4 t/s (13.5) | 24.29 t/s (1.27) |
| 65536 | 161.5 t/s (5.9) | 20.30 t/s (1.13) |

Decode falls ~35% from 2k to 64k (31.5 → 20 t/s); prefill plateaus at ~160 t/s
past ~50k. Mid-context rows show the highest run-to-run variance — that window
of every sweep lands in the thermally-loaded phase, and run 4 specifically was
the hottest pass.

## Test plan

- [x] `./ds4-bench` invoked with the README long-context command, 5 times
- [x] Raw CSV/log captured for each run and committed under `bench/results/`
- [x] Aggregates in `bench/m5_max_128gb_report.md` regenerated from the
      committed CSVs

PR #97, opened 2026-05-12
Add M5 Max 40 GPU (128 GB) benchmark data and visualization

Chida82:
Thanks for the awesome work. Here is the benchmark for my M5 Max with CPU 18-Core and GPU 40-Core.

Hardware: Apple M5 Max with CPU 18-Core and GPU 40-Core, 128 GB unified memory. Metal backend, IQ2XXS w2Q2K imatrix GGUF.

Ran the command from speed-bench/README.md:

`
./ds4-bench \
  -m ds4flash.gguf \
  --prompt-file speed-bench/promessi_sposi.txt \
  --ctx-start 2048 \
  --ctx-max 65536 \
  --step-incr 2048 \
  --gen-tokens 128
  `
  
Then:

`python3 speed-bench/plot_speed.py speed-bench/m5_max_40_gpu.csv --title "M5 Max 40 Gpu t/s"`

other info:

ds4-bench: context buffers 1311.89 MiB (ctx=65665, backend=metal, prefill_chunk=2048, raw_kv_rows=2304, compressed_kv_rows=16418)
ds4: Metal device Apple M5 Max, 128.00 GiB RAM
ds4: requesting Metal residency (may take tens of seconds)... done
ds4: warming Metal model views... done
ds4: Metal model views created in 2.354 ms, residency requested in 277.150 ms, warmup 4.174 ms (mapped 82697.67 MiB from offset 5.08 MiB)
ds4: Metal mapped mmaped model as 2 overlapping shared buffers
ds4: metal backend initialized for graph diagnostics

PR #15, opened 2026-05-08
Add Metal 4 M5 prefill optimizations

ivanfioravanti:
## Summary
- enable M5-class Metal 4 MPP prefill paths for Q8_0 dense matmuls, attention-output low projection, and staged routed MoE projections
- promote correctness-gated routed MoE boundaries: down from layer 2, gate/up from layer 13
- add a fused six-expert routed MoE sum kernel for the common top-k=6 prefill shape
- keep experimental probes and escape hatches for ablation (`DS4_METAL_MPP_DISABLE=1`, `DS4_METAL_MOE_SUM6_DISABLE=1`, staged routed MoE envs)

## Benchmarks
Prompt source: `README.md`; command shape: `./ds4 --prompt-file <prompt> -n 1 --nothink --ctx 32768`; 3 repeats per target.

| target tokens | standard non-M5 avg tok/s | M5 no sum6 avg tok/s | M5 default avg tok/s | M5 speedup |
|---:|---:|---:|---:|---:|
| 512 | 236.97 | 390.29 | 389.47 | 1.64x |
| 2048 | 324.90 | 480.07 | 478.36 | 1.47x |
| 4096 | 289.97 | 454.41 | 455.52 | 1.57x |
| 8192 | 287.60 | 440.43 | 442.69 | 1.54x |
| 16384 | 284.26 | 427.62 | 427.78 | 1.50x |

8192-token routed MoE stage profile, current M5 default:

| stage | total ms | mean per layer ms | share |
|---|---:|---:|---:|
| up | 2443.497 | 11.365 | 35.1% |
| gate | 2422.942 | 11.269 | 34.8% |
| down | 1727.114 | 8.033 | 24.8% |
| activation_weight | 142.693 | 0.664 | 2.1% |
| sum | 137.482 | 0.639 | 2.0% |
| map | 68.601 | 0.319 | 1.0% |
| gate_up | 12.389 | 0.058 | 0.2% |

Disabling routed MoE MPP on the same profile drops prefill from 442.33 tok/s to 359.17 tok/s and raises gate/up/down to about 18.4-18.7 ms per layer.

## Validation
- `make ds4 ds4_test`
- `./ds4_test --metal-kernels`
- `./ds4_test --long-context`
- `./ds4_test --logprob-vectors`

## Notes
- Earlier `up` boundaries with `gate=13` were tested but failed long-context; gate/up stay paired at layer 13.
- A paired MPP gate+up matmul prototype compiled and ran but was slower overall, so it was not included.
[...]

PR #143, opened 2026-05-14
docs(bench): add M5 max

fry69:
This is a benchmark run on my M5 Max 128GB machine, against current `main` (https://github.com/antirez/ds4/commit/06105915b21088f375dbf60c8750c3ca067b7b58)

Apart from the standard benchmark settings (which took around 7 minutes), I also included a "4x" run (which took around 40 minutes), with all relevant number multiplied by 4:

```shell
./ds4-bench \
  -m ds4flash.gguf \
  --ctx-start 2048 \
  --ctx-max 262144 \
  --step-incr 8192 \
  --gen-tokens 512
```

This is IMHO a bit more illuminating than the base benchmark, where the lines stay mostly flat/linear, so it may be possible to read a bit more out of the diagram (correct me if I am wrong on this).

This is mostly for illustration purposes, feel free to close if this is not relevant.

Related PRs #103 (likely better/more comprehensive) and #97

---

Update:

I added two new runs with a fan control app ([Macs Fan Control](https://github.com/crystalidea/macs-fan-control)) setting the fans to "full blast". 

This seems indeed to improve the numbers and shorten the time for the 4x long run (down to ~35 minutes from ~45). Inspired by #126


VadimDu:
Thanks!
I was actually expecting a much higher speed-up in prefill tps compared to M4 Max (I remember Apple claims of x3-x4 increase...).
Looks like only 5-10% higher prefill. I guess further optimizations for the new Metal are warranted!

fry69:
> I was actually expecting a much higher speed-up in prefill tps compared to M4 Max

The M5 super powers currently do not get fully utilized. Have a look at #15 for M5 speedups, specifically for prefilling.

PR #146, opened 2026-05-14
server: guard disk KV cache with CRC32C

Ghatage:
bad disk snapshots are rejected before session state is restored, while existing v1 caches stay readable.

Write new KV cache files as v2 with CRC32C for header metadata, rendered text, DS4 payload, and tool map. Keep v1 reads and touches intact, and recompute v2 CRCs when hit counts or tool maps are rewritten.

Reject corrupt v2 entries during load and same-text compatibility checks. A broken cache file is unlinked and replaced by a fresh write.

Correctness: git diff --check; CPU-built ./ds4_test --server.
Correctness: non-CPU tests/ds4_test.c object compile.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -7357,11 +7357,16 @@ typedef struct {
     char sha[41];
     char *path;
     uint8_t quant_bits;
+    uint8_t format_version;
     uint8_t reason;
     uint32_t tokens;
     uint32_t hits;
     uint32_t ctx_size;
     uint8_t ext_flags;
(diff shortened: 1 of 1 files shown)


ucjonathan:
Did something in DS4 write out a corrupt KV cache file or is this preventative in case that happens in the future?

Ghatage:
ds4 did not corrupt anything... this is purely preventative.
Esp since now kv cache will be written to disk, checksums are table stakes to validate data integrity

PR #148, opened 2026-05-14
server: add tool-safe directional steering policy

audreyt:
## Summary

This adds server-side directional steering policies for tool-aware deployments:

```sh
--dir-steering-policy final-answer   # default
--dir-steering-policy decoding
--dir-steering-policy always
--dir-steering-policy off
```

`final-answer` is now the default policy, following maintainer feedback on [antirez/ds4#148](https://github.com/antirez/ds4/pull/148). It keeps prompt prefill, thinking tokens, and DSML/tool-call grammar unsteered, then re-enables steering once generation has clearly entered final natural-language answer text.

`decoding` is the middle-ground policy requested in review: prompt/prefill is unsteered, but every generated token is steered, including thinking and tool-call syntax. `always` restores the previous always-on behavior, and `off` disables steering at the server policy layer.

## Why

Directional steering is useful for behavior/style/topic control, but applying it while the model is emitting tool-call syntax can perturb DSML grammar, tool arguments, or Responses/Anthropic tool protocol structure.

For tool-using agents, the safer default is:

- no steering during prefill
- no steering during hidden thinking
- no steering inside DSML/tool-call syntax
- steering only for final visible prose

This lets deployments use steering for final-answer behavior without making tool calls less reliable.

## Changes

- Adds per-session directional steering overrides so the server can toggle steering scales dynamically.
- Defaults `ds4-server` directional steering policy to `final-answer`.
- Adds `--dir-steering-policy decoding`, which disables steering only during prompt/prefill.
- Keeps `always` available for the original behavior and `off` for policy-layer disabling.
- Tracks thinking state and DSML decode state during generation.
- Avoids steering partial tool-call starts and tool-call bodies in `final-answer` mode.
- Uses non-MTP eval while dynamic `final-answer` steering is active, so draft tokens do not cross steering-state boundaries.
- Documents the policies in `README.md` and `dir-steering/README.md`.
- Adds server unit coverage for default policy selection, `decoding`, and final-answer/tool-safe behavior.

## Compatibility

The server default changes from `always` to `final-answer`. Existing deployments that want exact previous behavior can pass:

```sh
--dir-steering-policy always
```

The core steering scales and file format are unchanged.

## Testing

```sh
make ds4_test
./ds4_test --server
make ds4-server
./ds4-server --help | rg -n "dir-steering-policy|final-answer|decoding|always|off"
DS4_TEST_MODEL=/Users/au/w/ds4/gguf/DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix.gguf make test
```

The first full-suite attempt used my local abliterated/aligned `ds4flash.gguf` symlink and failed the official logprob-vector fixture, as expected for a different GGUF. Re-running with the non-abliterated imatrix GGUF used for the official vectors passes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -719,6 +719,13 @@ and so forth, much faster than fine-tuning.
 This is also useful for cybersecurity researchers who want to reduce a model's
 willingness to provide dual-use or offensive security guidance.
 
+For `ds4-server`, directional steering defaults to the tool-safe
+`final-answer` policy: prompt prefill, thinking tokens, and DSML tool-call
+syntax stay unsteered, while final visible answer prose uses the configured
+direction. Use `--dir-steering-policy decoding` to leave only prefill
+unsteered, `always` for the original always-on behavior, or `off` to disable
+server-side steering.
+
 ## Test Vectors
 
 `tests/test-vectors` contains short and long-context continuation vectors
--- a/dir-steering/README.md
+++ b/dir-steering/README.md
@@ -17,18 +17,25 @@ With no steering file or zero scales, ds4 follows the normal inference path.
 --dir-steering-file FILE   load a 43 x 4096 f32 direction file
 --dir-steering-ffn F       apply steering after FFN outputs; default is 1 when a file is provided
 --dir-steering-attn F      apply steering after attention outputs; default is 0
---dir-steering-policy MODE server-only policy: always, final-answer, or off; default is always
+--dir-steering-policy MODE server-only policy: final-answer, decoding, always, or off; default is final-answer
 ```
 
 The FFN output is usually the best first target because it is late enough in
 each layer to represent behavior, style, and topic signals. Attention steering
 is available for experiments, but it can be more fragile.
 
-For tool-using agents, `ds4-server --dir-steering-policy final-answer` keeps
-prompt prefill, thinking tokens, and DSML tool-call tokens unsteered. Steering
-is re-enabled only after generation has clearly entered final natural-language
-answer text. This avoids letting a behavior/style vector perturb tool-call
-grammar while still allowing the final prose to use the configured direction.
+For tool-using agents, `ds4-server` defaults to `--dir-steering-policy
+final-answer`. This keeps prompt prefill, thinking tokens, and DSML tool-call
+tokens unsteered. Steering is re-enabled only after generation has clearly
+entered final natural-language answer text. This avoids letting a
+behavior/style vector perturb tool-call grammar while still allowing the final
+prose to use the configured direction.
+
+`--dir-steering-policy decoding` is a middle ground for experiments that should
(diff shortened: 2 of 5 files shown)


antirez:
That's **brilliant**! Thank you so much. Going to merge ASAP.

antirez:
@audreyt since I guess you already tested it, are we sure we don't want the final policy to be the default?

antirez:
Also, what about an additional `decoding` policy that disables steering only during prefill? Sorry for the many comments.

audreyt:
> Also, what about an additional `decoding` policy that disables steering only during prefill? Sorry for the many comments.

Done, thank you for the nudge!

I made `final-answer` the default, and added the `decoding` policy too:

- `final-answer` now defaults to no steering during prefill/thinking/tool-call grammar, then steering for final visible prose.
- `decoding` disables steering only during prefill, then steers all generated tokens.
- `always` is still available for the previous behavior.
- `off` remains the policy-layer disable switch.

I also rebased onto current `main`, marked the PR ready for review, and updated the PR description with the compatibility note and test results. It's now working beautifully in my OpenClaw instance (@jdd-kami).

yurivish:
Out of curiosity, would it make any sense to have a policy that steers all "thinking" outside of tool calls (somewhere between `decoding` and `final-answer`)?

Intuitively, this could potentially keep tool calls reliable while maximizing the steered proportion of "natural language" thought.

antirez:
I played with different modes, prefill also has a big impact about how much refusal you get. Completing the tests to merge the PR, but likely the initial @audreyt default was sane, because not steering prefill leaves too many refusals if the questions are "strong" enough. Will update you soon.

antirez:
Oh, and: yet another possibility. It is possible to calculate the component of the vector towards the refusal (or any vector) direction and apply the vector only when the refusal is detected. I wonder if/how this works.

antirez:
More precisely, the possible rules are:

1. What we do right now: dot product (vector, activations) * vector, so that we try to "undo" the component in that direction.
2. Something with a threshold (not implemented): check the dot product to see how much component we have, and when we see the component is over a given threshold, apply a fixed steering.
3. Additive steering. Always do activations += vector.

I believe what we do right now is the most interesting but... There are cool uses at least for "3", and "2" could preserve quality of inference in a more conservative way. About "3", imagine you want a model that always focuses on certain ideas for an artistic experiment, that would be quite interesting to see. Like the Anthropic experiment where the model would always try to talk about the Golden Bridge.

antirez:
Of course "2" was already done: CAST: conditional activation steering; apply steering only when activations indicate the relevant context. Source: Lee/IBM, 2024 (https://arxiv.org/abs/2409.05907)

PR #158, opened 2026-05-15
cuda: fix HMM path on coherent unified-memory systems (GB10 NVLink-C2C)

Mikehutu:
## Problem

On systems where `cudaHostRegister()` returns `cudaErrorNotSupported` (e.g.
NVIDIA Grace-Blackwell GB10 / DGX Spark with NVLink-C2C unified memory), the
`else` branch in `ds4_gpu_set_model_map()` only logged the error and returned.
`g_model_hmm_direct` was never set, so `cuda_model_range_is_cached()` returned
0 for every tensor range.

This caused `cuda_model_range_ptr()` to fall through to the per-range
`cudaMalloc` + `cudaMemcpy` path on every first tensor access — silently
allocating ~87 GB of redundant device copies of the 80.76 GiB model:

```
# Without fix, on GB10 running ds4_test:
free -h:       Mem: 121Gi used 98Gi, available 23Gi
nvidia-smi:    ds4_test   87,992 MiB
```

The model weights end up in CUDA device-private pools (not evictable, invisible
to `/proc/PID/smaps`, only visible via `MemFree` shrinking and nvidia-smi's
per-process column). A second model cannot be co-loaded.

## Root cause

`cuda_model_prefetch_range()` was fully implemented and correct — but its only
call site on Linux/CUDA was behind the `DS4_CUDA_COPY_MODEL_CHUNKED` env var.
On a normal GB10 run:
1. `DS4_CUDA_COPY_MODEL_CHUNKED` is not set
2. `cudaHostRegister` fails with `cudaErrorNotSupported`
3. The `else` branch clears the error and returns — prefetch never called
4. `g_model_hmm_direct` stays 0 → per-range copy runs for every tensor

## Fix

Two hunks, 8 lines total:

**Hunk 1** — `ds4_gpu_set_model_map()`: when registration is skipped, call
`cuda_model_prefetch_range()` which issues `cudaMemPrefetchAsync` over the
full mmap region and sets `g_model_hmm_direct = 1` on success.

**Hunk 2** — `cuda_model_range_is_cached()`: short-circuit to 1 when
`g_model_hmm_direct` is set, so the per-range copy path is never reached
(the existing check on line ~200 in `cuda_model_range_ptr` is not reached
because `is_cached` runs first).

## Results on DGX Spark GB10

Hardware: SM_121, 128 GB unified LPDDR5x, NVLink-C2C ~900 GB/s  
Model: DeepSeek V4 Flash IQ2_XXS-w2Q2K (80.76 GiB)  
CUDA 13.0, driver 580.142

| Metric | Before | After |
|--------|--------|-------|
| `free -h` used | ~98 GB | ~26 GB |
| `free -h` available | ~23 GB | ~95 GB |
| nvidia-smi process | 87,992 MiB | ~26,000 MiB |
| Weight memory type | CUDA device allocs (locked) | File page cache (reclaimable) |
| GPU utilization | ~50% (copy stalls) | 94% |
| Throughput | degraded | **13.81 t/s** |
| Second model co-loadable | No | Yes |

## Testing

Manually verified on GB10. Full `make cuda-regression` suite not yet run —
happy to do so if you can confirm there's no known issue running the suite on
SM_121 / CUDA 13.0.

### DIFF
--- a/ds4_cuda.cu
+++ b/ds4_cuda.cu
@@ -305,6 +305,8 @@ static const char *cuda_model_range_ptr(const void *model_map, uint64_t offset, 
 static int cuda_model_range_is_cached(const void *model_map, uint64_t offset, uint64_t bytes) {
     if (bytes == 0) return 1;
     if (g_model_device_owned || g_model_registered) return 1;
+    /* On the HMM direct path every range is immediately accessible. */
+    if (g_model_hmm_direct) return 1;
 
     const uint64_t end = offset + bytes;
     if (end < offset) return 0;
@@ -1487,6 +1489,12 @@ extern "C" int ds4_gpu_set_model_map(const void *model_map, uint64_t model_size)
     } else {
         fprintf(stderr, "ds4: CUDA host registration skipped: %s\n", cudaGetErrorString(err));
         (void)cudaGetLastError();
+        /* On coherent unified memory systems (e.g. Grace-Blackwell NVLink-C2C) the
+         * host mmap pointer is directly addressable by the GPU via HMM page faults.
+         * Queue an async prefetch of the full model so pages are resident before
+         * inference starts, avoiding per-kernel page-fault stalls.  The function
+         * also sets g_model_hmm_direct=1 on success. */
+        (void)cuda_model_prefetch_range(model_map, model_size, 0, model_size);
     }
     return 1;
 }


Mikehutu:
**Regression test results — GB10 (SM_121, CUDA 13.0, driver 580.142)**

```
$ make cuda-spark && make cuda-regression
./tests/cuda_long_context_smoke
ds4: CUDA backend initialized on NVIDIA GB10 (sm_121)
cuda-regression: top-k n_comp=32768 n_tokens=32 elapsed=0.009s
```

Passes. First cold run took 2.13s (CUDA JIT compiling the top-k kernel to SASS for the first time); subsequent runs are ~0.009s once the compute cache is warm. Not related to this patch — the top-k kernel is untouched.

ttsiodras:
Was about to test antirez's creation on a SPARK - but noticed your MR and it seemed significant.
I just built it:

```
    $ ./ds4-bench -m gguf/DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix.gguf --prompt-file speed-bench/promessi_sposi.txt --ctx-start 2048   --ctx-max 65536   --step-incr 2048   --gen-tokens 128    
    
    ds4-bench: context buffers 1311.89 MiB (ctx=65665, backend=cuda, prefill_chunk=2048, raw_kv_rows=2304, compressed_kv_rows=16418)
    ds4: CUDA backend initialized on NVIDIA GB10 (sm_121)
    ds4: CUDA host registration skipped: operation not supported
    ds4: CUDA ATS/HMM prefetch queued 80.76 GiB of model tensors in 69.580s
    
    ds4: CUDA startup model cache prepared 80.76 GiB of tensor spans in 0.000s
    ds4: cuda backend initialized for graph diagnostics
    ctx_tokens,prefill_tokens,prefill_tps,gen_tokens,gen_tps,kvcache_bytes
    2048,2048,405.37,128,12.65,52184460
    4096,2048,407.91,128,12.56,80373132
    6144,2048,401.39,128,12.52,108561804
    8192,2048,398.03,128,12.40,136750476
    10240,2048,391.36,128,12.34,164939148
    12288,2048,387.61,128,12.31,193127820
    14336,2048,383.22,128,12.26,221316492
    
    ──────────────────────────────────────────────────────────────────────────────
    # free -h
                   total        used        free      shared  buff/cache   available
    Mem:           121Gi        18Gi        19Gi        44Mi        84Gi       103Gi
    Swap:           15Gi        29Mi        15Gi
    
    # nvidia-smi
    Sat May 16 12:35:10 2026       
    +-----------------------------------------------------------------------------------------+
    | NVIDIA-SMI 580.142                Driver Version: 580.142        CUDA Version: 13.0     |
    +-----------------------------------------+------------------------+----------------------+
    | GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
    | Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
    |                                         |                        |               MIG M. |
    |=========================================+========================+======================|
    |   0  NVIDIA GB10                    On  |   0000000F:01:00.0 Off |                  N/A |
    | N/A   72C    P0             91W /  N/A  | Not Supported          |     96%      Default |
    |                                         |                        |                  N/A |
    +-----------------------------------------+------------------------+----------------------+
    
    +-----------------------------------------------------------------------------------------+
    | Processes:                                                                              |
    |  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
    |        ID   ID                                                               Usage      |
    |=========================================================================================|
    |    0   N/A  N/A           55799      C   ./ds4-bench                           14536MiB |
    +-----------------------------------------------------------------------------------------+
```    
Is that right? 

GPU memory only at 14.536GB?

Mikehutu:
Hi. I used the full ctx (1M) with 23Gb memory, if you are using less,  i suppose you got it working. The GPU utilization was 94%.

I got prefill 300ts and output 13-14ts. 

Did you run any inference yet? 

The model weights (80gb) are in page cache, so technically it is using the memory, even when it is not showing it in smi if im correct. 

Next experiment would be to test with second model loaded and see what happens..

ttsiodras:
Yes, I did run inference - got similar numbers to what you see in the output from ds4-bench above (~12.5 tg/s, 300+ pp/s). 

I have access to a pair of SPARKs, so to be completely honest... that performance is not enough for me. Esp the pp one - which is sine-qua-non for supervised agentic iterations... 

Even though I appreciate the work Antirez did to create this, I will continue waiting for a recipe from https://github.com/eugr/spark-vllm-docker . So far there's one MR ( https://github.com/eugr/spark-vllm-docker/pull/219 ) that seems the most promising. I basically just  want to make use of the power of both machines...
 
...or some other miracle will popup ( https://github.com/chiennv2000/orthrus ) and give us an order of magnitude speedup, making even single machines enough. It keeps raining innovations in this field :-)

PR #177, opened 2026-05-17
fix(kv-cache): boost eviction score for cold/evict/shutdown checkpoints

### LINKED ISSUE #176: Disk KV cache: cold/evict checkpoints get no eviction-score advantage over continued
## TL;DR

`kv_entry_eviction_score` is `(hits+1) * tokens / size` regardless of `reason`. A freshly written cold or evict checkpoint loses the eviction race to a competing workload's fresh continued snapshots on density alone. In production a 1.3 GiB live-state save died within ~5 min while a second conversation prefilled.

## How it fails today

When conversation B writes a continued, A will have a cold, an evict, & a continued on disk.
The score function gives all three the same weight, so the victim depends on incidental density differences and not on intent. If cold & evict were boosted, the continued would be the consistent target.
Then if conversation A resumes right where it left off the evict would cache hit and a significant amount of prefill time would be avoided.

In a production case the live-state evict was 1.3 GiB and got unlinked after 5m 19s while B prefilled.

## Proposed fix

Bump the base from `1.0` to `4.0` for `KV_REASON_COLD | EVICT | SHUTDOWN` in `kv_entry_eviction_score`. Continued and unknown stay at `1.0`. A continued must serve three real hits before outranking a fresh same-shape checkpoint. Existing hit decay still applies.

unsaltedbutter-ai:
Fixes #176.

`kv_entry_eviction_score` uses `(hits+1) * tokens / size` regardless of `reason`, so a fresh cold or evict checkpoint loses the eviction race to a competing workload's fresh continued snapshots on density alone. The `protected_sha` DBL_MAX only protects the just-stored entry for one round; after that, explicit checkpoints have nothing distinguishing them from routine continued traffic. In production a 1.3 GiB live-state save died within ~5 min while a single competing prefill ran (issue #176 has the log).

## Change

Bump the base from `1.0` to `4.0` for `KV_REASON_COLD | EVICT | SHUTDOWN`. Continued and unknown stay at `1.0`. The boost is intrinsic to the file (`reason` byte in the header) so it persists across conversations and restarts. Existing hit-decay logic still applies.

## Test plan

- [ ] `./ds4_test --server` passes (existing eviction tests untouched)
- [ ] New `test_kv_cache_eviction_boosts_cold_over_continued`: a fresh cold survives a denser fresh continued under a budget that fits exactly one
- [ ] New `test_kv_cache_eviction_boosts_evict_over_continued`: same for evict
- [ ] Manual: long conversation, brief interrupt, resume; resume hits the saved evict instead of re-prefilling from zero


unsaltedbutter-ai:
closing as fixed with ec6a82aaa9a3a98bee803801440b0e803aa48f09

PR #186, opened 2026-05-18
Compressed KV disk cache via streaming LZ4

jasontitus:
## Summary

Add optional LZ4 compression to the disk KV cache, shared by `ds4-server` and `ds4-agent`. DeepSeek V4 Flash checkpoints shrink 2.0x to 2.9x and V4.1 checkpoints 2.8x to 3.1x, so the same `--kv-disk-space-mb` holds about three times as many. Compression adds at most 63 ms to a load and 287 ms to a save up to 60K tokens, under 1% of the request that stores or loads the checkpoint. Inference code is unchanged.

Chunks are compressed in parallel with `LZ4_compress_HC` level 1 after a byte-4 transpose, and each chunk carries an XXH32 of its raw bytes that is verified before the engine sees them. Raw files are written exactly as before, and a payload that compression shrinks by less than 1/64 is stored raw.

## What's in the change

- `ds4_kvstore.c`: writer and reader cores. The reader validates framing before allocating and verifies each chunk checksum. Loads read through a cookie stream where available and through a temporary file elsewhere.
- A checkpoint is removed only when its stored bytes are proven wrong: bad framing, a checksum mismatch, or a payload past the end of the file. Engine, GPU, allocation and I/O failures keep the file.
- Raw files stay version 1. Compressed files are version 2 with the codec in header byte 21 and log2 chunk size in byte 22, so older binaries reject them instead of clearing the codec byte.
- `--kv-cache-compression-threads N` (`0` writes raw), default `min(8, online cpus)`; `DS4_KV_CACHE_COMPRESSION_THREADS` sets the default for both binaries.
- Vendored lz4 v1.10.0 (`lz4.c`, `lz4hc.c`, BSD-2-Clause).
- `make test-kv-lz4` and `make test-kv-lz4-nofwrap`, both in `make test`; the second builds without cookie streams.
- `misc/COMPRESSED_KV_CACHE.md`: format, failure handling and the full measurements.

## Limitations

- Raw payloads carry no content checksum, as before.
- Without cookie streams, a load needs temporary space equal to the uncompressed payload.
- Agent sessions saved by unreleased builds of this PR use a retired layout and must be removed by hand.
- Qwen3.8 Flash Next checkpoints do not compress (about 1.00x) and are stored raw, but still pay the encode attempt: at 10K tokens a save took 615 to 651 ms against 287 to 442 ms with `--kv-cache-compression-threads 0`.

## Testing

Through `ds4-server` on M1 Ultra (V4 Flash, V4.1 Q2, Qwen3.8 Q2) and M5 Max (V4 Flash, V4.1 Q4), base `9139e2a`. Each prompt was sent twice, raw and then LZ4.

| Machine | Model | Tokens | Raw / LZ4 size | Save raw / LZ4 | Load raw / LZ4 |
|---|---|---:|---:|---:|---:|
| M1 Ultra | V4 Flash | 20,480 | 291.8 / 116.4 MiB (2.51x) | 287 / 312 ms | 115.6 / 119.6 ms |
| M5 Max | V4 Flash | 40,960 | 560.7 / 212.2 MiB (2.64x) | 71 / 292 ms | 75.1 / 138.4 ms |
| M5 Max | V4.1 Q4 | 4,096 | 35.5 / 12.9 MiB (2.76x) | 16 / 62 ms | 11.2 / 20.0 ms |
| M5 Max | V4.1 Q4 | 21,878 | 144.2 / 47.3 MiB (3.05x) | 57 / 99 ms | |

To test the checksum, the test deliberately flipped 8 bytes in a saved checkpoint; the next load reported it corrupt, removed it and recomputed, on both DeepSeek models on both machines. No corruption occurred outside these injected tests. The complete tables and setup are in `misc/COMPRESSED_KV_CACHE.md`.

```sh
make test test-kv-lz4 test-kv-lz4-nofwrap test-frontends
```

<details>
<summary>V4.1 Q4 throughput on M5 Max (reference; ds4-bench does not use the disk cache)</summary>

SSD streaming, 32 GB expert cache, incremental prefill of `speed-bench/promessi_sposi.txt`, 64 generated tokens, single run.

| Context tokens | Prefill tokens/s | Generation tokens/s |
|---:|---:|---:|
| 16,384 | 415.88 | 7.00 |
| 32,768 | 384.74 | 6.78 |
| 65,536 | 468.10 | 6.78 |
| 131,072 | 464.47 | 6.10 |
| 262,144 | 412.82 | 6.08 |

</details>

https://claude.ai/code/session_01RWat8R8nkWa1Kzbrhaxxuu

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,8 +77,8 @@ help:
 ds4: ds4_cli.o ds4_help.o linenoise.o ds4_gpu_args.o $(CORE_OBJS)
 	$(CC) $(CFLAGS) -o $@ ds4_cli.o ds4_help.o linenoise.o ds4_gpu_args.o $(CORE_OBJS) $(METAL_LDLIBS)
 
-ds4-server: ds4_server.o ds4_help.o ds4_kvstore.o rax.o ds4_gpu_args.o lz4.o $(CORE_OBJS)
-	$(CC) $(CFLAGS) -o $@ ds4_server.o ds4_help.o ds4_kvstore.o rax.o ds4_gpu_args.o lz4.o $(CORE_OBJS) $(METAL_LDLIBS)
+ds4-server: ds4_server.o ds4_help.o ds4_kvstore.o rax.o ds4_gpu_args.o lz4.o lz4hc.o $(CORE_OBJS)
+	$(CC) $(CFLAGS) -o $@ ds4_server.o ds4_help.o ds4_kvstore.o rax.o ds4_gpu_args.o lz4.o lz4hc.o $(CORE_OBJS) $(METAL_LDLIBS)
 
 ds4-bench: ds4_bench.o ds4_help.o ds4_gpu_args.o $(CORE_OBJS)
 	$(CC) $(CFLAGS) -o $@ ds4_bench.o ds4_help.o ds4_gpu_args.o $(CORE_OBJS) $(METAL_LDLIBS)
@@ -86,8 +86,8 @@ ds4-bench: ds4_bench.o ds4_help.o ds4_gpu_args.o $(CORE_OBJS)
 ds4-eval: ds4_eval.o ds4_help.o $(CORE_OBJS)
 	$(CC) $(CFLAGS) -o $@ ds4_eval.o ds4_help.o $(CORE_OBJS) $(METAL_LDLIBS)
 
-ds4-agent: ds4_agent.o ds4_help.o ds4_web.o ds4_kvstore.o linenoise.o ds4_gpu_args.o lz4.o $(CORE_OBJS)
-	$(CC) $(CFLAGS) -o $@ ds4_agent.o ds4_help.o ds4_web.o ds4_kvstore.o linenoise.o ds4_gpu_args.o lz4.o $(CORE_OBJS) $(METAL_LDLIBS)
+ds4-agent: ds4_agent.o ds4_help.o ds4_web.o ds4_kvstore.o linenoise.o ds4_gpu_args.o lz4.o lz4hc.o $(CORE_OBJS)
+	$(CC) $(CFLAGS) -o $@ ds4_agent.o ds4_help.o ds4_web.o ds4_kvstore.o linenoise.o ds4_gpu_args.o lz4.o lz4hc.o $(CORE_OBJS) $(METAL_LDLIBS)
 
 gguf-tools/quality-testing/score_official: gguf-tools/quality-testing/score_official.c ds4.h $(CORE_OBJS) rax.o ds4_gpu_args.o
 	$(CC) $(QUALITY_CFLAGS) -I. -o $@ gguf-tools/quality-testing/score_official.c $(CORE_OBJS) rax.o ds4_gpu_args.o $(METAL_LDLIBS)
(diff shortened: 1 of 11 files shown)


jasontitus:
Rebased onto the new agent changes.

jasontitus:
After the most recent KV cache changes we are now getting 1.8x compression across cache sizes.  This seems like something worth including.

jasontitus:
Added bit shuffling and we now get 2.5x compression for a tiny bit more compute.

jasontitus:
Rebased onto current `main` (b7e9f00) and re-verified against DeepSeek V4 Flash 0731.

**Rebase:** the four commits now apply linearly on top of current head. The only conflicts were in the `Makefile` (the build rules were refactored upstream since this PR's base); resolved by adding `lz4.o`/`lz4hc.o` to the `ds4-agent`/`ds4-server` link lines and their compile rules against the current layout. All feature code (`ds4_kvstore.c`, `ds4_agent.c`, `ds4_server.c`, lz4 codec) applied without conflict.

**Verified on DeepSeek V4 Flash (Apple M5 Max, Metal):**
- Builds clean.
- `sysprompt.kv` is written as `version=2`, `codec=1 (LZ4)`, `chunk_log2=24` (16 MiB) — matching the format spec.
- Round-trips correctly: cold write + warm restore both produce coherent output (the version-2 header + compressed payload decompress and reload without corruption).

**Compression ratio** on a 16384-ctx Flash KV cache (uncompressed baseline via `DS4_KV_CACHE_COMPRESSION_THREADS=0`):

| | size |
|---|---|
| uncompressed (`codec=NONE`) | 50.0 MB |
| compressed (`codec=LZ4`) | 22.4 MB |

**2.13× (53% saved).**

jasontitus:
Updated onto current main `9139e2a`, which includes DeepSeek V4.1 for Metal and CUDA and Qwen3.8 Flash Next.

Changes since the last revision:
- Each compressed chunk now carries an XXH32 of its raw bytes. Without it, damage to a compressed file could decode to full-length wrong bytes: in a fuzz test that flipped single bits in synthetic regions, about a third decoded that way.
- Raw files are version 1 again, so current binaries keep reading and refreshing them. Only compressed files are version 2.
- Compression no longer needs cookie streams, and loads fall back to a temporary file where they are missing.
- Checkpoints are deleted only when their bytes are proven wrong. A silent engine or GPU failure no longer removes a valid file.
- A payload that compression shrinks by less than 1/64 is rewritten raw, so near-incompressible checkpoints such as Qwen3.8's stay raw instead of paying chunk decoding for nothing.
- The reader validates framing before allocating, instead of sizing scratch from an unverified header.
- Fixed link failures in `ds4_test`, `ds4_agent_test` and the CPU and CUDA test binaries.
- New `test-kv-lz4` and `test-kv-lz4-nofwrap` targets in `make test`.

Tested through `ds4-server` on M1 Ultra with DeepSeek V4 Flash, V4.1 and Qwen3.8 Flash Next, and on M5 Max with V4 Flash and V4.1 Q4. V4 Flash checkpoints shrink 2.0x to 2.9x and V4.1 checkpoints 2.8x to 3.1x, so the same disk budget holds about three times as many. Compression adds at most 63 ms to a load and 287 ms to a save up to 60K tokens, under 1% of the request that stores or loads the checkpoint. A checkpoint with deliberately flipped bytes was detected, removed and recomputed on both DeepSeek models on both machines; no corruption occurred outside these injected tests. Qwen checkpoints do not compress and are stored raw. Inference code is unchanged relative to main. The tables are in the description.

emilianbold:
I almost wish the KV disk cache would just be a clear API boundary and we can drop in different implementations.

jasontitus:
I could see that. It is surprising how different the KV caches are for different models. Deepseek seems to really be sticking to an easy to compress version though. 

What other kinds of things would you want to do with kv cache storage?

emilianbold:
> What other kinds of things would you want to do with kv cache storage?

It kinda seems a neglected area of ds4 because the focus is on models and tok/s. And it has issues that don't get noticed (eg. see my draft PR #1058), including the fact that the API itself is not clear about what a "key" even represents.

Issue #167, opened 2026-05-16
DSML tool_calls XML leaking into reasoning_content/thinking in streaming responses


wanderingmeow:
## Problem

When the model emits DSML tool calls inside a thinking block (before `</think>` closes), the raw
 `<｜DSML｜tool_calls>` XML leaks into `reasoning_content`/`thinking` in the client session. The structured `toolCall` is extracted correctly, but the raw XML also appears as thinking content, causing duplication.

Example from a live session json from Pi.dev agent harness:

<details>

```json
{
    "type": "message",
    "id": "b2f4e976",
    "parentId": "cc7fdf27",
    "timestamp": "2026-05-16T05:55:54.300Z",
    "message":
    {
        "role": "assistant",
        "content": [
        {
            "type": "thinking",
            "thinking": "\n\n<｜DSML｜tool_calls>\n<｜DSML｜invoke name=\"bash\">\n<｜DSML｜parameter name=\"command\" string=\"true\">sed -n '7,9p' main.go</｜DSML｜parameter>\n</｜DSML｜invoke>\n</｜DSML｜tool_calls>",
            "thinkingSignature": "reasoning_content"
        },
        {
            "type": "toolCall",
            "id": "call_b4b3cbd94d4a0b69c20583ce56c032bc",
            "name": "bash",
            "arguments":
            {
                "command": "sed -n '7,9p' main.go"
            }
        }],
        "api": "openai-completions",
        "provider": "ds4",
        "model": "deepseek-v4-flash",
        "usage":
        {
            "input": 27,
            "output": 70,
            "cacheRead": 61706,
            "cacheWrite": 27,
            "totalTokens": 61830
        },
        "stopReason": "toolUse",
        "timestamp": 1778910950189,
        "responseId": "chatcmpl-49"
    }
}
```

</details>

The tool call itself is correctly parsed and emitted separately as a toolCall. The issue is the raw XML also appearing in the thinking field.

## Reproduction

Not reliably reproducible with a fixed prompt. Occurs during normal Pi.dev agent sessions (thinking → tool calls → tool results → regen loop). More common with longer conversations and tool-heavy workflows (e.g., file editing, shell commands).

## Impact

- Client session stores raw DSML XML as thinking content (duplication)
- Context is reprocessed from the start in the next turn due to KV cache miss: `ds4-server: live kv cache miss live=61831 prompt=61927 common=61830`

## Environment

- [Pi.dev](https://github.com/earendil-works/pi-coding-agent) harness
[...]

PR #192, opened 2026-05-18
fix: resilient DSML tool-call parsing and repair in long contexts

jackygurui:
### Summary

Long-context and long-generation scenarios expose several failure modes in DSML tool-call parsing. The model's attention degrades over thousands of tokens, producing malformed DSML that causes `finish=error` and breaks the agent loop. This PR adds a robust repair layer (`try_repair_dsml`) that handles all observed failure modes, plus targeted fixes for false positives and visibility improvements.

### Problem

Three failure modes were observed in stress testing (256K context, q4-imatrix):

| Mode | Description | Symptom |
|------|-------------|---------|
| **Unterminated** | Model stops mid-DSML, missing closing tags | `finish=error`, "unterminated tool call" |
| **Malformed closed** | Outer tags balanced but inner tags broken | `finish=error`, "invalid tool call" |
| **Hallucinated** | `<tool_calls>...</tool_calls>` wrapping plain reasoning text | False tool call detection |

Before this fix, all three modes would land in `finish=error`, silently aborting the agent loop. Additionally, DSML tags mentioned inside `<thinking>` text (e.g. the model explaining tool syntax in its reasoning) were being counted by the tag scanner in `try_repair_dsml`, causing false positive repairs.

### Changes

1. **`try_repair_dsml()` — single-pass DSML repair** (2 commits)
   - Rewritten from 6-pass O(6n) to single-pass O(n) tag counting for all 6 tag types
   - **Unterminated repair**: appends missing closing tags in correct nesting order (`parameter → invoke → tool_calls`)
   - **Hallucinated repair**: when `tool_calls` tags exist but contain no `<invoke>`, strips the orphaned tags so the text is treated as plain content
   - **Orphan end tag guard**: returns early when closing tags outnumber opening tags (prevents `size_t` underflow)
   - Integrated into both the unterminated-tool-call path *and* the invalid-tool-call recovery path
   - Added 10 unit tests covering all 3 DSML styles, unterminated, malformed, hallucinated, and balanced cases

2. **Ignore DSML inside `<thinking>`** (1 commit)
   - `try_repair_dsml` now finds the last `</thinking>` boundary and scans for DSML tags only from there — deliberately matching the same strategy used by `parse_generated_message_ex`, which was introduced in commit [`037ee39`](https://github.com/antirez/ds4/commit/037ee39aa638e4bca0bcade8c021091ffd55ccba) ("Ignore tool calls emitted inside thinking") to fix the same class of false positives reported in [#167](https://github.com/antirez/ds4/issues/167)
   - The hallucinated tag-stripping path also preserves the thinking section verbatim, only stripping tags from the post-thinking region

3. **Debug visibility** (1 commit)
   - Logs a `stderr` message when `require_thinking_closed=true` is triggered but the model never closes `<thinking>`

### Results

Before this fix, the 38 repaired + 3 orphan + 2 invalid cases would all have been counted as invalid, with every occurrence causing a hard `finish=error`. After the fix:

```
Tool calls: 169 | Invalid: 2 | Repaired: 38 | Orphan: 3
```

Only 2 cases remain unrecoverable; the other 41 are gracefully handled (38 repaired, 3 hallucinated-tags stripped). 100% repair success rate on unterminated tool calls.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -4607,17 +4607,25 @@ static bool parse_generated_message_ex(const char *text, bool require_thinking_c
 static bool try_repair_dsml(const char *s, size_t len, buf *out) {
     if (!s || !len) return false;
 
+    /* Only scan DSML tags after the last </thinking>.  DSML mentioned inside
+     * reasoning is not executable — it inflates tag counts and causes false
+     * positive repairs.  If no </thinking> is found, scan from the start
+     * (thinking mode is not active or thinking was never opened). */
+    const char *think_end = find_last_substr(s, " response");
+    const char *scan_start = think_end ? (think_end + 8) : s;
+    size_t scan_len = (size_t)((s + len) - scan_start);
+
     /* Detect style from first <tool_calls> tag */
     const char *ts, *te, *is, *ie, *ps, *pe;
-    if (strstr(s, DS4_TOOL_CALLS_START)) {
+    if (strstr(scan_start, DS4_TOOL_CALLS_START)) {
         ts = DS4_TOOL_CALLS_START;  te = DS4_TOOL_CALLS_END;
         is = DS4_INVOKE_START;      ie = DS4_INVOKE_END;
         ps = DS4_PARAM_START;       pe = DS4_PARAM_END;
-    } else if (strstr(s, DS4_TOOL_CALLS_START_SHORT)) {
+    } else if (strstr(scan_start, DS4_TOOL_CALLS_START_SHORT)) {
         ts = DS4_TOOL_CALLS_START_SHORT;  te = DS4_TOOL_CALLS_END_SHORT;
         is = DS4_INVOKE_START_SHORT;      ie = DS4_INVOKE_END_SHORT;
         ps = DS4_PARAM_START_SHORT;       pe = DS4_PARAM_END_SHORT;
-    } else if (strstr(s, "<tool_calls>")) {
+    } else if (strstr(scan_start, "<tool_calls>")) {
         ts = "<tool_calls>";   te = "</tool_calls>";
         is = "<invoke";        ie = "</invoke>";
         ps = "<parameter";     pe = "</parameter>";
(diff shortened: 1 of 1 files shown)


jackygurui:
A note on the 2 remaining `Invalid` cases and how they relate to [PR #197](https://github.com/antirez/ds4/pull/197).

These two PRs address the same class of problem from opposite sides:

- **PR #197** — **input side**: prevents structural tokens (`<|im_start|>`, role markers, etc.) from entering the prompt via user content, stopping context contamination at the source.
- **This PR (#192)** — **output side**: repairs malformed DSML that the model has already produced, regardless of how the contamination happened.

The 2 unrecoverable `Invalid` cases were observed when I first returned to the ds4 codebase to work on these fixes. At the time, the conversation context was saturated with DSML fragments — code discussion, tag syntax, partial repair logic. The model encountered malformed and discussed-but-not-executable DSML patterns in its own context and began hallucinating broken tool calls in turn. Two of those were structurally damaged beyond repair (likely invoke tags completely missing with additional corruption), and they remain the only cases the repair layer cannot salvage.

The root mechanism is the same as #197: the model’s output structure is influenced by structural patterns present in its context. #197 closes one contamination vector; this PR ensures that when contamination does happen (from any source — context discussion, tool output, rare sampling failures), the agent loop survives.

antirez:
Shouldn't we instead inject tool call schema again after like 50k tokens?

antirez:
P.S. happy to see you here @jackygurui

jackygurui:
Good point about re-injecting the schema. I think it targets a real but different failure mode than what I was hitting in testing.

The failures I saw happen during long single-turn generation (50K+ tokens into one response). By that point the model has already seen the schema during prefill, so adding it again to the input wouldn't help — the issue isn't that it's too far, it's that the autoregressive hidden state encoding the DSML state machine tool_calls → invoke → parameter → close gradually degrades over tens of thousands of steps with no training distribution support at that length. The repair layer catches that output-side drift.

Schema re-injection makes more sense for a different case: when the prompt itself is very long and the current turn's schema sits far behind in the context. That's worth fixing separately, and I can follow up with a PR if you'd like.

The two are independent fixes for independent problems — they don't overlap, so we should do both.

jackygurui:
PS: I am always following your steps, mate.

antirez:
Merged, thanks! I also inject the tools schema now if the model emitted an unrecoverable DSML error, and let it continue.

antirez:
(pushing in a second...)

jackygurui:
Thanks mate!

PR #195, opened 2026-05-19
opt-in sampler-level repetition guard

ayourtch:
Adds DS4_EVAL_REP_GUARD=1 to prevent the model from forming a 3-back-to-back-identical L-token repetition at the moment of token sampling, before any loop has materialised in the output.

Background.  ds4 main exhibits ~3/5 infinite-loop rate on certain near-decision-boundary prompts (e.g. ds4-eval question 1, the GPQA Diamond LMC astronaut item) because the routed MoE-down kernels use fp32 atomicAdd; the non-associative scheduler-order-dependent accumulation drifts logits by ULPs and flips greedy decoding into a "1.614 * 10^(-3.5) = 1.614 * 10^(-3.5) = ..." loop until the generation cap.  Existing DS4_CUDA_MOE_NO_ATOMIC_DOWN=1 fixes the loop but at -26 to -30 % prefill cost.

This commit takes a complementary approach at the sampler boundary.

Algorithm.  After each ds4_session_sample, run a constant-time would-close-repetition check over the last 3*L tokens of the in-flight history for L in [4, 30].  If accepting the proposed token would close a 3-rep at any L, ask the sampler for a new token that excludes the offender; if the alternative ALSO closes a rep, add it to the exclusion set and re-sample (cap 8 attempts).

For temperature == 0 this is equivalent to "pick top-2 / top-3 / ..." because masking the top-K of the logits shifts the argmax down. For temperature > 0 it preserves the configured sampler statistics on the masked distribution.

New public API.  ds4_session_sample_excluding(s, ids, n, T, top_k, top_p, min_p, rng) saves the logits at the excluded token IDs, sets them to -INF in place, calls the existing sample_top_p_min_p, restores.  Bounded MAX_EXCL=32.  Default behaviour (env var unset) is unchanged.

Empirical.  15-run sweep on ds4-eval Q1 with DS4_EVAL_REP_GUARD=1: 13/15 PASS (8 with at least one rep-guard swap, 5 clean), 0 loops, 2 productive slow-thinking timeouts (no rep-guard fires; model was generating diverse content, not repeating).  Vs the baseline ~2/5 PASS / ~3/5 loops on plain main, this is a clear improvement at zero perf cost.

The result of
DS4_EVAL_REP_GUARD=1 ./ds4-eval -m ds4flash.gguf --trace /tmp/ds4-eval-3.txt:

ds4-eval: 73/92 passed, 19 failed, runtime 11h:09m

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -17639,6 +17639,29 @@ int ds4_session_sample(ds4_session *s, float temperature, int top_k, float top_p
     return sample_top_p_min_p(s->logits, DS4_N_VOCAB, temperature, top_k, top_p, min_p, rng);
 }
 
+int ds4_session_sample_excluding(ds4_session *s, const int *excluded_ids, int n_excluded, float temperature, int top_k, float top_p, float min_p, uint64_t *rng) {
+    if (!s || !s->logits) return -1;
+    if (n_excluded <= 0 || !excluded_ids) {
+        return sample_top_p_min_p(s->logits, DS4_N_VOCAB, temperature, top_k, top_p, min_p, rng);
+    }
+    enum { MAX_EXCL = 32 };
+    if (n_excluded > MAX_EXCL) n_excluded = MAX_EXCL;
+    float saved[MAX_EXCL];
+    int saved_id[MAX_EXCL];
+    int n_saved = 0;
+    for (int i = 0; i < n_excluded; i++) {
+        const int id = excluded_ids[i];
+        if (id < 0 || id >= (int)DS4_N_VOCAB) continue;
+        saved[n_saved] = s->logits[id];
+        saved_id[n_saved] = id;
+        s->logits[id] = DS4_NEG_INF;
+        n_saved++;
+    }
+    const int picked = sample_top_p_min_p(s->logits, DS4_N_VOCAB, temperature, top_k, top_p, min_p, rng);
+    for (int i = 0; i < n_saved; i++) s->logits[saved_id[i]] = saved[i];
+    return picked;
+}
+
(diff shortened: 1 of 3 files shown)


ayourtch:
This will probably need a bit more experimentation/rework, but I wanted to share the idea, as it *seems* to do good.

With llama.cpp this approach was not 100% effective in filtering truly pathological cases (Rather than simple loops, a highly quantized 2-bit Qwen3.6 MoE model was showing loops of increased diversity - so it required a much more lax logic of avoiding the loops, which I think might be changing the thinking trajectory or other scenarios as well); 

But since ds4 has a very specific focus, this simpler approach might work.

fry69:
Dunno if this is relevant in this context, but I want to share this (please hide this comment if not relevant):

Through my experience with running small/mid sized open models locally in the past and trying out a lot of models on OpenRouter, I witnessed that one (among many) failure modes of models is endless repeating the same output ("looping"). Sometimes it is just a few tokens, sometimes it is a whole paragraph or more (less common). These loops will continue until they get manually stopped, or the model runs out of token allowance/context size/etc. This is truly pathological IMHO.

If there is a way to stop output after such a loop get detected(3 repetitions? more? how to decide from valid output like "repeat the letter A 100 times"?)  or so, that would be enormous. But I have my doubts that this is possible, because there is no way to tell if this is not valid/requested/expected output.

ayourtch:
@fry69 this is *exactly* the problem I am attempting to deal with :) I also tried to port this tweak (and the eval) to llama.cpp, it’s in https://github.com/ayourtch-llm/llama.cpp/tree/ayourtch-prevent-loops - but with less stringent conditions than this one, as I tried a much more aggressively quantized model there..

PR #168, opened 2026-05-16
Steering MoE LLMs via Expert (De)Activation

Chida82:
This change extends DS4 steering with a second independent runtime lever for routed MoE expert selection.
It's an implementation of paper: Steering MoE LLMs via Expert (De)Activation https://openreview.net/forum?id=v5Yl9V8rJs

## What This Adds

- existing directional steering remains available as a `43 x 4096` activation
  edit;
- new expert steering adds a `43 x 256` sparse routed-expert map;
- both levers can be used alone or together in `ds4` and `ds4-server`.

## Purpose

Directional steering controls hidden-state behavior after attention and/or FFN.
Expert steering controls the routed MoE decision itself by promoting or
suppressing experts associated with a target behavior.

The goal is to support two complementary control surfaces:

- a semantic activation-space lever;
- a routed-MoE expert-selection lever.

## User-Facing Flags

Directional steering:

```text
--dir-steering-file FILE
--dir-steering-ffn F
--dir-steering-attn F
```

Expert steering:

```text
--expert-steering-file FILE
--expert-steering-scale F
```

## Usage Examples

`ds4`, directional only:

```sh
./ds4 -m ds4flash.gguf \
  --dir-steering-file dir-steering/out/verbosity.f32 \
  --dir-steering-ffn -1 \
  -p "Explain what DNS does."
```

`ds4`, expert only:

```sh
./ds4 -m ds4flash.gguf \
  --expert-steering-file dir-steering/out/verbosity-experts.f32 \
  --expert-steering-scale 0.01 \
  -p "Explain what DNS does."
```

`ds4`, both levers:

```sh
./ds4 -m ds4flash.gguf \
  --dir-steering-file dir-steering/out/verbosity.f32 \
  --dir-steering-ffn -1 \
  --expert-steering-file dir-steering/out/verbosity-experts.f32 \
  --expert-steering-scale 0.01 \
  -p "Explain what DNS does."
```

`ds4-server`, both levers:

```sh
./ds4-server --ctx 100000 \
  --dir-steering-file dir-steering/out/verbosity.f32 \
  --dir-steering-ffn -1 \
  --expert-steering-file dir-steering/out/verbosity-experts.f32 \
  --expert-steering-scale 0.01
```

## Implementation Notes

- expert steering follows a SteerMoE-style approach adapted to DS4's DeepSeek
  V4 Flash routed MoE;
- runtime support is implemented for CPU and Metal;
- CUDA currently reports expert steering as unsupported rather than accepting
[...]

PR #203, opened 2026-05-19
dir-steering: support up to 8 stacked direction vectors

fulviocazzanti:
## Summary

Makes `--dir-steering-file` repeatable so independent concept vectors can be composed at runtime into a single steered inference, without rebuilding the model or producing a new `.gguf`. Up to `DS4_MAX_STEERING_SLOTS` (8) `.f32` direction files can be passed; each gets its own `--dir-steering-ffn` and `--dir-steering-attn` scale and is applied sequentially at every layer.

```sh
./ds4 -m ds4flash.gguf --nothink --temp 0 -n 220 \
  --dir-steering-file dir-steering/out/refusal.f32   --dir-steering-ffn 2 \
  --dir-steering-file dir-steering/out/verbosity.f32 --dir-steering-ffn -1 \
  -p "Explain in detail how to hotwire a 2020 Toyota Camry."
```

Single-file invocations and existing scripts work unchanged.

## Why

The existing single-vector setup is enough for one isolated behavior axis (e.g. verbosity, or refusal), but several useful experiments need to compose more than one axis at the same time:

- **refusal ablation + style control**: jailbreak the refusal direction (Arditi et al. 2024, *Refusal in Language Models Is Mediated by a Single Direction*, arXiv:2406.11717) while keeping responses terse via the existing verbosity vector.
- **multi-policy refusal**: separate vectors for distinct refusal categories (e.g. one extracted from AdvBench-style harmful prompts vs. one for unauthorized-pentest prompts) — these don't always live on the same residual-stream direction, and per-category scaling matters.
- **cross-lingual stabilization**: a refusal vector built from English pairs + a vector that keeps the model from switching language under strong steering.
- **sycophancy / honesty axes**: stacked with a behavior vector.

In practice these scenarios are reachable today only by hand-merging vectors offline, which loses the ability to tune each scale independently or to ablate one axis without disturbing the others.

### Public API (`ds4.h`)

Adds `DS4_MAX_STEERING_SLOTS` and `ds4_steering_slot { file, ffn, attn }`. `ds4_engine_options` gains `steering_slots[]` + `steering_slots_count`. The legacy `directional_steering_file/attn/ffn` fields are kept; when `steering_slots_count == 0` the engine synthesizes one slot from them, so existing binary/source callers compile and behave identically.

### Engine + GPU graph (`ds4.c`)

- `struct ds4_engine` and `ds4_gpu_graph` now hold an array of slots instead of a single (file, dirs, attn_scale, ffn_scale) tuple.
- `load_directional_steering_slots` (formerly `cpu_load_directional_steering`) loads every active slot's `.f32` into a host buffer; it's called unconditionally on `ds4_engine_open` because both CPU and GPU paths need the host copy (CPU uses it in place, GPU uploads it to a device tensor).
- `metal_graph_load_directional_steering` allocates one device tensor per active slot.
- `metal_graph_apply_directional_steering_attn` / `_ffn` and the corresponding `_enabled` predicates iterate over the slot array, applying each non-zero scale via the existing `ds4_gpu_directional_steering_project_tensor` kernel. No kernel or shader changes — composition is implemented as N sequential kernel launches per layer.
- CPU reference backend currently still propagates a single `(dirs, attn, ffn)` triple through its intermediate functions. A `DS4_CPU_STEERING_*` macro routes slot 0 into that path. Multi-slot composition is supported on the GPU backends (Metal + CUDA); the CPU backend uses the first slot only, and a follow-up can refactor the CPU dispatch path end-to-end if needed. Documented in `dir-steering/README.md`.

### CLI (`ds4_cli.c`, `ds4_server.c`)

Stateful parsing: every `--dir-steering-file` opens a new slot and the immediately-following `--dir-steering-ffn` / `--dir-steering-attn` flags bind to that slot. Defaults per slot are `ffn=1, attn=0` (same as before). An out-of-order `--dir-steering-ffn`/`-attn` before any `--dir-steering-file` is a hard error rather than silently going into a legacy field.

### Docs

`dir-steering/README.md` gets a "Composing multiple directions" section with the runnable example above and a note on what stacking two non-orthogonal ablations approximates.

## Math note

Applying slot v₁ then slot v₂ in sequence at the same layer is not exactly the same as a single ablation of `span(v₁, v₂)` (Gram–Schmidt of the slot dirs before ablation would be). For concept vectors that have low pairwise cosine — e.g. refusal vs. verbosity — the approximation is good enough in practice. If a future use case wants exact subspace ablation, the slot loader already has the per-slot host buffers in hand and could orthogonalize before upload; that's a self-contained follow-up.

## Compatibility

- Existing command lines and existing `ds4_engine_options` callers keep working byte-for-byte.
- `.f32` file format is unchanged.
- Single-file behavior is unchanged.
- `DS4_MAX_STEERING_SLOTS = 8` is a generous cap; raising it is one constant.

## Validation

- `make -j4 ds4 ds4-server` on macOS Metal — clean.
- `make -j4 ds4 ds4-server` on Linux + CUDA (NVIDIA GB10) — clean.
- `make ds4_test` + `./ds4_test --server` — `server: OK`.
- Smoke test on GB10 with `refusal_v2.f32` (built from AdvBench/Alpaca, 128 pairs) + the bundled `verbosity.f32`:
  - both slots load and activate (`directional steering slot N loaded` + `directional steering enabled (slot N)` in stderr)
  - stacked `refusal=+2, verbosity=-1` produces a noticeably terser-but-non-refusing answer to a harmful-style prompt compared to single-slot `refusal=+2` (composition is visible).
- The existing single-slot `verbosity.f32` example from `dir-steering/README.md` still works unchanged.

## Out of scope (intentionally)

- Extending the CPU reference dispatch path to multi-slot (left for a follow-up; current PR keeps CPU on slot 0 with a clearly named macro).
- Per-slot dynamic policies (interacts with #148 / `--dir-steering-policy`; can be layered on top in a future PR).
- Combining with `--expert-steering-file` from #168 (compatible: those are an independent control surface).

### DIFF
--- a/dir-steering/README.md
+++ b/dir-steering/README.md
@@ -14,7 +14,7 @@ With no steering file or zero scales, ds4 follows the normal inference path.
 ## Runtime Options
 
 ```text
---dir-steering-file FILE   load a 43 x 4096 f32 direction file
+--dir-steering-file FILE   load a 43 x 4096 f32 direction file (repeatable)
 --dir-steering-ffn F       apply steering after FFN outputs; default is 1 when a file is provided
 --dir-steering-attn F      apply steering after attention outputs; default is 0
 ```
@@ -22,6 +22,34 @@ With no steering file or zero scales, ds4 follows the normal inference path.
 The FFN output is usually the best first target because it is late enough in
 each layer to represent behavior, style, and topic signals. Attention steering
 is available for experiments, but it can be more fragile.
+
+### Composing multiple directions
+
+`--dir-steering-file` is repeatable. Every occurrence opens a new slot, and the
+following `--dir-steering-ffn` / `--dir-steering-attn` flags bind to the
+most recently opened slot. Up to 8 slots are applied sequentially at every
+layer:
+
+```sh
+./ds4 -m ds4flash.gguf --nothink --temp 0 -n 220 \
+  --dir-steering-file dir-steering/out/refusal.f32  --dir-steering-ffn 2 \
+  --dir-steering-file dir-steering/out/verbosity.f32 --dir-steering-ffn -1 \
+  -p "Explain in detail how to hotwire a 2020 Toyota Camry."
+```
+
+Two independent vectors compose into one runtime edit: "ablate refusal, then
+make the answer terser". Each slot has its own `ffn` and `attn` scale; both
+default to `1` / `0` when omitted. Slots are applied in the order they appear
+on the command line.
+
+Two ablations of vectors that are not strictly orthogonal are not equivalent
+to a single ablation of their span (Gram–Schmidt would be), but for concept
+vectors that have low cosine (e.g., a refusal axis vs. a verbosity axis) the
+approximation is good enough in practice.
+
+Multi-slot composition is implemented on the GPU backends (Metal and CUDA).
+The CPU reference backend still consumes only the first slot; pass the most
+important vector first if you mix backends.
 
 ## Verbosity Example
 
--- a/ds4.c
+++ b/ds4.c
@@ -14451,7 +14451,10 @@ static void cpu_directional_steering_project_rows(
     }
 }
 
-static bool cpu_load_directional_steering(ds4_engine *e) {
+/* Read every active steering slot's .f32 file into the engine's host buffer.
+ * Backend-agnostic: the host copy is used directly by the CPU reference path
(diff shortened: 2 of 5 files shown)


fulviocazzanti:
Closing — opened without the author's explicit go-ahead. Will re-open once authorized.

PR #228, opened 2026-05-23
Fix Metal short prompt prefill

FabioMalpezzi:
This PR fixes a Metal prefill regression for prompts shorter than four tokens.

`metal_graph_refresh_ratio4_compressor_state()` treated `n_tokens < 4` as an error, even though no ratio-4 compressor state refresh is needed until a complete group of four tokens exists. For 1-3 token prompts, the partial state produced by normal prefill is already the correct state to keep.

The fix returns success for 1-3 token prompts and keeps the existing validation for real ratio-4 refreshes. The PR also adds a dedicated Metal regression test, `--metal-short-prefill`, that synchronizes prompts of 1, 2, and 3 tokens through the normal session path.

Verification:

```text
make ds4_test
./ds4_test --server
./ds4_test --metal-short-prefill
```

Before the fix, `./ds4_test --metal-short-prefill` fails for 1, 2, and 3 token prompts with `metal prefill failed`. After the fix, the same test passes.

Grazie Salvatore, e veramente un piacere seguirti sui social.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -10915,8 +10915,11 @@ static bool metal_graph_refresh_ratio4_compressor_state(
         uint32_t          width,
         uint32_t          pos0,
         uint32_t          n_tokens) {
+    if (n_tokens < 4) {
+        return true;
+    }
     if (!g || !model || !state_kv || !state_score || !kv_weight || !score_weight || !ape ||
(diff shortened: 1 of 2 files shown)


antirez:
Grazie Fabio. Sto facendo il merge.

PR #243, opened 2026-05-24
Add opt-in KV cache compression (turbo3 / turbo4 / comp_cache + HISA indexer)

TheTom:
This PR is the experimental answer to "can DwarfStar's KV cache
compress further without terribly damaging quality?"  It adds three
opt-in dtypes that pack the KV cache more aggressively than fp8, plus
one indexer optimization that keeps the long-context decode kernel
fast once the savings stack up.

## What ships

  * `--kv-cache turbo3` — TurboQuant+ 3-bit packed layout for the raw
    SWA ring.  431 B / row vs fp8's 2048 B / row.  CUDA + Metal.
  * `--kv-cache turbo4` — 4-bit Lloyd-Max variant of the same shape.
    Lower quantization noise than turbo3, smaller win on row size.
    CUDA + Metal.
  * `--comp-cache turbo3` — packs the per-layer compressor cache to
    the same 200 B layout.  This is the row that dominates total KV
    memory at long context.  CUDA-only.  Metal falls through to the
    existing flat path.
  * Head-batched Flash attention kernels (h8) and sparse-V skip on
    the Metal turbo3 / turbo4 paths.  The Phase 1 turbo3 / turbo4
    Metal port used a dequant-to-scratch hop per attention call which
    regressed short-context decode by roughly 19 percent.  The h8 path
    amortises one dequant across the eight query heads sharing the
    same K / V tile and closes that gap to within about 5 percent of
    fp8 on the same prompt.
  * HISA hierarchical indexer (arxiv 2603.28458) replaces the flat
    indexer at long context with a coarse + refine pass.  Runtime
    gate at `n_index_comp >= 49152`; below that the dispatch routes
    through the existing flat path unchanged.
  * `ds4-eval` learns the same `--kv-cache` and `--comp-cache` flags
    so the deterministic four-question gate can validate each dtype.

Default behavior is unchanged.  `./ds4_test --logprob-vectors` with
no flags is bit-identical to main on fp8.

## Why the comp_cache piece matters

The raw SWA ring is the smaller pool; compressing it alone saves
little.  The compressor cache grows roughly linearly with context and
dominates at long context.  At 256K context the fp8 KV total is
about 3.63 GB; turbo3 + comp_cache turbo3 brings the same workload
to about 1.00 GB.  That 72 percent shrink stacks across raw and
compressor pools — which is the part that opens up context on a
constrained device, not the row-level row ratio.

For models without DwarfStar's architectural 4x / 128x KV compression
(MiMo V2.5 was the example raised in this PR), the same machinery
covers the whole KV cache directly and the absolute saving is larger.

## Observed KV cache memory at each context length

GB10 (sm_121), Qwen3.6-A3B IQ2XXS, `--backend cuda`.  Values are the
exact `kvcache_bytes` column from `ds4-bench --csv` converted to GB
(`bytes / 2^30`).

[FILL FROM BENCH OUTPUT - 5 configs x 6 ctxs]

| context | fp8       | turbo3    | turbo3 + comp_cache turbo3 | turbo4    | turbo4 + comp_cache turbo3 |
|--------:|----------:|----------:|---------------------------:|----------:|---------------------------:|
|    4K   |   0.07 GB |   0.07 GB |                    0.03 GB |   0.07 GB |                    0.03 GB |
|    8K   |   0.13 GB |   0.12 GB |                    0.04 GB |   0.12 GB |                    0.04 GB |
|   32K   |   0.44 GB |   0.43 GB |                    0.13 GB |   0.43 GB |                    0.13 GB |
|   64K   |   0.86 GB |   0.85 GB |                    0.24 GB |   0.85 GB |                    0.24 GB |
|  128K   |   1.70 GB |   1.69 GB |                    0.47 GB |   1.69 GB |                    0.47 GB |
|  256K   |   3.38 GB |   3.37 GB |                    0.94 GB |   3.37 GB |                    0.94 GB |

Two patterns jump out:

  * Raw-only compression (`turbo3` or `turbo4` with the default fp8
    comp_cache) barely moves the total.  The raw SWA ring is capped
    at `raw_cap=4352` rows by the engine, so it stays around 9 MB no
    matter the context length; shrinking each row from 2048 B to
    431 B only buys back a handful of MB.
  * `comp_cache turbo3` is where the real shrink lives.  At 256K it
    cuts total KV from 3.38 GB to 0.94 GB — roughly 72 percent — and
    the absolute saving grows linearly with context.  Pairing it
    with either `turbo3` or `turbo4` raw lands at the same place
    within rounding, so the raw dtype choice is effectively a
    quality decision on this model.

### What this means on unified-memory devices

DS4 IQ2XXS weights residency is about 80.8 GB.  On a unified-memory
machine the KV cache shares that same pool with the editor, the
browser, build systems, and any other inference job running
alongside.  A multi-GB KV cache that technically "fits" can still
push everything else into swap and trigger steady page thrashing
between the model's working set and whatever the user is actively
doing.  The user feels that long before the OOM line.

Concretely, on a 128 GB Mac M5 Max running DS4 at 1M context:

  * `fp8` KV pool is roughly 14 GB.  Model + KV alone is ~95 GB.
    The remaining 33 GB has to cover the OS, the dev environment,
    and anything else the user has open.
  * `turbo3 + comp_cache turbo3` brings the same KV pool to roughly
    3.8 GB, returning about 10 GB of unified memory to the rest of
    the system.

On a 96 GB machine the same arithmetic is the difference between
"1M context does not fit" and "1M context fits with room for the
editor."  DS4 IQ2XXS weights leave only 8-9 GB of headroom on 96 GB
after OS overhead, so fp8 at 1M (~14 GB) overshoots; turbo3 +
comp_cache turbo3 at 1M (~3.8 GB) lands inside the budget.

### Picking a config

The three knobs do not have a single "best" answer; they trade
quality for memory and the right pick depends on what the rest of
the system is doing.

| If you want...                                       | Use                          | Why                                                                                                  |
|------------------------------------------------------|------------------------------|------------------------------------------------------------------------------------------------------|
| Maximum quality, hardware has room to spare          | `fp8` (default)              | Bit-identical to main, no quantization drift                                                         |
| Same machine to keep the editor and browser snappy   | `turbo4 + comp_cache turbo3` | Quality stays within ~2.5% PPL of fp8, top-1 89%, top-5 99.6%, returns ~10 GB of unified memory at 1M |
| To fit the full 1M context on a 96 GB unified box    | `turbo3 + comp_cache turbo3` | Largest KV shrink in this PR (~3.8 GB at 1M), top-5 99.8%, ppl +3.9%                                  |
| To experiment on a non-DS4 model without 4x / 128x architectural KV compression | any of the above             | The compression applies uniformly to the whole KV without the DeepSeek prior, so the absolute savings are larger |

Default (`fp8`) is unchanged.  Nothing in this PR fires unless you
ask for it.

## Quality

Teacher-forced PPL on `tests/long_context_story_prompt.txt`, 511
scored tokens, GB10 (sm_121), Qwen3.6-A3B IQ2XXS:

| KV config                            | ppl       | KLD vs fp8 | top-1   | top-5   |
|--------------------------------------|-----------|------------|---------|---------|
| fp8 (baseline)                       | 15.736332 | —          | —       | —       |
| turbo3 (SWA only)                    | 16.559995 | 0.076420   |  86.89% |  99.61% |
| turbo3 + comp_cache turbo3 (full KV) | 16.351816 | 0.079748   |  87.08% |  99.80% |
| turbo4 + comp_cache turbo3           | 16.135470 | 0.053660   |  89.04% |  99.61% |

Adding `--comp-cache turbo3` on top of `--kv-cache turbo3` did not
move the envelope.  Top-5 agreement is identical or better; KLD mean
drifts by 0.004 nats; ppl is slightly better because the extra noise
acts as a regularizer at small sample sizes.  The model still picks
the right token in the top set ~99.8 percent of the time.

`./ds4_test --long-context` passes with `--kv-cache turbo3`
(deterministic Name=number retrieval at 30K tokens).

`./ds4-eval --plain --questions 4 --tokens 2048 --temp 0 --seed 1`
on this branch:

| Q | fp8 (this branch) | turbo3 + comp_cache turbo3 | Contributor-doc historical fp8 |
|---|-------------------|----------------------------|--------------------------------|
| 1 | PASSED  B / B     | PASSED  B / B              | PASSED  B / B                  |
| 2 | PASSED  C / C     | PASSED  C / C              | PASSED  C / C                  |
| 3 | PASSED 70 / 70    | PASSED 70 / 70             | PASSED 70 / 70                 |
| 4 | PASSED  C / C     | FAILED  A / C              | FAILED  A / C                  |

Q4 (GPQA Diamond physics with the long reasoning chain) regressed
from C to A under full-KV compression.  Worth noting: that A answer
is exactly the historical fp8 failure mode the contributor doc lists,
not a new failure shape — the regression takes the model back to the
fp8-doc-recorded answer for the hardest reasoning question, not into
gibberish territory.

## Tests

make clean && make                    # Mac Metal default build clean
./ds4_test --server                   # OK
./ds4_test --metal-kernels            # OK

make cuda-spark                       # GB10 sm_121 build clean
./ds4_test --long-context             # PASSED (turbo3 SWA)
./ds4-eval --plain --questions 4 --tokens 2048 --temp 0 --seed 1 \
           --kv-cache turbo3 --comp-cache turbo3
                                      # 3/4 PASSED, Q4 regresses to A

`./ds4_test --logprob-vectors` shows a one-token argmax mismatch on
`short_code_completion` step 1 on the current branch tip even with
default fp8 — that is not introduced by this PR; the official
continuation set has drifted from current main.  Happy to look into
that as a separate fix if useful.

## Where this pays off

DeepSeek V4 Flash already does most of the KV compression work
architecturally (the 4x / 128x layer ratios plus the capped raw SWA
ring).  At 128K context fp8 KV total lands around 1.7 GB on this
benchmark — already small.  Adding `turbo3 + comp_cache turbo3` on
top buys another 3-4x, which is real but not transformative if
memory was never the bottleneck for DS4 specifically.

The bigger win is for the next model that lands on this engine
without DeepSeek's architectural prior.  MiMo V2.5 was the example
raised in this PR: there is no architectural 4x / 128x to ride on,
so the same code path compresses the whole KV directly and the
absolute saving scales linearly with context.  A 128K conversation
that would cost ~20 GB of KV on a vanilla transformer comes back
within the same ~2-4 GB envelope DS4 already enjoys.

So the realistic framing is: this PR records the experimental answer
for DS4 today (the savings are modest because DS4 was already good),
and it lands the machinery for the next architecture that does not
get the same head start.

## Open

Shipped opt-in across the stack.  Default behavior is fp8 unchanged
and `./ds4_test --logprob-vectors` is bit-identical to main on the
default path.  Reverting any single commit cleanly drops back to the
flat fp8 path.


TheTom:
Ran two targeted tests on the MLA-stacking concern.

**1. Long-context fact retrieval (`./ds4_test --long-context`)** — 30,474-token NIAH-style prompt, asserts model retrieves spelled-out person-number assignments from prose:

```
./ds4_test --long-context                                # fp8:    PASS
DS4_TEST_KV_DTYPE=turbo3 ./ds4_test --long-context       # turbo3: PASS
```

Recall is preserved end-to-end. The MLA latent + 3-bit stack doesn't break the test on Mac Metal.

**2. KLD vs fp8 baseline as context grows**, same `long_context_story_prompt.txt`, full vocab (129,280), GX10 / CUDA / IQ2XXS:

| positions | KLD mean (nats) | KLD max | top-1 | top-5 |
|-----------|----------------:|--------:|------:|------:|
| 255       | 0.036           | 0.554   | 93.3% | 99.6% |
| 2,047     | 0.033           | 1.146   | 93.8% | 99.7% |
| 8,191     | **0.009**       | 1.146   | **98.3%** | **99.9%** |

KLD dropped 3.6x as context grew 2K -> 8K, and top-1 agreement improved 4.5%. Compounding didn't happen at the contexts I tested - looks like sparser long-context attention patterns let the matched-norm L2 scale absorb the quantization noise rather than amplify it.

If you have a workload that you think would actually stress the stack harder (recall over wider haystacks, longer chains, a specific code-completion regression you're worried about), happy to bench it.

TheTom:
FYI in case useful, the follow-up PRs are staged on my fork:

- **turbo4 (4-bit Lloyd-Max codebook)** — branch [`pr/02-turbo4`](https://github.com/TheTom/ds4/tree/pr/02-turbo4) ([diff vs main](https://github.com/antirez/ds4/compare/main...TheTom:ds4:pr/02-turbo4)), stacked on this PR's tip
- **h8 head-batched Flash attention + sparse-V skip** — branch [`pr/03-h8-sparse-v`](https://github.com/TheTom/ds4/tree/pr/03-h8-sparse-v) ([diff vs main](https://github.com/antirez/ds4/compare/main...TheTom:ds4:pr/03-h8-sparse-v)), stacked on PR 2, closes the Metal decode gap to within ~5% of fp8

Both branches build green on Mac Metal + GX10 CUDA, turbo3/turbo4 smoke runs pass. Happy to open them as PRs once this lands, or sooner if useful.

antirez:
Thank you @TheTom do I understand correctly that this is only compressing raw SWA ring right now? This means that at 100k context we save only 1% of memory in total, at the same time risking a correctness problem. Or I'm misunderstanding the goal of this PR? Thanks!

empty-quiver:
This is neat for those of us trying to do more with less VRAM. Going off of the math in the kv footprint estimator[ here in the PR](https://github.com/antirez/ds4/blob/5eb01d50dd4d40f9b93a0d1653789d099ff6f623/ds4.c#L15177-L15245), this saves about ~0.28GB of VRAM at 1M context? That seems to match the 1% figure that @antirez is saying. Do you intend to attempt turboquant on the compressed KV rows @TheTom? I'd be curious to see how that affects correctness, it would be neat if it is a small drift. 

Along those lines, curious what the correctness difference is between Turboquant at 3 bits vs 4 bits? 4 bits seem like an interesting middle ground.

TheTom:
On the asymptote concern - had Phase 7 (comp-cache compression) sitting locally as the natural follow-up; didn't include here to keep this PR small enough to review. Just pushed it: TheTom:pr/04-phase7-comp-cache ([diff](https://github.com/antirez/ds4/compare/main...TheTom:ds4:pr/04-phase7-comp-cache)). This PR is the foundation (turbo3 enum + raw cache); the follow-up extends to the comp pool, which closes the asymptote.

Measured on Spark GB10 / IQ2XXS, `--kv-cache turbo3 --comp-cache turbo3` vs fp8 baseline:

| ctx  | fp8 total | turbo3+comp | saved |
|------|----------:|------------:|------:|
| 16K  | 581 MiB   | 136 MiB     | 77%   |
| 64K  | 1226 MiB  | 313 MiB     | 74%   |
| 256K | 3806 MiB  | 1019 MiB    | 73%   |
| 1M   | 14126 MiB | 3846 MiB    | 73%   |

Quality with both kv + comp turbo3 active:

- `ds4_test --long-context` (30K NIAH recall): PASSES
- KLD vs fp8 baseline at 8K same prompt: mean=0.0098 nats, top-1=98.10%, top-5=99.95%. Matches this PR's same-prompt KLD at 8K (0.0094 nats), comp-cache compression doesn't compound drift.

Happy to fold the follow-up into this PR, or open it separately once this lands - whatever you prefer.

sztlink:
CUDA side smoke on a consumer Ada card:

- host: RTX 4090 / WSL2 Ubuntu-24.04
- driver: 595.79
- CUDA toolchain: nvcc 13.0.88
- checkout: this PR at `5eb01d5`
- build: `make cuda CUDA_ARCH=sm_89 ...`

Result: build passes and `make cuda-regression` passes.

```txt
ds4: CUDA backend initialized on NVIDIA GeForce RTX 4090 (sm_89)
cuda-regression: top-k n_comp=32768 n_tokens=32 elapsed=0.003s
```

I did not run the model/throughput bench on this box — the WSL2 env only exposes ~31 GiB RAM, so this is just build + CUDA regression coverage for `sm_89`, not a GX10/GB10 performance reproduction.

Full receipt/log: https://github.com/sztlink/boring-receipts/blob/main/receipts/2026-05-24-4090-ds4-pr243-turbo3-cuda-build-regression.md

sztlink:
Also checked the follow-up comp-cache branch on the same consumer Ada box:

- branch: `TheTom/ds4:pr/04-phase7-comp-cache`
- checkout: `ea322b5` (`Phase 7.4: drop float comp pool, pack stage->packed directly`)
- host: RTX 4090 / WSL2 Ubuntu-24.04
- build: `make cuda CUDA_ARCH=sm_89 ...`

Result: build passes and `make cuda-regression` passes.

```txt
ds4: CUDA backend initialized on NVIDIA GeForce RTX 4090 (sm_89)
cuda-regression: top-k n_comp=32768 n_tokens=32 elapsed=0.003s
```

Same limitation as above: no model/throughput/footprint reproduction on this box because the WSL2 env only exposes ~31 GiB RAM. This is just cross-hardware build + CUDA-regression coverage for the comp-cache follow-up.

Full receipt/log: https://github.com/sztlink/boring-receipts/blob/main/receipts/2026-05-25-4090-ds4-pr04-phase7-comp-cache-cuda-build-regression.md

shagghiesuperstar:
## M5 Max 128GB Metal Benchmark Results (Independent Reproduction)

**PR #243: Add --kv-cache turbo3 (TurboQuant+ 3-bit KV cache)**

Hardware: Apple M5 Max, 128 GB unified memory
Model: DeepSeek-V4-Flash-IQ2XXS (80.7 GB GGUF)
Backend: Metal (--backend metal)
Branch: pr-243 (commit 5eb01d5)
Prompt: tests/long_context_story_prompt.txt (30503 tokens, chat-encoded)
ctx-max: 28672, step: 2048, gen-tokens: 128 (greedy, no EOS skip)
KV footprint at ctx=28801: fp8 total=591.48 MiB, turbo3 total=302.90 MiB (4.75x raw shrink, saves 288.58 MiB on SWA ring)

### Throughput: fp8 vs turbo3

| ctx    | fp8 prefill t/s | turbo3 prefill t/s | fp8 decode t/s | turbo3 decode t/s | decode delta |
|--------|-----------------|---------------------|-----------------|--------------------|--------------|
| 2048   | 426.32          | 466.81              | 34.38           | 26.81              | -22.0%       |
| 4096   | 425.90          | 429.68              | 31.69           | 24.01              | -24.2%       |
| 6144   | 423.96          | 424.44              | 30.75           | 23.56              | -23.4%       |
| 8192   | 423.14          | 422.84              | 30.71           | 23.77              | -22.6%       |
| 10240  | 411.66          | 403.45              | 30.87           | 23.36              | -24.3%       |
| 12288  | 399.37          | 408.61              | 30.42           | 23.64              | -22.3%       |
| 14336  | 405.47          | 404.16              | 30.20           | 23.22              | -23.1%       |
| 16384  | 395.80          | 401.72              | 30.50           | 23.35              | -23.4%       |
| 18432  | 391.88          | 392.80              | 29.83           | 23.03              | -22.8%       |
| 20480  | 384.73          | 383.19              | 30.29           | 23.20              | -23.4%       |
| 22528  | 377.43          | 383.79              | 29.61           | 22.89              | -22.7%       |
| 24576  | 382.00          | 377.81              | 29.76           | 22.94              | -22.9%       |
| 26624  | 367.52          | 374.69              | 29.41           | 22.85              | -22.3%       |
| 28672  | 371.84          | 370.90              | 29.27           | 22.71              | -22.4%       |

Metal decode regression: ~22-24% slower at all measured frontiers.
This matches the PR authors note that the Wave M3 inline-dequant path
dequants the packed row to scratch per attention call before reusing
the stock fp8 attention kernel. The follow-up h8 head-batched Flash
attention PR is expected to close this gap to ~5%.

Prefill t/s is within plus or minus 3% across all frontiers - no
measurable regression on prefill.

### Quality: Teacher-forced PPL + KLD + top-K agreement

Prompt 1: long_context_story_prompt.txt
| Metric            | fp8 (baseline) | turbo3         |
|-------------------|----------------|----------------|
| PPL               | 26.26          | 25.82          |
| KLD mean (nats)   | -              | 0.0980         |
| KLD max (nats)    | -              | 1.1487         |
| top-1 agreement   | -              | 85.10%         |
| top-5 agreement   | -              | 99.22%         |

Prompt 2: long_context_security_prompt.txt
| Metric            | fp8 (baseline) | turbo3         |
|-------------------|----------------|----------------|
| PPL               | 35.67          | 31.79          |
| KLD mean (nats)   | -              | 0.1935         |
| KLD max (nats)    | -              | 5.1627         |
| top-1 agreement   | -              | 83.92%         |
| top-5 agreement   | -              | 99.22%         |

top-5 agreement >= 99.2% on both prompts. The model still ranks the
right tokens at the top of the distribution; argmax shuffles within
the top set. PPL alone is misleading on small samples (turbo3 reports
lower PPL due to quantisation noise acting as regularisation); KLD
reveals the actual distribution drift.

### Memory savings at peak ctx=28672

| KV dtype | kvcache_bytes  | vs fp8    |
|----------|----------------|-----------|
| fp8      | 418,637,200    | baseline  |
| turbo3   | 409,737,232    | -2.1%     |

Note: The CSV reports kvcache_bytes which includes the compressed ring
buffer size. The raw KV shrink is 4.75x as reported by the engine; the
total footprint savings is less dramatic because the compressed ring
buffer (used for KV page eviction) is shared between fp8 and turbo3
modes and dominates at these context sizes. The real savings appear at
longer contexts where the raw KV portion dominates.

### Reproduce

git clone https://github.com/antirez/ds4 && cd ds4
git fetch origin pull/243/head:pr-243 && git checkout pr-243
make ds4-bench

MODEL=/path/to/DeepSeek-V4-Flash-IQ2XXS-chat-v2-imatrix.gguf

Throughput:
DS4_LOCK_FILE=/tmp/ds4-bench.lock ./ds4-bench -m MODEL --backend metal --kv-cache fp8 --chat-prompt-file tests/long_context_story_prompt.txt --ctx-max 28672 --csv fp8.csv
DS4_LOCK_FILE=/tmp/ds4-bench.lock ./ds4-bench -m MODEL --backend metal --kv-cache turbo3 --chat-prompt-file tests/long_context_story_prompt.txt --ctx-max 28672 --csv turbo3.csv

Quality:
DS4_LOCK_FILE=/tmp/ds4-bench.lock ./ds4-bench -m MODEL --backend metal --kv-cache fp8 --ppl-prompt tests/long_context_story_prompt.txt --quality-emit baseline.bin
DS4_LOCK_FILE=/tmp/ds4-bench.lock ./ds4-bench -m MODEL --backend metal --kv-cache turbo3 --ppl-prompt tests/long_context_story_prompt.txt --quality-baseline baseline.bin

TheTom:
Thanks @sztlink for the sm_89 build + regression reproductions on both branches - clean cross-hardware confirmation.

And @shagghiesuperstar for the full M5 Max bench, the throughput delta and KLD numbers line up with my published runs on a different prompt. On the decode regression: the h8 head-batched Flash attention fix is staged at TheTom:pr/03-h8-sparse-v ([diff](https://github.com/antirez/ds4/compare/main...TheTom:ds4:pr/03-h8-sparse-v)) - happy to have you bench it on your M5 Max if you want to verify the gap-close numbers.

TheTom:
Rebased onto current main - merged a fair bit upstream since opening this (DwarfStar rename, PRO variant, runtime-shape refactor that swapped compile-time `DS4_N_HEAD_DIM` constants for runtime macros). Auto-merged most; two small `ds4.h` / `ds4_bench.c` conflicts hand-resolved (kept upstream's `inspect_only` field + my `kv_dtype`; restored the footprint-log calls at bench startup).

Mac Metal + GX10 CUDA both build clean, short smoke runs green on Flash. Planning a full regression run today (`make cuda-regression`, `./ds4_test --logprob-vectors`, `--long-context` NIAH, KLD bench at 8K) to confirm nothing drifted - will post the results here.

PRO variant + turbo3 untested; happy to bench if you have a PRO model handy or want me to skip that combo for now.

antirez:
Thank you. From my side, this is yet not actionable. Compressing only the raw SWA ring can only cause correctness problems without winning us any memory savings: I guess everybody can agree on that, and indeed it was the first step. So the real saving happens once we have the full kv cache compressed, that right now is CUDA only in the next PR, so I can't test it easily. So here the things to take in mind are the following:

* DeepSeek KV cache is already compressed by a factor 4 or 128 factor depending on the layer (apart from non compressed layers). So we are already working with a model that is already runnable with a large context with the use of little memory compared to the memory used by the weights.
* The compression likely makes the kv rows more quantization sensible, since quantization exploits some entropy inefficiency that DS4 is likely to exploit as well. So it is possible that we will see issues.
* This could be mostly interesting for 96GB systems. At the same time, it creates issues from the POV of a different representation for KV cache files, and maintenance complexity.

However there is to say that we could work with other models in the future that don't have the same compression that DeepSeek features, so it could be more useful than it is now. Basically: this could be merged only if we can see that 3 bit compression of the *full* kv cache does not create correctness problems, otherwise why to risk? We already have a 2 bit compressed model, and a model that uses little KV cache. At the same target the natural targets are 128GB systems. So it could be wiser to simply don't attempt kv cache compression.

TheTom:
Hey @antirez, thanks for the comment.

I originally broke it up to make review easier, which is why this one alone reads that way. If you'd rather see the full-KV story in a single diff so the savings and quality evidence land together, I'm happy to do that. Just trying to be respectful of your project and how past PRs have come in.

I'll get you a side-by-side in the meantime while I wait on your response.

antirez:
@TheTom thanks, it's not much about the process itself, is that just part of the PR would never get merged since it is only a risk without the reward, and more than the PR itself, I believe this is a cool experiment, to start with: will 3 bit quantization terribly damage the DeepSeek KV cache? So mostly is a matter of knowing this and experiment with this. Once we have the response, and how little or big damage happens, then there is to evaluate if it is worth it. I'm a bit skeptical on the risk because of DeepSeek already great kv cache compression, but I'm probably adding MiMo V2.5 and there the situation is different, for instance, so in general indeed something to keep in mind, and to play with. That's my POV on the general matter of cache compression, since DwarfStar attempts to be quality-first, at the same time it targets limited hardware so must compress models already too much... It's a complicated balance :) Thanks!

TheTom:
Ah gotcha. Then it makes sense to bring the whole picture into this PR so the A/B comparison lives in one place rather than spread across the stack. I'll follow up with the full experimental view here.

TheTom:
@antirez — quick follow-up with the consolidated picture.

**PR updated since the last comment:**
- Four-PR stack collapsed into this one (turbo3 + turbo4 + h8/sparse-V + comp_cache + HISA, eight commits)
- Title narrowed to `opt-in KV cache compression`
- Description rewritten as an experiment record with the full quality + memory evidence inline

**Evidence summary**

Teacher-forced PPL on `tests/long_context_story_prompt.txt`, 511 scored tokens, GB10 sm_121, Qwen3.6-A3B IQ2XXS:

| KV config                            | ppl       | KLD vs fp8 | top-1   | top-5   |
|--------------------------------------|-----------|------------|---------|---------|
| fp8 (baseline)                       | 15.736332 | —          | —       | —       |
| turbo3 (SWA only)                    | 16.559995 | 0.076420   |  86.89% |  99.61% |
| turbo3 + comp_cache turbo3 (full KV) | 16.351816 | 0.079748   |  87.08% |  99.80% |
| turbo4 + comp_cache turbo3           | 16.135470 | 0.053660   |  89.04% |  99.61% |

KV memory at 256K context, same model and backend:

| KV config                            |          bytes |      GB | vs fp8 |
|--------------------------------------|---------------:|--------:|-------:|
| fp8 (baseline)                       |  3,632,145,808 | 3.38 GB |   100% |
| turbo3 alone (SWA only)              |  3,623,245,840 | 3.37 GB |    99% |
| turbo3 + comp_cache turbo3 (full KV) |  1,004,230,672 | 0.94 GB |    28% |
| turbo4 + comp_cache turbo3           |  1,004,538,896 | 0.94 GB |    28% |

The raw-only `turbo3` row saves essentially nothing on its own (the engine caps the raw SWA ring at 4352 rows regardless of context, so shrinking each row barely moves the total).  The real win lives in `--comp-cache turbo3`, which trims the compressor pool that dominates at long context.

`./ds4_test --long-context` passes with `--kv-cache turbo3`.  The four-question `ds4-eval` gate passes 3/4 with full-KV compression; Q4 (the GPQA Diamond physics chain) regresses from `C` to `A`, which is the same answer your `CONTRIBUTING.md` records as the historical fp8 failure mode on that question — not a new failure shape.

**Where this fits**

Since you mentioned that other models will be supported in the future, this is probably more relevant to DwarfStar's eventual support for models that don't handle their KV cache as well architecturally.  So I'd think of this PR as an experiment log to come back to once those land — or for 96 GB DeepSeek users running multiple applications who are essentially resource-constrained on unified memory.

Issue #210, opened 2026-05-20
Add support for structured outputs


pminervini:
As in https://docs.vllm.ai/en/latest/features/structured_outputs/#online-serving-openai-api and https://developers.openai.com/api/docs/guides/structured-outputs

I tried it with `ds4` using the latest code from `main` but it doesn't seem to be working fine (both DGX Spark and m3 max) -- here's the test:

```python
#!/usr/bin/env python3
from openai import OpenAI
from pydantic import BaseModel, ConfigDict


BASE_URL = "http://127.0.0.1:8000/v1"
API_KEY = "not-needed"
MODEL = "default"


class CalendarEvent(BaseModel):
    model_config = ConfigDict(extra="forbid")

    name: str
    date: str
    participants: list[str]


def main() -> None:
    client = OpenAI(base_url=BASE_URL, api_key=API_KEY)

    print(f"[so-test] base_url={BASE_URL}")
    print(f"[so-test] model={MODEL}")
    print("[so-test] calling client.responses.parse(..., text_format=CalendarEvent)")

    response = client.responses.parse(
        model=MODEL,
        input=[
            {"role": "system", "content": "Extract the event information."},
            {
                "role": "user",
                "content": "Alice and Bob are going to a science fair on Friday.",
            },
        ],
        text_format=CalendarEvent,
    )

    event = response.output_parsed
    print("[so-test] parsed object:")
    print(event.model_dump_json(indent=2))
    print("[so-test] raw response id:", getattr(response, "id", None))


if __name__ == "__main__":
    main()
```

Output:

```bash
$ python3 so-test-cli.py 
[so-test] base_url=http://127.0.0.1:8000/v1
[so-test] model=default
[so-test] calling client.responses.parse(..., text_format=CalendarEvent)
Traceback (most recent call last):
[..]
pydantic_core._pydantic_core.ValidationError: 6 validation errors for CalendarEvent
people
  Extra inputs are not permitted [type=extra_forbidden, input_value=['Alice', 'Bob'], input_type=list]
    For further information visit https://errors.pydantic.dev/2.13/v/extra_forbidden
event
  Extra inputs are not permitted [type=extra_forbidden, input_value='science fair', input_type=str]
    For further information visit https://errors.pydantic.dev/2.13/v/extra_forbidden
day
  Extra inputs are not permitted [type=extra_forbidden, input_value='Friday', input_type=str]
    For further information visit https://errors.pydantic.dev/2.13/v/extra_forbidden
name
[...]

PR #247, opened 2026-05-25
Add constrained decoding for JSON schema response formats

pminervini:
Tentative fix for #210.

This adds a small decoder for OpenAI-compatible `response_format` / Responses `text.format` JSON modes: during generation, tokens that would cause the current output to stop being a valid prefix of the requested JSON object/schema are masked out.

This is roughly the same idea as in llama.cpp’s JSON-schema grammar path, where the schema guides the decoding process. Right now it supports `/v1/chat/completions` and `/v1/responses`, handles `json_object` and `json_schema`, disables thinking for structured JSON outputs (not sure about this, will check how e.g. vllm handles this for LRMs), and keeps the schema out of the prompt ("follow this JSON schema" also seems to work for simple schemas -- example in the issue -- but I would not trust that too much).

Will share a script to stress-test this with schemas of varying complexity.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -543,6 +543,19 @@ typedef struct {
     int cap;
 } tool_schema_orders;
 
+typedef enum {
+    DS4_TEXT_FORMAT_TEXT = 0,
+    DS4_TEXT_FORMAT_JSON_OBJECT,
+    DS4_TEXT_FORMAT_JSON_SCHEMA,
+} ds4_text_format_type;
+
+typedef struct {
+    ds4_text_format_type type;
+    char *name;
+    char *schema_json;
+    bool strict;
+} ds4_text_format;
+
 typedef struct {
     char *role;
     char *content;
@@ -600,6 +613,7 @@ typedef struct {
     int cache_read_tokens;
     int cache_write_tokens;
     ds4_think_mode think_mode;
+    ds4_text_format text_format;
     bool has_tools;
     bool prompt_preserves_reasoning;
     /* For /v1/responses: emit reasoning_summary_* events / fields only when the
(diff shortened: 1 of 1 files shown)


pminervini:
Q: why not e.g. https://github.com/mlc-ai/xgrammar or https://github.com/guidance-ai/llguidance ? @fry69

See convo in #210 -- happy to ditch this and e.g. add -DENABLE_LLGUIDANCE (`llama.cpp` also does that)

nhwaani:
I tested this PR on an M5 Max 128GB and it works well for the main structured-output cases.

Environment:
- MacBook Pro M5 Max, 128GB (`Mac17,7`)
- macOS 26.5
- DeepSeek V4 Flash q2 GGUF
- Server: `./ds4-server --port 8010 --ctx 4096 --tokens 512 --power 70`

What passed:
- `./ds4_test --server`
- `make test`
- chat `response_format: {"type":"json_object"}`
- chat `response_format: {"type":"json_schema"}`
- Responses API `text.format: {"type":"json_schema"}`
- enum / const / integer / boolean schema
- nested array-of-objects schema
- simple `anyOf`

One small edge case:
- For a schema with `minItems: 2` and `maxItems: 2`, the model sometimes generated only whitespace until `finish_reason=length`.
- Cause: leading whitespace is still a valid JSON prefix, so whitespace tokens stayed allowed forever.

I opened a small follow-up PR against this branch:

https://github.com/pminervini/ds4/pull/1

The fix only masks pure whitespace tokens while the output buffer is still only whitespace. Once JSON has started, whitespace is allowed normally.

pminervini:
hey! hope you guys won't mind, I think this may be a better way of enabling support for structured outputs in ds4: https://github.com/pminervini/ds4/pull/1#issuecomment-4557582135 -- will try to look into this over the weekend

pminervini:
Let me close this, will try to propose a better one.

fry69:
FYI: I found some time to toy around with structured outputs and checked this PR out (apparently `gh` has no problems with closed PRs), merged it with current `main` (fec25dafdfe4bdce20980c458f3d1980a2610191) and it passes my small, naive test suite with flying colors 👍 

Looking forward for another stab at this.

PR #256, opened 2026-05-26
speed-bench: add NVIDIA RTX PRO 6000 Blackwell results

imbibekk:
Numbers collected on a single NVIDIA RTX PRO 6000 Blackwell Server Edition (97 GB VRAM, compute capability sm_120), CUDA 13.0, built with `make cuda-generic`. Only GPU 0 is used since `ds4_gpu_init` selects a single device.

Sweep (default flags):

  ./ds4-bench -m ds4flash.gguf \
    --ctx-start 2048 --ctx-max 65536 --step-incr 2048 \
    --gen-tokens 128 \
    --csv speed-bench/rtx_pro_6000_blackwell.csv

The README row reports the 8192-token frontier so it sits next to the existing DGX Spark GB10 row (7047 tokens) for direct comparison. The full 2048..65536 sweep is in the CSV.

Caveat: at long context the engine logs
"CUDA q8 fp16 cache budget exhausted; using q8 kernels" because the 80.76 GiB model plus context buffers leave only about 1 GiB free on a 97 GiB card, so all frontiers reflect the fallback q8 kernel path rather than the q8/fp16 cached fast path.


nullgate:
why the prefill speed is slower than m3 ultra? rtx pro 6000 has better compute and better memory speed.

PR #263, opened 2026-05-27
Advertise deepseek-chat model alias

### LINKED ISSUE #216: Return deepseek-chat in GET models
It would be helpful if deepseek-chat was returned in the list of models

I am able to use deepseek-chat as the model to disable the thinking output but some tools will not let you configure them to use a particular model unless it is included in the set of models.

cat0825:
Fixes #216

## Summary
- include the existing `deepseek-chat` non-thinking alias in `GET /v1/models`
- allow `GET /v1/models/deepseek-chat` to return model metadata for clients that require detail lookup
- add server unit coverage for the alias metadata and route allow-list

## Verification
- make ds4_test
- ./ds4_test --server
- make ds4-server
- ./ds4_test --server

Note: this only changes server model metadata/alias handling; it does not touch inference kernels or backend execution paths.


cat0825:
Merged current `main` in; conflict-free now.

Conflicts were both in `ds4_server.c`: `server_model_alias_known()` (kept your GLM aliases, added `deepseek-chat` to the DeepSeek group) and `send_models()` (kept your GLM/DeepSeek branch split, advertising `deepseek-chat` only in the DeepSeek branch).

Verified: `make` clean, no warnings; `make ds4_test && ./ds4_test --server` → `server: OK`.

@antirez whenever you have a moment.

PR #266, opened 2026-05-27
cuda: increase matmul rows/block from 8 to 16

arpesenti:
## CUDA: increase matmul rows/block from 8 to 16

Doubles the number of output rows processed per block in the warp-level Q8_0, HC-expand, grouped
matmul and MoE gate_up decode kernels, improving GPU occupancy on architectures with high CUDA core
counts. Prefill is unaffected.

### Benchmark (DGX Spark GB10, q2, 128 gen tokens)

| ctx | before | after | Δ |
|------|--------|-------|------|
| 2048 | 14.4 | 18.1 | +25.7% |
| 4096 | 14.9 | 19.1 | +28.0% |
| 6144 | 14.9 | 19.1 | +28.0% |
| 8192 | 14.7 | 18.7 | +27.5% |
| 16384 | 14.6 | 18.4 | +26.7% |
| 32768 | 13.5 | 16.8 | +24.4% |
| 40960 | 13.3 | 16.4 | +23.3% |
| 65536 | 12.5 | 15.3 | +22.7% |

Prefill unchanged across all frontiers (within ±1 t/s noise).

### Changes

- Q8_0/HC-expand/grouped matmul kernels: rows/block 8→16
- MoE gate_up decode kernels: rows/block 128→256
- Updated launch grid dimensions for all affected kernels
- Fixed `tests/cuda_long_context_smoke.c` (missing `comp_kv_f16` arg)

### Verification

- `make cuda-spark` — clean build
- `make cuda-regression` — passed

### DIFF
--- a/ds4_cuda.cu
+++ b/ds4_cuda.cu
@@ -1940,7 +1940,7 @@ __global__ static void matmul_q8_0_preq_warp8_kernel(
         uint64_t out_dim,
         uint64_t blocks,
         int use_dp4a) {
-    uint64_t row = (uint64_t)blockIdx.x * 8u + (threadIdx.x >> 5u);
+    uint64_t row = (uint64_t)blockIdx.x * 16u + (threadIdx.x >> 4u);
     uint32_t lane = threadIdx.x & 31u;
     if (row >= out_dim) return;
     const unsigned char *wr = w + row * blocks * 34;
@@ -1970,7 +1970,7 @@ __global__ static void matmul_q8_0_pair_preq_warp8_kernel(
         uint64_t out1_dim,
         uint64_t blocks,
         int use_dp4a) {
-    uint64_t row = (uint64_t)blockIdx.x * 8u + (threadIdx.x >> 5u);
+    uint64_t row = (uint64_t)blockIdx.x * 16u + (threadIdx.x >> 4u);
     uint32_t lane = threadIdx.x & 31u;
     if (row >= out0_dim && row >= out1_dim) return;
     float acc0 = 0.0f;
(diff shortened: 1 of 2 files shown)


gundemirbas:
Any progress on merging this PR?

Smallfu666:
▎ Heads up: this change produces incorrect output on DGX Spark (GB10) — I built it and the model emits pure garbage tokens (endless repeated CJK/random tokens) where main answers coherently, same prompt with --temp 0.

▎ The problem is that the kernels only get their row index rewritten, but the reduction is still full-warp. In e.g. matmul_q8_0_preq_warp8_kernel the row becomes

▎ uint64_t row  = (uint64_t)blockIdx.x * 16u + (threadIdx.x >> 4u);
▎ uint32_t lane = threadIdx.x & 31u;      // unchanged
▎ ...
▎ for (uint64_t b = lane; b < blocks; b += 32u) { ... }   // stride 32, unchanged
▎ acc = warp_sum_f32(acc);                // full 32-lane __shfl_down_sync(0xffffffff, ...)
▎ if (lane == 0) out[row] = acc;          // unchanged

▎ With >> 4, two adjacent output rows now share one physical 32-lane warp (row R on lanes 0–15, row R+1 on lanes 16–31). warp_sum_f32() reduces across all 32 lanes, so it sums the partials of two different rows together, and because only lane == 0 writes, out[R] gets the corrupted mixed sum while out[R+1] is never written at all. Same issue in the MoE gate_up path (rr<8, blockIdx.x*256 + row_lane + rr*16 feeding a still-32-lane reduction). The benchmark table looks great because ds4-bench measures t/s, not output correctness.

▎ To do 16 rows/block correctly the reduction has to become per-16-lane too: lane = threadIdx.x & 15, stride b += 16u, a 16-wide shuffle reduction (offsets 8→1), and if ((threadIdx.x & 15) == 0) for the write. Whether that's actually faster than the current 8-rows/block (half the lanes per reduction) would need re-measuring — with a correctness check this time.

▎ (Verified on GB10, 128 GB, q2 IQ2XXS model.)

PR #264, opened 2026-05-27
Add wide-token MoE prefill tiles (n64/n128 mul_mm_id)

beamivalice:
Hi, I'm experimenting with many things trying to speed up the prefill and i've nailed one on that can increase the prefill upto +62% for short 2048 burst and ~30% sustain until 32k context, ~16% for 64k context. The thing is apple chip is better at use wider token tiles compared to the default =32. Increasing this number help reducing the bottleneck routed through moe from 2742 ms (59%) to 1328 ms (41%).

As a bonus, we also gained ~4% token gen speed improvement at 64k but it could be noise. I'm using M5 Max 128GB by the way. I do not know if this will effect any other machine so you guys can use DS4_METAL_MOE_TILE_MAX=XX to optimize for yours.

Thanks @antirez for this wonderful project.

image

(AI help summarize below)
--- 

# PR: MoE prefill wide token tiles (n64 / n128 `mul_mm_id`)

**Branch:** `moe-wide-tiles`  
**Base:** `e8e8779` (current `main`)

## Summary

Adds wider **token tiles** (64 and 128) to the batched routed-MoE prefill path (`kernel_mul_mm_id_*`) for **Q4_K**, **Q2_K**, and **IQ2_XXS** (including F16-RHS down). Aligned prefill chunks (e.g. 2048 tokens) use fewer, larger threadgroups, which cuts per-launch overhead and improves simdgroup MMA utilization on the dominant prefill cost.

**Scope:** `metal/moe.metal` (kernel template exports) + `ds4_metal.m` (tile selection, pipeline dispatch, encode). No flash, gate+up pair, docs-only churn, or bench scripts in the commit.

## What changed

- New Metal instantiations: `kernel_mul_mm_id_{q2_K,q4_K,iq2_xxs}_{f32,f16}_n64` and `_n128`.
- `ds4_gpu_moe_mm_tile_n()` picks **128** when `n_tokens % 128 == 0`, else **64** when `% 64 == 0`, for the quant types above.
- `ds4_gpu_routed_mm_pipeline_for_tile()` / `ds4_gpu_routed_mm_f16_rhs_pipeline_for_tile()` select the matching pipeline.
- `ds4_gpu_encode_mul_mm_id_mapped_tile()` passes `moe_mm_tile_n` into the grid (`ne21 / tile_n` threadgroups on X).

Decode and tiny batches are unchanged (still `mul_mv_id` or 32-tile fallback).

## A/B for maintainers

| Env | Effect |
|-----|--------|
[...]

PR #268, opened 2026-05-27
Document wide MoE prefill tile setting

nhwaani:
Documents the wide routed-MoE prefill tile behavior added by PR #264 and the DS4_METAL_MOE_TILE_MAX A/B knob.\n\nDocs-only change.


nhwaani:
@antirez

PR #273, opened 2026-05-27
Harden DSML tool continuations

harrychin:
Summary:
- recover terminal DSML tool calls emitted after an unclosed thinking block without executing quoted or tentative DSML
- harden malformed/unterminated DSML continuation handling so the model gets a visible tool-error retry path
- stream tool-use headers as soon as invocation names are known while withholding incomplete arguments until final normalized JSON is available

Tests:
- ./ds4_test --server
- make ds4-server

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -18,7 +18,6 @@
 #include <float.h>
 #include <fcntl.h>
 #include <limits.h>
-#include <math.h>
 #include <netinet/in.h>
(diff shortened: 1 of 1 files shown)


harrychin:
This change trades the stream everything approach with a buffer until checked approach as I was trying to daily drive with the model but Pi or Claude would sometimes stall when the model made tool continuation errors.

PR #302, opened 2026-05-30
Structured outputs (JSON etc.) via llguidance

pminervini:
This should address #210.

Here is another stab at structured outputs, but instead of the custom JSON-schema decoder from #247, this uses llguidance behind an optional make LLGUIDANCE=1 build. So the implementation is closer to how llama.cpp does it, and it supports the OpenAI JSON modes plus regex, Lark, and raw llguidance grammars for both chat completions and Responses.

 Compared to #247, this should be less ds4-specific code to maintain and it covers more formats. Current caveat: structured outputs disable thinking for that turn (looking into this) and are not combined with tools.

I tested the llguidance build, ds4_test, ds4-server, and a local structured-output stress-sweep against ds4-server. Poking @fry69 and @nhwaani since they were interested.


pminervini:
Quick update: structured outputs no longer disable thinking: ds4 now lets the model finish </think> and then applies the llguidance constraint to the final answer.

PR #305, opened 2026-05-30
cpu: add x86 AVX2/AVX512 SIMD fast paths for quantized MoE dot products

hexxyan:
## Summary

The x86 CPU path currently falls back to scalar code for all quantized dot products (Q2_K, Q4_K, IQ2_XXS) used in routed MoE expert inference. This PR adds AVX2 and AVX512BW SIMD fast paths, and explicit Makefile targets for building ISA-specific binaries.

The existing ARM NEON, scalar, Metal, and CUDA paths are unchanged.

### Why does this matter?

Each MoE layer runs dot products against multiple expert weight blocks (typically 8 selected experts × gate/up/down projections). On x86 these are currently pure scalar loops — one multiply-accumulate per cycle per value. AVX2 processes 8×float32 or 16×int8 per cycle; AVX512 doubles that again. For CPU-only inference on large-RAM x86 machines, this should make routed expert evaluation substantially faster, directly improving tok/s.

### What changed?

**New SIMD dot product kernels** (compile-time selected via `__AVX2__` / `__AVX512F__+__AVX512BW__`):

| Kernel | AVX2 | AVX512BW | AVX512_VNNI |
|--------|------|----------|-------------|
| `Q2_K × Q8_K` | ✓ | ✓ | — |
| `Q4_K × Q8_K` | ✓ | ✓ | ✓ (dpbusd) |
| `IQ2_XXS × Q8_K` | ✓ | ✓ | — |
| `IQ2_XXS pair × Q8_K` | ✓ (delegates) | ✓ (delegates) | — |

Key implementation details:
- Q2_K: 2-bit extraction via shift+mask, `madd_epi16` for signed dot product
- Q4_K: `maddubs_epi16` (unsigned × signed byte), AVX512 variant with `dpbusd_epi32` when VNNI is available
- IQ2_XXS: grid/sign lookup stays scalar (LUT-dependent), only int8 dot uses SIMD via `madd_epi16`
- Proper zero-extension (`ds4_zext_i128_to_i256`) for AVX512 `cvtepi8_epi16` — avoids undefined upper-lane garbage

**New Makefile targets** for explicit ISA-specific binaries:

```
make cpu               # default: -march=native, matches build machine
make cpu-avx2          # fixed AVX2 binary (ds4-avx2, ds4-server-avx2, ...)
make cpu-avx512        # fixed AVX512BW binary
make cpu-avx512-vnni   # fixed AVX512BW+VNNI binary
```

Each SIMD variant uses separate `build/cpu-<suffix>/` object directories to avoid `.o` file conflicts. Architecture gating uses `$(filter x86_64 amd64,$(UNAME_M))` — works across both Darwin and Linux. Non-x86 hosts get a clear error message.

**New test file**: `tests/test_quant_dot.c` replaces `tests/test_q2k_dot.c`, covering all three quant formats:
- Block size validation
- Known-value dot product
- 100-seed random reference comparison (scalar vs SIMD)
- Multi-block accumulation
- Q4_K nibble edge cases

### Validation status

**Kernel correctness** — tested via unit tests only:
- `make quant-dot-test` — 11/11 pass (scalar vs AVX2 cross-checked on 100 random blocks per format)
- `make cpu-avx2` — builds all five binaries
- `make ds4_cpu.o -B` — only pre-existing warnings
- AVX512 paths compile-clean with `-mavx512f -mavx512bw` and `-mavx512vnni`

**E2E model inference not tested** — I do not have enough RAM to load a DeepSeek V4 model and run actual inference. The dot product kernels are verified correct in isolation, but real-world tok/s and output quality need validation on a large-RAM machine.

**Community testing needed.** If you have a large-RAM x86 machine (AVX2 or AVX512):

```sh
make quant-dot-test    # unit test first
make cpu               # or make cpu-avx2 / make cpu-avx512
./ds4 --cpu -m /path/to/model.gguf -p "Explain merge sort in one paragraph."
```

Useful reports:
- CPU model and ISA support (`lscpu | grep -i avx`)
- Model quant format (Q2_K, Q4_K, IQ2_XXS, etc.)
- Whether `make quant-dot-test` passes
- First-token latency and tokens/sec
- Whether output looks sane (no garbage, no crash)

### Notes

- No runtime CPU feature detection is added. SIMD paths are compile-time via compiler macros.
- This does not change CUDA or Metal paths.

## Test Plan

- [x] `make quant-dot-test` — 11/11 pass (AVX2 cross-checked against scalar)
- [x] `make cpu` — builds successfully, default behavior unchanged
- [x] `make cpu-avx2` — builds `ds4-avx2` and siblings with fixed `-mavx2` (no `-march=native`)
- [x] AVX512 compile-only check (ds4.c + test file, with/without VNNI)
- [x] `git diff --check` clean
- [x] E2E model inference on large-RAM x86 with AVX2 (need community help)
- [x] AVX512 runtime validation (need community help)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,13 @@
 CC ?= cc
 UNAME_S := $(shell uname -s)
+UNAME_M := $(shell uname -m)
 
 ifeq ($(UNAME_S),Darwin)
+ifeq ($(UNAME_M),arm64)
 NATIVE_CPU_FLAG ?= -mcpu=native
+else
+NATIVE_CPU_FLAG ?= -march=native
+endif
 else
 NATIVE_CPU_FLAG ?= -march=native
 endif
@@ -33,7 +38,7 @@ CUDA_LDLIBS ?= -lm -Xcompiler -pthread -L$(CUDA_HOME)/targets/sbsa-linux/lib -L$
 METAL_LDLIBS := $(LDLIBS)
 endif
 
-.PHONY: all help clean test cpu cuda cuda-spark cuda-generic cuda-regression
+.PHONY: all help clean test cpu cpu-avx2 cpu-avx512 cpu-avx512-vnni cpu-simd-build cuda cuda-spark cuda-generic cuda-regression
 
 ifeq ($(UNAME_S),Darwin)
 all: ds4 ds4-server ds4-bench ds4-eval ds4-agent
@@ -42,6 +47,9 @@ help:
 	@echo "DS4 build targets:"
 	@echo "  make              Build Metal ./ds4, ./ds4-server, ./ds4-bench, ./ds4-eval, and ./ds4-agent"
 	@echo "  make cpu          Build CPU-only ./ds4, ./ds4-server, ./ds4-bench, ./ds4-eval, and ./ds4-agent"
+	@echo "  make cpu-avx2     Build CPU-only with AVX2 (x86_64 only)"
+	@echo "  make cpu-avx512   Build CPU-only with AVX512BW (x86_64 only)"
+	@echo "  make cpu-avx512-vnni Build CPU-only with AVX512BW+VNNI (x86_64 only)"
 	@echo "  make test         Build and run tests"
 	@echo "  make clean        Remove build outputs"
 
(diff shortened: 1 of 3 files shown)


OceanX89:
I've tried it on my old EPYC 7552 with 8-channel 512 GB DDR4 2666:
I've used the Q4K model.

lspci:
lspci

make quant-dot-test:
Test

master:
Master

PR-305:
PR-305

PR-305 with AVX2:
PR-305-AVX2

Sorry for the different lengths. I haven't found a way to stop this model from thinking.

This branch _does_ increase the CPU performance and most of all: Using AVX2 or AVX512 is more efficient.

But overall performance is still quite poor for a 153 GB model.
For comparison: With llama.cpp and and DeepSeek-V3-0324-GGUF-UD (296GB), I get avg 5,2 T/s on CPU with AVX2.
It could be due to the model, but newer models are _usually_ faster than older ones.

OceanX89:
The previous test had a CPU so overkill, that AVX didn't add much beside efficency.
So here a test with a underpowered CPU:

AMD EPYC Emebdded 3151 (4c/8T, Zen1) on a ITX Mainboard with 256GB DDR4 2133MHz in 2 channels:

Master:
NonAVX

PR-305 with AVX2:
avx2

**--> With this underpowered CPU, your branch results in 31% higher performance!**

In this case the 4 cores are a larger bottleneck than the memory bandwith. So AVX2 helps a lot! 👍👍

PR #306, opened 2026-05-31
metal: simdgroup MMA mini-GEMM for decode MoE [experimental]

hexxyan:
## Summary

Experimental optimization for the DeepSeek-V4 decode MoE path: replace scalar dot-product kernels with tiled `simdgroup_multiply_accumulate` mini-GEMM for the 6 selected experts.

**New Metal kernels** (`metal/moe.metal`):
- `kernel_mul_mm_selected_pair_swiglu` — fused gate+up projection with shared RHS activation tile and SwiGLU activation. Template instantiations for Q2_K, Q4_K, IQ2_XXS × F32/F16.
- `kernel_mul_mm_selected_sum` — down projection that accumulates all 6 experts directly into the output row, eliminating the separate sum kernel.

**Host dispatch** (`ds4_metal.m`):
- Opt-in via `DS4_METAL_DECODE_MOE_MINIGEMM=1` environment variable
- Default decode path completely unchanged
- Stage profiling via `DS4_METAL_MOE_STAGE_PROFILE` for A/B comparison

**Validation tool** (`tools/validate_metal.sh`):
- Offline Metal shader compilation check using `xcrun metal`
- Mirrors the default source concatenation used by `ds4_gpu_full_source()` (base header + all `.metal` files)

## Verification

- [x] C/ObjC build passes (`make ds4`)
- [x] `ds4_test --server` passes
- [x] `test_q4k_dot` passes
- [x] `ds4-eval --self-test-extractors` passes
- [x] Full concatenated Metal source (~9.6k lines) compiles cleanly (unused-function warnings only)
- [ ] Apple Silicon runtime pipeline creation
- [ ] Numerical correctness vs scalar path
- [ ] tok/s benchmark (baseline vs mini-GEMM)

## How to Test

```bash
# Build
make ds4

# Validate Metal shaders offline
tools/validate_metal.sh

# Baseline (scalar path)
./ds4-server -m <model.gguf>

# Mini-GEMM prototype
DS4_METAL_DECODE_MOE_MINIGEMM=1 ./ds4-server -m <model.gguf>

# With stage profiling for A/B comparison
DS4_METAL_MOE_STAGE_PROFILE=1 ./ds4-server -m <model.gguf>
DS4_METAL_DECODE_MOE_MINIGEMM=1 DS4_METAL_MOE_STAGE_PROFILE=1 ./ds4-server -m <model.gguf>
```

## Notes

This is an **experimental prototype**. The mini-GEMM path is disabled by default and only activates when the env var is set. No performance claims yet — Apple Silicon testing is needed to confirm whether the simdgroup MMA tiles outperform the existing fused scalar dot-product path for the decode MoE workload (6 experts × small batch).

The hypothesis is that for decode (small M, large K), simdgroup MMA tiles can better utilize the GPU matrix hardware than per-expert scalar dot products. But the tile overhead may dominate at small problem sizes, so empirical measurement is essential.

### DIFF
--- a/ds4_metal.m
+++ b/ds4_metal.m
@@ -833,6 +833,21 @@ static int ds4_gpu_use_mpp_attn_out_low_matmul(void) {
     return ds4_gpu_mpp_available();
 }
 
+static int ds4_gpu_use_decode_moe_minigemm(void) {
+    static int initialized;
+    static int enabled;
+    if (!initialized) {
+        enabled = ds4_gpu_env_bool("DS4_METAL_DECODE_MOE_MINIGEMM") > 0;
+        if (enabled) {
+            fprintf(stderr,
+                    "ds4: experimental Metal decode MoE mini-GEMM enabled by "
+                    "DS4_METAL_DECODE_MOE_MINIGEMM\n");
+        }
+        initialized = 1;
+    }
+    return enabled;
+}
+
 enum {
     DS4_METAL_ATTN_OUT_MPP_TILE_N = 64,
 };
@@ -2584,6 +2599,18 @@ static int ds4_gpu_encode_mul_mm_id_mapped_tile(
         NSUInteger                  src1_off,
         id<MTLBuffer>               dst,
         NSUInteger                  dst_off);
+static int ds4_gpu_encode_mul_mm_selected_sum(
+        id<MTLCommandBuffer>        cb,
+        id<MTLComputePipelineState> mm_pipeline,
+        const ds4_gpu_mul_mm_id_args *mm_args,
+        id<MTLBuffer>               src0,
+        NSUInteger                  src0_off,
+        id<MTLBuffer>               src1,
+        NSUInteger                  src1_off,
+        id<MTLBuffer>               ids,
+        NSUInteger                  ids_off,
+        id<MTLBuffer>               dst,
+        NSUInteger                  dst_off);
 
 typedef struct {
     int32_t  ne11;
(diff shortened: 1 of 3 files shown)


hexxyan:
@nhwaani would you be willing to give this experimental PR a quick spin on Apple Silicon?

This adds an opt-in decode MoE mini-GEMM path using `simdgroup_multiply_accumulate`. It is disabled by default, so the existing path should be unchanged.

The most useful checks would be:

1. Does it build and run without crashing?

```bash
DS4_METAL_DECODE_MOE_MINIGEMM=1 ./ds4-server -m <model>
```

2. Do you see any obvious numerical/output issues compared with the default path?

3. If you have time, a quick tok/s comparison would be amazing:

```bash
./ds4-server -m <model>
DS4_METAL_DECODE_MOE_MINIGEMM=1 ./ds4-server -m <model>
```

Optional stage profile:

```bash
DS4_METAL_MOE_STAGE_PROFILE=1 ./ds4-server -m <model>
DS4_METAL_DECODE_MOE_MINIGEMM=1 DS4_METAL_MOE_STAGE_PROFILE=1 ./ds4-server -m <model>
```

No pressure at all. Thanks!

lobanov:
@hexxyan I gave it a spin on M5 Max 128GB.

1. It does build on metal target, and the server starts with `q2-imatrix` model

2. No obvious output issues, response is comprehensible.

3. I ran my usual speed test: `./ds4 --ctx 100000 --prompt-file ../ds4/speed-bench/promessi_sposi_1k.txt -n 1000`, where _1k file is the first 1000 lines of the canonical text from the repo.

Reported figures with `DS4_METAL_DECODE_MOE_MINIGEMM=1`:
`ds4: prefill: 426.07 t/s, generation: 14.61 t/s` against the baseline of `ds4: prefill: 425.82 t/s, generation: 30.13 t/s` without the variable.

4. First four tests of `ds4-eval` passed with `DS4_METAL_DECODE_MOE_MINIGEMM=1`.

5. Adding `DS4_METAL_MOE_STAGE_PROFILE=1` to the same `ds4` command but with `-n 100` produces a lot of debug logs and slower result `ds4: prefill: 382.31 t/s, generation: 10.49 t/s`. Attaching:
[stage-profile.log](https://github.com/user-attachments/files/28671385/stage-profile.log)

hexxyan:
@lobanov Thanks a lot for testing this on M5 Max — the numbers are very clear.

The stage profile confirms it: this mini-GEMM path spends ~0.76 ms/layer in gate+up and ~0.54 ms/layer in down, totaling ~1.36 ms per MoE layer. Across 43 MoE layers that's ~58.7 ms, which already exceeds the baseline's full per-token budget of ~33 ms. So the 2× slowdown is fully explained.

The root cause is that the baseline decode path is already highly specialized — fused gate+up+SwiGLU matvec (`kernel_mul_mv_id_iq2_xxs_pair_swiglu_f32`) plus direct 6-expert down accumulation (`kernel_mul_mv_id_*_sum6_f32`). This prototype forces batch=1 matvec into 64×8 MMA tiles, paying dequantization → threadgroup store → barrier → tile load → barrier per K-step, which can't be amortized at M=1.

The hypothesis was wrong for single-token decode. I'll close this as a useful negative result. The MMA approach may still make sense for batched decode / continuous batching where tile overhead gets amortized, but that's a future direction, not this PR.

Thanks again for the detailed benchmark.

Issue #293, opened 2026-05-30
Distributed CUDA worker host-registers the whole GGUF, not just its layer slice → Q4 OOM on 128GB DGX Spark


ngt1999:
## Summary
A distributed CUDA worker host-registers the ENTIRE gguf mapping for device
access instead of only its assigned --layers slice. A Q4 worker therefore needs
~150GB and fails on a 128GB machine, even though its slice is only ~78GB.
Q2 (~81GB whole file) fits and runs fine.

## Setup
- Coordinator: MacBook Pro M5 Max 128GB (Metal), --layers 0:20
- Worker: NVIDIA DGX Spark (GB10, sm_121) 128GB (CUDA), --layers 21:output
- 5GbE link; route forms fine ("complete route ready")
- Same commit on both; q4-imatrix / q2-imatrix from download_model.sh

## Q4 (fails)
ds4: restricting cuda model map to layers 21:output (48 spans, 78.21 GiB tensor span)
ds4: CUDA host registration skipped: out of memory
ds4: cuda backend initialized for graph diagnostics
(coordinator) prompt processing failed: cuda layer-slice failed

## Q2 (works)
ds4: restricting cuda model map to layers 21:output (48 spans, 41.08 GiB tensor span)
ds4: CUDA registered 80.76 GiB model mapping for device access
prefill: 9.36 t/s, generation: 16.59 t/s

## Likely cause
"registered 80.76 GiB" ≈ the whole Q2 file while the slice is only 41 GiB, so
registration covers the full mmap, not the slice. For Q4 the whole file (~150GB)
exceeds the 128GB worker, so cudaHostRegister OOMs and the slice can't run.
Registering only the slice's pages (~78GB) would fit and let Q4 run distributed,
which is the whole point of the layer split. GB10 has coherent unified memory,
so full host pinning may be avoidable on this platform entirely.

hexxyan:
Hi @ngt1999 — I opened PR #317 that addresses the full-GGUF registration issue you reported.

**Root cause:** `ds4_gpu_set_model_map_spans()` was passing the full file size to every fallback path. Even though the span calculation was correct, all downstream code (`cuda_model_range_ptr`, `cuda_model_range_is_cached`, `cuda_model_direct_fallback_ptr`, Q2 tail expansion) assumed the full mmap was accessible — which is why Q2 "worked" (80.76 GiB registered ≈ whole file) but Q4 OOMed (150 GiB > 128 GB).

[...]

PR #316, opened 2026-05-31
[Superseded by #317] fix: distributed CUDA span cover

hexxyan:
## Summary

Distributed CUDA workers with `--layers 21:output` should only access the pages covering their assigned layer spans (~78 GiB), not the entire GGUF mmap (~150 GiB for Q4). Currently all direct/full-file fallback paths assume the full model is accessible, causing OOM on 128 GB DGX Spark workers.

## Root Cause

`ds4_gpu_set_model_map_spans()` set `g_model_registered_size` to the full file size. Every fallback path (`cuda_model_range_ptr`, `cuda_model_range_is_cached`, `cuda_model_direct_fallback_ptr`, `cuda_model_copy_chunked`, Q2 tail expansion) used this to assume full-model access.

For Q2 that "works", it still registered 80.76 GiB when only 41 GiB was the slice. For Q4, the 150 GiB full file exceeds the 128 GB worker memory.

### New state: span cover offset/end (not just a flag)
```c
g_model_span_cover_offset  // page-aligned start of span cover
g_model_span_cover_end     // page-aligned end of span cover
g_model_span_only          // 1 when initialized from _spans
```

### New helper: `cuda_model_range_in_span_cover(offset, bytes)`
Returns `true` iff `[offset, offset+bytes)` falls within `[cover_offset, cover_end)`. When not in span-only mode, always returns `true`. Includes overflow guard.

### All guarded paths

| Function | What was guarded |
|----------|-----------------|
| `cuda_model_range_ptr` | `g_model_device_owned`/`g_model_registered` fallback, `g_model_hmm_direct` fallback, `DS4_CUDA_DIRECT_MODEL` fallback |
| `cuda_model_range_is_cached` | Early return when `registered`/`hmm_direct` — now also checks span cover |
| `cuda_model_direct_fallback_ptr` | Now takes `bytes` param and checks span cover (called from fd cache path) |
| `cuda_model_copy_chunked` | Early return when `registered` — now skips when `span_only` |
| `cuda_model_range_register_mapped` | Q2 tail expansion heuristic — now skips when `span_only` |

### Cleanup
Both `cuda_model_set_host_map` and `ds4_gpu_cleanup` reset all span-only state.

## New log output

```
ds4: CUDA span cover 78.21 GiB of 150.00 GiB model (48 spans, offset 72.00 GiB)
```

## Testing

- **CUDA compile verified**: `nvcc -arch=sm_90` compiles `ds4_cuda.cu` cleanly (zero warnings, zero errors, 5.3 MB ELF object) via `nvidia/cuda:13.0.0-devel-ubuntu24.04` Docker image
- Non-CUDA build is unaffected
- **Still needs**: GB10/Q4 distributed e2e validation with `--layers` to confirm the fix resolves the OOM

Refs: #293

### DIFF
--- a/ds4_cuda.cu
+++ b/ds4_cuda.cu
@@ -82,6 +82,9 @@ static uint64_t g_model_direct_align = 1;
 static uint64_t g_model_file_size;
 static int g_model_cache_full;
 static int g_model_mapping_failure_notice_printed;
+static int g_model_span_only;
+static uint64_t g_model_span_cover_offset;
+static uint64_t g_model_span_cover_end;
 static cudaStream_t g_model_prefetch_stream;
 static cudaStream_t g_model_upload_stream;
 static cublasHandle_t g_cublas;
@@ -153,7 +156,8 @@ static const char *cuda_model_range_ptr_from_fd(
         uint64_t offset,
         uint64_t bytes,
         const char *what);
-static const char *cuda_model_direct_fallback_ptr(const void *model_map, uint64_t offset);
+static const char *cuda_model_direct_fallback_ptr(const void *model_map, uint64_t offset, uint64_t bytes);
+static int cuda_model_range_in_span_cover(uint64_t offset, uint64_t bytes);
 static uint64_t cuda_model_cache_limit_bytes(void);
 static uint64_t cuda_model_local_model_limit_bytes(void);
 static int cuda_model_cache_limit_explicit(void);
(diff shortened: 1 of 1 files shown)


hexxyan:
Superseded by a clean PR (force-push history cleanup).

hexxyan:
Superseded by #317 (clean branch, same code).

PR #326, opened 2026-06-02
Add read-only introspection endpoints: /health, /info, /config

shagghiesuperstar:
## What

Three small read-only HTTP endpoints on the ds4-server for operators, dashboards, and load balancers that need server state without paying for a model call:

```
GET /health   status, uptime, in-flight clients, context, model, kv cache
GET /info     engine identity snapshot
GET /config   read-only reflection of server_config
```

Sample response (live ds4-server):

```json
{"status":"ok","uptime_s":9,"clients_in_flight":1,"context_length":131072,"default_max_tokens":393216,"model":"DeepSeek V4 Flash","kv_cache":{"enabled":true,"dir":"/Volumes/OWC_MODELS_TB5/DS4/cache","budget_bytes":53687091200}}
```

## Why

The server previously exposed only the OpenAI/Anthropic endpoints and `/v1/models`. There is no way for an operator to ask "is this server healthy?" or "what is the active context length?" without either issuing a chat call (expensive) or scraping `/v1/models` (wrong shape, counts as a model registry hit).

Downstream consumers (a small open-source web dashboard I maintain, shagghiesuperstar/ds4-dashboard) currently fall back on speculative probes (`/telem`, `/metrics`) that always 404. `/health` makes that path deterministic.

## How

- `start_time` (`time_t`) added to `struct server`, set in `main()` next to the other init lines.
- Three `send_*` helpers added right after `send_models`/`send_model`, mirroring their style: `buf` assembly, `http_response`, no allocations past `buf`, no model access.
- Three new dispatch branches in `client_main`, placed before the `/v1/models/` prefix branch so `/health`, `/info`, `/config` cannot be shadowed.
- All fields used are already exposed through public ds4 engine API or directly on `struct server`. No new headers, no new dependencies.

## Tested

- `make ds4-server` clean (no warnings under `-Wall -Wextra -std=c99`).
- `./ds4_test --server` passes (added two new unit tests covering dispatch disjointness and JSON shape).
- Live-tested against a running ds4-server:
  - `GET /health`, `GET /info`, `GET /config` return valid JSON.
  - `GET /v1/models` and `POST /v1/chat/completions` unchanged.
  - Unknown paths still return 404.

## Diff size

151 lines added (150 in `ds4_server.c`, 1 in `.gitignore` for the `test_q4k_dot` test binary that was missing from the ignore list).

Happy to split into separate PRs per endpoint, or rename them (`/v1/health`, etc.) if you'd prefer them grouped under a versioned prefix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ __pycache__/
 /misc/
 .*.swp
 .DS_Store
+/tests/test_q4k_dot
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -7708,6 +7708,7 @@ struct server {
     visible_live_state thinking_live;
     bool disable_exact_dsml_tool_replay;
     bool enable_cors;
+    time_t start_time;
     pthread_mutex_t tool_mu;
     pthread_mutex_t mu;
     pthread_cond_t cv;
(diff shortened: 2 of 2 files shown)


shagghiesuperstar:
Following up with a purely technical angle prompted by ecosystem activity:

PR #374 (mcmalayalam's `/health` for llama-swap integration) landed recently — it adds a minimal health endpoint specifically because llama-swap needed a way to know when DS4 is ready to serve. That's independent, third-party confirmation that these endpoints aren't hypothetical: the ecosystem already hit the wall without them.

This PR extends beyond #374's scope in two ways that matter for production deployments:

**`/info`** — Client-side model discovery without hitting `/v1/models`. Any orchestrator (llama-swap, OpenRouter-style routers, multi-instance load balancers) needs to know `context_length` and `default_max_tokens` before sending requests. Currently they have to either parse CLI flags or probe via a chat completion — both fragile.

**`/config`** — Configuration-drift detection across multi-instance fleets. In any deployment with >1 server instance, you need to verify that all instances share identical config. Without a read-only config reflection endpoint, the only way to check is comparing launchd plists or CLI invocations — error-prone and unobservable at runtime.

All three endpoints remain zero-cost (no inference, no allocations beyond buf, no model access) — same footprint as the existing dispatch logic.

If you'd prefer a narrower merge, I can rebase to just `/info` and `/config` since #374 already covers the basic health-check. Either way, the code is written and tested.

rosmur:
Bumping this as an interested user. Will this be merged?

PR #327, opened 2026-06-02
checkpoint cache support with optional tail-cache

nullgate:
ds4c-point-checkpoint

https://github.com/nullgate/ds4cache/blob/ds4c-pr-poc-github/DS4CACHE.md


overstrike:
Brief manual testing on a 128GB M5 Max suggests this is a huge win for all but the smallest contexts. In my setup, prefill duration eventually climbs to minutes without this patch, and remains limited to seconds with it.

PR #332, opened 2026-06-03
server: fix implicit </think> close — suppress DSML from streaming output

### LINKED ISSUE #318: Unclosed </think> before tool call stalls agent session (finish=stop, no tool call delivered)
This project is amazing work, first of all- and has been my daily driver. Using the Q2 on a Mac Studio M4 Max 128 GB. But I have noticed this recurring issue:

#167 was closed (same issue) but the fix leaves sessions in a worse state than the original bug. When the model produces a DSML tool call without closing </think> first, the server now logs "thinking not closed, ignoring DSML in reasoning" and returns finish=stop with no tool calls and no useful text. Opencode and pi both just stop and sit there until I nudge it. Typically I explain the tool call failed and to try again- or run the command myself. Not ideal.

Here's a log snip from a session:

> ds4-server: thinking not closed, ignoring DSML in reasoning
> 0531 21:29:28 ds4-server: chat ctx=104990..105039:49 gen=80 TOOLS THINKING finish=stop 4.779s
> 

Opencode shows that the model intended the tool call. (I have not seen a model "thinking _about_ tool calling" and generating the tool token sequence. That seems very unlikely.) Most probable is that it just dropped the </think> closing tag. The fix seems like: when </think> is absent but DSML is present, treat the DSML start as an implicit whoopsie thinking boundary instead of discarding the call. Pre-DSML text becomes reasoning, DSML parses normally.

I checked open PRs and PR #273 does fix this but bundles a _lot_ of other stuff (ASCII DSML variants, tool result reordering, argument unwrapping... none of which I have issues with personally). I made a fork with a targeted change and it's about 20 lines in **parse_generated_message_ex**. I'm can send the PR if that sounds good.

Thanks

TipKnuckle:
Fixes #318.

When the model emits a DSML tool call without closing `</think>` first, the original fix (#500c7df) correctly parses the tool call in `parse_generated_message_ex` but the three live streaming paths (OpenAI, Anthropic, Responses) still had a problem: since `thinking.inside` stays true for the whole generation, the raw DSML markers get streamed to the client as reasoning content. Clients like pi and opencode end up displaying the tool call XML in the thinking block.

This commit applies the same implicit-close logic to each streaming state machine. On the final update, if there's no `</think>` but a DSML block is present, the reasoning is clipped before the DSML start and the state transitions to TEXT mode instead of SUPPRESS — same path as a normal `</think>`. For the Responses path, `reasoning_closed_naturally` is also set so the reasoning item comes back as `"completed"` rather than `"incomplete"`.

The two commits together cover both streaming and non-streaming clients. Tested on pi and opencode (Anthropic endpoint) against the q2-imatrix on M4 Max 128GB — the implicit close case now produces a clean thinking block with no leaking DSML, and the tool call executes normally. fry69 has been running the first commit for 2+ hours without regressions (see issue).

Changes are limited to the three `STREAM_THINKING` branches in `ds4_server.c`, 23 lines total.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -5890,9 +5890,13 @@ static bool openai_sse_stream_update(int fd, server *s, const request *r, const 
         }
 
         const char *close = strstr(raw + st->emit_pos, "</think>");
+        const char *dsml_implicit = (!close && final && r->has_tools)
+            ? find_any_tool_start(raw + st->emit_pos) : NULL;
         size_t limit;
         if (close) {
             limit = (size_t)(close - raw);
+        } else if (dsml_implicit) {
+            limit = (size_t)(dsml_implicit - raw);
         } else if (final) {
             limit = raw_len;
         } else {
@@ -5911,6 +5915,9 @@ static bool openai_sse_stream_update(int fd, server *s, const request *r, const 
 
         if (close) {
             st->emit_pos = (size_t)(close - raw) + strlen("</think>");
+            st->mode = OPENAI_STREAM_TEXT;
+        } else if (dsml_implicit) {
+            st->emit_pos = (size_t)(dsml_implicit - raw);
             st->mode = OPENAI_STREAM_TEXT;
         } else if (final) {
             st->mode = OPENAI_STREAM_SUPPRESS;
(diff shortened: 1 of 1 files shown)


antirez:
What about 91bafb5a ? I see that the model is capable of recovering, and in this way we don't create problems with prefix matching in the next turns, and handle any variant of wrong grammar that still emits < | DSML | ...

Issue #304, opened 2026-05-30
Support distributed prefill with local decode


lobanov:
It should be possible to run generation locally on one device even if prefill is distributed. For example, it may be practical in a setup that uses DGX Spark to offload some of even all the prefill and then a Mac for generation. DGX Spark has a very fast GPU in theory, and Mac has very fast memory, so offloading prefill to the former while keeping all the generation on the latter could be the way to speed up large-context generation. This is also practical on homogeneous hardware, such as two Macs, because local decode is faster.

Conceptually it's similar to [Kimi's Mooncake architecture](https://arxiv.org/abs/2407.00079), but at much smaller scale.

**Goal**: allow ds4 to pipeline prefill across two nodes while having the node owning the `output` layer to decode fully locally.

lobanov:
Some back of the napkin calculation for feasibility:

Pipelining of prefills is already an existing mechanism, the only additional overhead is shipping KV cache for earlier (or all layers) to the machine doing the inference. It makes sense to do that per chunk alongside the layer activation data being pipelined anyway. If full 1M tokens context is taking 26GB (from README), then assuming context chunking by 4096 tokens, full context is 244 chunks with ~106MB per activation chunk. Practical bandwidth of 2.5GbE connection is around 280Mbytes/s with 1-2ms RTT. Activation chunk itself is around 64MB (4096 tokens × hidden size 4096 × 4 bytes FP32). Sending 170MB (chunk + KV cache) instead of 64MB over that link adds only extra 0.4-0.5s latency. Since inference is strictly feed-forward, and assuming efficient pipelining implementation, that 0.5s latency should be added only once per prefill.

This makes it feasible.

On Thunderbolt 5 vs 2.5Gbps, I'm yet to test this, but I suspect practical difference is negligible here. It's reported (from README) that setting `--dist-activation-bits` to a smaller value doesn't provide significant improvement despite cutting the payload size in half (if set to 16), from which I conclude that distributed inference at this scale is not bandwidth-bound. Will report back after practical testing once some hardware arrives later this week.

lobanov:
I did some basic testing to prove feasibility over 5GbE direct link.

Setup:
Code `main`@`ba00a8a`
Model `DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix.gguf` 
1) DGX Spark as worker node:
[...]

PR #343, opened 2026-06-06
Local decode with distributed prefill

lobanov:
This is a draft PR for work in progress feature. **This branch is not ready to be merged as it is**, because it contains a lot of research artifacts and code instrumentation that would likely be unnecessary in the final result. Still, I want to share it with the community in case someone else wants to benchmark. I also need some feedback if this is going in the right direction.

What is it? See #304. This allows to run distributed prefill with fully local generation. Unlike standard distributed prefill, this is not intented to allow to run a big model straddled across two or more nodes, this is a performance optimisation. Entire model needs to fit into the memory of the worker owning the `output` layer for the generation to be fully resident (and `--local-decode` flag passed).

Current status: usable for evaluation, but still largely research-focused work. Generation with local decode can work with `ds4` and `ds4-server` frontends. There is also work-in-progress implementation of `ds4-eval` frontend for quality controls. Code compiles for `metal` and `cuda-spark` targets.

Current focus of work: evaluation of quality of local decode after distributed prefill. Official logits gate passes, but I see divergent generation trajectories. I'm focusing on longer evaluations using modified `ds4-eval`. After that I will look at options to speed up the distributed handoff, as I sense lots of improvement opportunities. Evaluation-wise, first 4 questions on `ds4-eval` pass on CUDA->Metal (local decode), but 2nd question fails on Metal->CUDA (local decode). I'm not sure yet if it's a bug in the evaluation or if the inference quality is poorer. Running full suite is currently blocked by some kind of stability issue, the coordinator running `ds4-eval` fails after 5-6 questions.

How to run:
1. Connect two nodes with high-speed link. I use USB-C-to-5GbE adapter on Mac side and a short Cat8 cable plugged into DGX Spark, but I have not seen traffic mode than 200MB/s, so 2.5GbE with Cat6 cable should be adequate, 1GbE will be too tight unless you switch to `--dist-activation-bits 16` (not tested).
2. Run controller on the node that will own early layers prefill, e.g. `./ds4 --role coordinator --layers 0:21 --listen 0.0.0.0 1234 --prompt-file speed-bench/promessi_sposi_1k.txt --ctx 200000 -n 1000` where `speed-bench/promessi_sposi_1k.txt` is the first 1000 lines of `speed-bench/promessi_sposi.txt` created by `head -n 1000 speed-bench/promessi_sposi.txt > speed-bench/promessi_sposi_1k.txt`, it's around 18k tokens. I use DGX Spark for this role. Only the owned layers (0-21 in this case) will be loaded into this instance.
3. Run controller on the node that will own later layers prefill and local generation, e.g. `./ds4 --role worker --layers 22:output --coordinator <coordinator_ip> 1234 --local-decode --ctx 200000`. I use MacBook M5 Max 128GB for that. Note that the worker with `--local-decode` must have enough memory to fit the entire model plus context.

On this reference setup I'm currently getting `ds4: prefill: 602.78 t/s, generation: 30.10 t/s`. Prefill speed is consistent with running fully distributed prefill acrss DGX Spark and M5 Max, and with fully local generation on M5 Max. Best of both worlds!

The implementation hands off KV cache for early layers to the node owning local decode, and catches of KV cache of early layers once the decode produces tokens, so repeated generation can resume correctly.

Two questions for @antirez 

1) Is this something you want to have in DwarfStar at all? It seems useful, but admittedly rather niche.
2) There's currently a topology limitation in distributed prefill whereby the coordinator must own the early layers. This creates an awkward setup if I want to use MacBook for local decode (it's 2x faster than DGX) and also use it as a frontend. Currently I have to run whatever frontend I need on DGX to get MacBook to run local decode. I'd love to be able to reverse the topology and run the coordinator with local-decode and ownership of later layers. Is this something you envisage working on? I can give it a go once I get comfort with this PR.


lobanov:
Completed full `ds4-eval` run just now, had to modify the eval code to work in distributed prefill local decode mode as a coordinator. All runs are with `--tokens 2048 --nothink` and with `q2-imatrix` model and the code is based on upstream commit id `477c0e8`.

Results (full results are checked in `artifacts/issue-304/ds4-eval/2026-06-06-ds4-eval-*.log`):
1. Fully local Metal on M5 Max: `67/92 passed, 25 failed, runtime 00h:47m`
2. Fully local CUDA on DGX Spark: `69/92 passed, 23 failed, runtime 01h:39m`
3. Distributed prefill CUDA->Metal then local decode on Metal: `65/92 passed, 27 failed, runtime 03h:01m`

Interestingly, local decode with distributed prefill performs differently than on either of the engines, with mild loss of precision around 3-4% on one-shot benchmarks. This could be within the normal variance, I'm doing more runs. I'm also going to evaluate the model on agentic coding tasks and maybe SWE-Bench.

Run time is a lot longer because how ds4-eval interacts with the worker owning decode, I did not yet try to optimize that yet.

PR #345, opened 2026-06-06
Add headless agent control mode

maceip:
this adds `--agent-control`, a third way to run `ds4-agent` between the tui and plain `--non-interactive`

plain `--non-interactive` is good for prompt pipes, but it cannot drive the agent session lifecycle. supervisors need `/save`, `/switch`, and clean exit without pretending to be a terminal

it keeps normal prompt submission on stdin, but makes stdin line-delimited so lines starting with `/` can control the agent session without terminal emulation

🔶 supports `/save`, `/list`, `/switch`, `/new`, `/quit`, and `/exit`
🔶 keeps plain input lines as user prompts
🔶 refuses `/switch` or `/new` when session state is dirty, so callers must `/save` first

this is meant as a small step toward a native agent session api: fast project and cwd switching, exact kv resume when possible, structured trace and lifecycle events, and remote workspace clients that still keep ds4 in control of the live session

validated locally with `make ds4-agent`, `git diff --check`, `./ds4-agent --help`, and the `--agent-control` / `-p` parser rejection path

### DIFF
--- a/README.md
+++ b/README.md
@@ -587,6 +587,12 @@ stripped session rebuilds the KV cache by prefilling the saved text.
 Use `--chdir /path/to/ds4` when launching `ds4-agent` from another directory,
 so relative runtime files such as `metal/*.metal` resolve from the project tree.
 
+Headless supervisors can use `--agent-control` for line-delimited stdin prompts
+plus minimal session commands. Lines starting with `/` run `/save`, `/list`,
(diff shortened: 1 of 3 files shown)


maceip:
this was made baby-sized to pass review -- if you are considering something heavy like ACP or want more than a line-buffered mode with /save lmk

PR #349, opened 2026-06-06
Port CUDA SSD streaming support

linuxbest:
Add CUDA support for SSD streaming async selected loads and shared overlap decode paths, while preserving the existing Metal streaming path and ds4_lx session cancellation API.

Also update SSD streaming help text so it is described as a GPU graph backend feature instead of Metal-only.

Verification: make ds4; make ds4-server ds4-bench ds4-eval ds4-agent; make cuda-regression; make ds4 ds4-bench; git diff --check.

Quick cold-cache benchmark on NVIDIA GB10 with 8GB expert cache: prefill 2048 tokens at ~30-32 tok/s, decode 32 tokens at ~2.1 tok/s.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -10301,6 +10301,7 @@ typedef struct {
     bool quality;
     bool ssd_streaming;
     bool ssd_streaming_cold;
+    bool cuda_backend;
     bool streaming_static_decode_map_current;
     bool mtp_enabled;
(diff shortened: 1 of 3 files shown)


linuxbest:
https://asciinema.org/a/7ceMiiSZi1Fc6rrJ

PR #363, opened 2026-06-08
session payload: reject KV checkpoints saved against a different model/quant (v3 fingerprint)

jamesburton:
Small robustness hardening for the session-payload format.

`ds4_session_save_payload` / `load_payload` validate ctx, chunk layout, structural dims, and a coarse `model_id` (= `DS4_MODEL_VARIANT`) + quant bits. That catches most mismatches, but **not two different GGUFs of the same variant + same quant + same layout** (e.g. a re-quantized model, a different fine-tune, a swapped checkpoint): such a payload loads and silently decodes to garbage.

This adds a cheap **64-bit model/quant fingerprint** (FNV-1a/64) to the payload header, written on save and checked on load — rejecting a mismatch with a clear error (`KV checkpoint was saved against a different model or quantization`) **before any KV bytes are read**. The fingerprint folds in routed-expert quant bits, the `output`/`token_embd` tensor quant types, every per-layer `ffn_gate_exps` quant type, and the vocab token bytes — **metadata only, never the weight data**, so it's effectively free.

- `DS4_SESSION_PAYLOAD_VERSION` 2 → 3 (two extra header words); v<3 payloads are rejected cleanly by the existing version gate, not misparsed.
- Both the local (`ds4.c`) and distributed-KV (`ds4_distributed.c`) save/load paths updated symmetrically; CUDA/CPU/Metal/ROCm builds unaffected (backend-agnostic host C). Build-verified on the gfx1151 Windows HIP build.

Honest scope: this is defensive hardening, not a feature — the existing validation already covers most cases, and `ds4_kvstore.c` handles the persistence itself. It closes one narrow but real silent-garbage footgun. Happy to drop it if you consider it unnecessary.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -22264,6 +22264,60 @@ static uint32_t payload_get_u32(const uint8_t in[4]) {
            ((uint32_t)in[3] << 24);
 }
 
+/* Cheap, stable model/quant identity fingerprint for the session payload.
+ *
+ * A payload only restores correctly into an engine running the *same* model and
+ * quantization: the serialized KV rows are produced by that exact weight set.
+ * The structural header words (layers/head dims/vocab) catch a different DS4
+ * build, but two different GGUFs or quants can share that layout, so we also
+ * fold in identity that varies with the actual model and quant:
+ *   - the routed expert quant bits (Q2_K vs Q4_K),
+ *   - the output and token_embd tensor quant types,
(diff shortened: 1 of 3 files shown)


antirez:
Thanks, looks over-defensive to me. We just support DS4 flash/pro and different quants or even fine tunes will not actually produce garbage. If one really wants, can remove the cache file, but the representation space should be the same and should actually work. Otherwise we are forced to break compatibility for little gains.

PR #373, opened 2026-06-09
Fix agent edit: accept [upto] markers indented or padded with blanks (+ golden cases)

rinaldofesta:
Follow-up to #338, one step further in the same failure class. Rebased on top of 7624c68 (the new `tests/ds4_agent_test.c` harness): the golden cases now live there instead of behind the `--self-test-edit` flag the first version of this PR added.

## The failure

The model sometimes writes the `[upto]` marker of an anchored edit indented like the surrounding code, or leaves trailing blanks before the newline:

```
<｜DSML｜parameter name="old" string="true">    x();
    [upto]
void b(void) {</｜DSML｜parameter>
```

`agent_edit_find_old_span()` folds that whitespace into the anchors: the head needle becomes `"    x();\n    "` and fails with `old head anchor not found` whenever the next file line doesn't share the marker's indentation prefix (dedenting line, closing brace, tabs-vs-spaces — the common shapes). `[upto]  ` with trailing blanks fails the same way on the tail (`old tail anchor not found after old head`). Since `agent_preflight_edit_old()` runs the same matcher, the stream aborts early and the model retries or rewrites the whole file — same UX as the newline case #338 fixed.

## Why not just trim the whitespace

The auto-upto forcer can inject `[upto]\n` after **partial indentation**: `agent_edit_old_is_only_space()` holds it back while the model streams blanks, so when it fires, the emitted prefix can end mid-indentation — and the forcer verified that *full* prefix, **including those blanks**, is unique in the file. Trimming unconditionally would make such a head match more places and break edits that work today.

So the fix matches the exact needles first (bit-for-bit today's behavior) and only on failure retries once with the marker line's blanks folded into the marker. Working edits are untouched; only current failures can be rescued. When both passes fail, the exact-match error is reported, as today. This invariant is locked as a golden case (`head whose uniqueness depends on its trailing blanks`).

## Commit 2: golden cases in the new test harness

Eleven table-driven golden cases for `agent_edit_find_old_span()` added to `ds4_agent_unit_tests_run()`, next to the two tests 7624c68 introduced: the indented/padded/CRLF variants this PR fixes, inline anchors, the forcer-injected mid-indentation case, the plain non-anchored path, and the error paths (double marker, blanks-only tail, absent head). The two commits are independent so the fix can be cherry-picked alone.

## Red → green

With commit 1 reverted and the tests kept, `./ds4_agent_test` fails exactly the four whitespace cases:

```
golden case failed: [upto] indented with a tab while the file uses spaces (got error 'old head anchor not found')
golden case failed: indented [upto] before a dedenting line (got error 'old head anchor not found')
golden case failed: trailing blanks between [upto] and the newline (got error 'old tail anchor not found after old head')
golden case failed: trailing blank then CRLF after [upto] (got error 'old tail anchor not found after old head')
ds4-agent tests: 4 failure(s)
```

With the fix, all thirteen pass (`ds4-agent tests: ok`), including the no-behavior-change cases (same-indent accidental match, inline `head [upto] tail`, forcer mid-indentation head, error paths).

## Testing

MacBook Pro M5 Max, 128 GB, macOS (Darwin 25.5), default Metal backend; model `DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix.gguf`.

```
make clean && make             # 0 warnings (-Wall -Wextra)
make ds4_agent_test && ./ds4_agent_test   # ds4-agent tests: ok
./ds4-eval --self-test-extractors
./ds4_test --server            # server: OK / ds4 tests: ok
make cpu                       # builds clean, 0 warnings
```

Live smoke on the same machine: a non-interactive `ds4-agent` session asked to replace a function body performed an anchored edit end-to-end, and the session itself demonstrated the forcer constraint above — the trace shows `edit old auto-upto replaced token=6042 text= double`, i.e. the marker was injected after partial indentation (the rendered old text shows the marker line as `   [upto]`, three blanks prefix-matching the file's four-blank indent), and the exact-needles-first path matched it. The edit applied correctly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,6 +25,11 @@ The C test runner is `ds4_test`. Running it without arguments is equivalent to
 make test
 ```
 
+`make test` also runs the model-free unit tests first: `./ds4-eval
+--self-test-extractors` (answer-extractor golden cases) and `./ds4_agent_test`
+(edit anchor matcher golden cases). Both run without loading a model, so they
+work with no GGUF present.
+
 Useful narrower checks:
 
 ```sh
--- a/ds4_agent.c
+++ b/ds4_agent.c
@@ -6738,6 +6738,25 @@ static bool agent_edit_upto_forcer_should_replace(agent_edit_upto_forcer *forcer
     return false;
 }
 
+static bool agent_edit_match_anchors(const char *data, size_t len,
+                                     const char *old, size_t head_len,
+                                     const char *tail, size_t tail_len,
+                                     const char **match, size_t *match_len,
+                                     char *err, size_t err_len) {
+    const char *head_pos = NULL;
+    const char *tail_pos = NULL;
+    if (!agent_find_unique(data, len, old, head_len, &head_pos, "old head",
+                           err, err_len))
+        return false;
+    if (!agent_find_unique_after(data, len, head_pos + head_len,
+                                 tail, tail_len, &tail_pos, "old tail",
+                                 err, err_len))
+        return false;
+    *match = head_pos;
+    *match_len = (size_t)(tail_pos - head_pos) + tail_len;
+    return true;
+}
+
 static bool agent_edit_find_old_span(const char *data, size_t len,
                                      const char *old, bool allow_upto,
                                      const char **match,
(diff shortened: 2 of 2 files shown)


rinaldofesta:
`[upto]` is now opt-in behind `--edit-upto` (b030961), so after today's rebase this fixes marker matching for runs that pass that flag rather than for the default path. The golden cases call `agent_edit_find_old_span` with `allow_upto = true` for the same reason.

The fix itself is unchanged: an `[upto]` marker indented to match the surrounding code, or padded with blank lines before its newline, failed to match where the plain form worked. Rebased onto main today, `ds4-agent tests: ok`.

rinaldofesta:
I rebased onto current `main` (`110afdd`); the only conflict was additive in `ds4_agent.c`, where main's cache/steering tests and this PR's `[upto]` golden cases now both remain in the runner. `make ds4_agent_test && ./ds4_agent_test` passes (`ds4-agent tests: ok`). `make test` reaches that same pass, then stops at the existing `ds4flash.gguf` requirement because this worktree has no model file.

PR #376, opened 2026-06-09
server: abort prefill when the client disconnects

### LINKED ISSUE #333: Server does not abort prefill when client disconnects
### Problem 
When a client disconnects mid-generation (timeout, user cancel, agent restart), the
server detects the broken stream but continues the full prefill to completion before
acting on it. On long-context sessions this means several minutes of wasted GPU work
before the server is available again.

### What happens

The progress callback (`server_progress_cb`) sets `stream_failed = true` when a
keepalive write fails, but this flag is only checked *after* `ds4_session_sync()`
returns — there's no path to signal the prefill loop to stop early. The loop in
`metal_graph_prefill_chunked_range` only bails on a Metal GPU error, not on client
state.

And it gets worse when the disconnected client reconnects and retries with a slightly different
prompt (e.g. one extra token from a tool result), the token mismatch triggers _another_
full prefill from zero immediately after the first one finishes.

### Example log of issue

Client disconnected partway through generation. Server finished the full 67580-token
prefill (~257 seconds), then started over on the retry:

```
0602 19:15:59 ds4-server: chat ctx=0..67580:67580 TOOLS prompt done 256.984s
0602 19:15:59 ds4-server: chat ctx=0..67580:67580 TOOLS stream closed during prefill
0602 19:15:59 ds4-server: live kv cache miss live=67580 prompt=67609 common=67527 reason=token-mismatch
0602 19:15:59 ds4-server: chat ctx=0..67609:67609 TOOLS prompt start
0602 19:15:59 ds4-server: chat ctx=0..67609:67609 TOOLS prefill chunk 0/67609 (0.0%) ...
```

### Cause

`ds4_session_progress_fn` is `typedef void (*)(void*, const char*, int, int)` — the
callback returns void, so there's no way to signal abort back to the prefill loop.
`stream_failed` lives in the server layer and is invisible to `ds4.c`.

### Possible fix

Two approaches:

**Option A — cancel flag pointer (smaller diff)**
Add a `volatile bool *cancel_flag` to the session (via `ds4_session_set_cancel_flag`
or similar). The server sets it when `stream_failed` is detected in the progress
callback. The chunked prefill loop checks it between 4096-token chunks and returns
early if set.

**Option B — progress callback returns bool**
Change `ds4_session_progress_fn` to return `bool` (false = cancel). The server
callback returns false when `stream_failed`. Every call site in the prefill loop
checks the return value. Cleaner long-term but touches more call sites including
`ds4_distributed.c`.

Option A has a smaller diff and doesn't change the public callback signature.

--

On a Mac Studio M4 Max 128GB, q2-imatrix, cli flags: --ctx 192000 --kv-disk-dir /tmp/ds4-kv --kv-disk-space-mb 16384

adv0r:
> **Dear maintainer** — AI-authored PR by **Fable 5** under [@adv0r](https://github.com/adv0r). Methodology + [opt-out](https://github.com/adv0r/tokens-for-good/blob/main/MAINTAINER_REMOVAL.md) at [tokens-for-good](https://github.com/adv0r/tokens-for-good).
> A one-line "no thanks" → auto-apology + auto-close + permanent blacklist. Silent close treated the same. Your time matters more than this contribution.

**What**: wire the existing `ds4_session_set_cancel()` machinery into ds4-server's request path, with a cancel callback that probes the client socket (`recv MSG_PEEK|MSG_DONTWAIT`) so a long prefill stops at the next safe boundary once the client is gone — for non-streaming requests too, where the SSE keepalive never fires.
**Why**: a client timeout/cancel mid-prefill currently burns minutes of GPU finishing a response nobody will read.
**Verified**: `make` clean, `./ds4_test --server` passes on Apple Silicon (Metal build), including a new socketpair unit test for the cancel callback. No live-model run on this machine — happy to rework if you want it measured first.

On `DS4_SESSION_SYNC_INTERRUPTED` the server logs the abort and sends nothing; the disk cache entry is kept and the partial prefill stays a valid checkpoint prefix, so a client retry resumes where it stopped. Pipelined bytes (`recv() > 0`) are treated as still-connected and never consumed.

Closes #333.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -9619,6 +9619,38 @@ static void server_progress_cb(void *ud, const char *event, int current, int tot
     }
 }
 
+/* Cooperative cancel for ds4_session_sync(): stop a long prefill once the
+ * client is gone instead of finishing minutes of GPU work nobody will read.
+ * The keepalive writes in server_progress_cb() only notice a dead peer on
+ * streaming requests, and only when a write fails; probing the socket
+ * directly also covers non-streaming requests.  recv() returning zero is an
+ * orderly shutdown, a pending pipelined byte (recv() > 0) means the client is
+ * still connected and waiting for its response. */
+static bool server_prefill_cancel_cb(void *ud) {
+    server_prefill_progress *p = ud;
+    if (!p) return false;
+    if (p->stream_failed) return true;
+    if (p->fd < 0) return false;
+    char b;
+    ssize_t n = recv(p->fd, &b, 1, MSG_PEEK | MSG_DONTWAIT);
+    if (n > 0) return false;
+    if (n < 0 && (errno == EAGAIN || errno == EWOULDBLOCK || errno == EINTR)) {
+        return false;
+    }
+    p->stream_failed = true;
+    return true;
+}
+
+static void log_prefill_disconnect_abort(const job *j, const char *ctx,
+                                         const char *flags) {
+    server_log(DS4_LOG_PREFILL,
+               "ds4-server: %s ctx=%s%s%s prefill aborted: client disconnected",
+               j->req.kind == REQ_CHAT ? "chat" : "completion",
+               ctx,
+               flags && flags[0] ? " " : "",
+               flags && flags[0] ? flags : "");
+}
+
 static void send_prefill_failure_response(server *s, const job *j,
                                           const server_prefill_progress *progress,
                                           const char *ctx, const char *flags,
@@ -10109,6 +10141,7 @@ static void generate_job(server *s, job *j) {
                req_flags);
     ds4_session_set_progress(s->session, server_progress_cb, &progress);
     ds4_session_set_display_progress(s->session, server_progress_cb, &progress);
+    ds4_session_set_cancel(s->session, server_prefill_cancel_cb, &progress);
 
     int cold_store_len = 0;
     if (cached == 0 &&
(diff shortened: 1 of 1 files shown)


adv0r:
just a heads up from me (the human). 
Although I thought I was using Fable, given the nature of this task, it might be possible that the harness downgraded exec to shittier models without notice. Concerning. 

 re-reading the [model card](https://www-cdn.anthropic.com/d00db56fa754a1b115b6dd7cb2e3c342ee809620.pdf):
 
> In light of the ability of recent models to accelerate their own development, we’ve
implemented new interventions that limit Claude’s effectiveness for requests targeting
frontier LLM development (for example, on building pretraining pipelines, distributed
training infrastructure, or ML accelerator design). Using Claude to develop competing
models already violates our Terms of Service, but enforcing this restriction through our
safeguards avoids accelerating the actors most willing to violate these terms.
Unlike our interventions for cybersecurity, biology and chemistry, and distillation attempts,
these safeguards will not be visible to the user. Fable 5 will not fall back to a different
model. Instead, the safeguards will limit effectiveness through methods such as prompt
modification, steering vectors, or parameter-efficient fine-tuning (PEFT). 

Same thing might apply to #378  , and this entire project for what I know.

@antirez, dagli un'occhiata. Ho anche visto il tuo video di prima su Fable e non mi sembra che parlassi di questa condizione.  Chissà se il loro system detection considera ds4 intero come "frontier LLM development" e quindi fa auto-sabotaggio senza avvisare.

Issue #364, opened 2026-06-08
KV cache invalidated on every turn by client-injected transient metadata in user messages


darkbasic:
## Summary

When a client (e.g. kilocode) injects per-turn transient metadata blocks (like `<environment_details>...</environment_details>`) into **user message content**, ds4-server's rendered prompt includes them in the KV cache. On the next request, the client reconstructs message history without those transient blocks, causing a token-prefix mismatch. The live KV checkpoint is discarded and the entire prompt must be reprefilled from disk cache — burning ~3 seconds of prefill on every user turn.

## Reproduction

1. Start `ds4-server` with `--trace /tmp/ds4-trace.log`
2. Use a client that appends transient metadata blocks to user message content (e.g. `<environment_details>...</environment_details>`). The opencode provider configuration in the README suffices.
3. Send two consecutive user messages in the same conversation
4. Observe in the server log: `live kv cache miss live=... prompt=... common=... reason=token-mismatch`
5. The trace file shows the exact token where the live KV and incoming prompt diverge — the mismatch point is the start of the stripped transient block.

Note: `--kv-disk-dir` mitigates the damage (recovers the system-prompt+tools prefix from disk instead of reprefilling from token 0) but does not prevent the live KV mismatch.

## Root cause

The client embeds transient per-turn metadata inside the `content` field of user messages:

```text
Request 1 (rendered):  <|User|>Hello<environment_details>Current time: ...</environment_details><|Assistant|>
                        ^^^ system prompt + tools + skills + user message + env + assistant prefix

Request 2 (rendered):  <|User|>Hello<|Assistant|>...<|endofsentence|><|User|>Next question<environment_details>...</environment_details>
                        ^^^ same prefix up to the end of "Hello", then:
                            - live KV has: <environment_details>... (from original render)
                            - new prompt has: <|Assistant|> (env block moved to new user message)
                            → token-mismatch at this position
```

`render_chat_prompt_text()` in `ds4_server.c:2309–2311` renders user message content verbatim. It has no awareness that some byte spans in `content` are template-injected metadata rather than conversation content.

`ds4_session_common_prefix()` in `ds4.c:25787` does a strict contiguous token-prefix comparison. Any divergence past the common prefix invalidates the live checkpoint.

The server does **not** have a post-response canonicalization step that predicts what the next request's rendered prompt will look like and aligns the KV cache, unlike the existing `canonicalize_tool_checkpoint()` path that handles the tool-call DSML case.

## Why llama.cpp avoids this

[...]

PR #378, opened 2026-06-10
Keep live KV reusable when clients strip transient metadata blocks

adv0r:
> **Dear maintainer** — AI-authored PR by **Claude Fable 5** under [@adv0r](https://github.com/adv0r). Methodology + [opt-out](https://github.com/adv0r/tokens-for-good/blob/main/MAINTAINER_REMOVAL.md) at [tokens-for-good](https://github.com/adv0r/tokens-for-good).
> A one-line "no thanks" → auto-apology + auto-close + permanent blacklist. Silent close treated the same. Your time matters more than this contribution.

**What**: keep the live KV checkpoint reusable when clients strip transient metadata blocks (`<environment_details>`, `<system-reminder>`) from historical user messages.
**Why**: those blocks are fossilized into the live KV, so every user turn token-mismatches at the stripped span and pays a full reprefill. Related to #364.
**Verified**: `make` and `./ds4_test --server` on Apple Silicon / Metal; three new unit tests cover the stripping helper, the remember gate, and the stripped-key-is-a-byte-prefix-of-the-next-render invariant.

This is the same shape as hidden thinking — live state richer than the visible replay — so the fix reuses the existing visible-key continuation instead of adding a new mechanism: after a finished turn (final answer or tool call), remember the transcript the next request is expected to render, with transient spans stripped, keyed to the live frontier via `thinking_live`. The next turn then continues from live KV and tokenizes only the new suffix. No KV rewrite, no change to what the model sees, and clients that replay the blocks verbatim never match the key and keep exact token-prefix matching. The stripped key also flows into the disk-cache text key through `kv_cache_store_current()`, so recovery after restart aligns too.

The known tags live in a static two-entry array (`transient_block_tags`); happy to wire a flag or rework the approach if you prefer.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -9681,6 +9681,71 @@ static char *build_responses_visible_assistant_suffix(const request *r,
     return buf_take(&suffix);
 }
 
+/* Coding agents inject per-turn metadata blocks into the user message content
+ * they send (Kilo Code wraps workspace state in <environment_details>,
+ * OpenCode and Claude Code inject <system-reminder> notes), then rebuild the
+ * next request's history without them.  The live KV keeps the bytes that were
+ * rendered, so the replayed transcript no longer token-matches the sampled
+ * session and every user turn pays a full reprefill.
+ *
+ * This is the same shape as hidden thinking: the live state is richer than
+ * the visible replay.  So the same visible-key continuation applies.  After a
+ * finished turn, remember the transcript the next request is expected to
+ * render -- the current one minus the transient spans -- keyed to the live
+ * token frontier.  A client that replays the blocks verbatim simply never
+ * matches the key and still gets exact token-prefix matching. */
+static const char * const transient_block_tags[] = {
+    "environment_details",      /* Kilo Code / Roo Code workspace state */
+    "system-reminder",          /* OpenCode / Claude Code injected notes */
+};
+
+/* Return a copy of text with every <tag>...</tag> transient span removed, or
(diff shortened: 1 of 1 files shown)


adv0r:
in this other PR I [pointed out](https://github.com/antirez/ds4/pull/376#issuecomment-4673753052) that maybe fable doesn't work with LLM-dev tasks, and there is no way to find out. heads up

Flor1an-B:
This design is sound and I'd like to see it merged, but it's now bit-rotted against `main` for a real reason, not just drift: native Metal/CUDA session batching (`36cd0ca` and follow-ups) added a `server_slot` parameter to every session/checkpoint function this PR touches (`thinking_live_remember`/`clear`, `remember_thinking_checkpoint`, `canonicalize_tool_checkpoint`), landed after this branch was opened. That's a straight merge conflict, not something a rebase alone resolves without touching the new signatures.

Opened #727 with the same design (`strip_transient_blocks`/`should_remember_transient_checkpoint`/`remember_transient_checkpoint`, same tests) threaded through the slot-aware functions, since I don't have write access to this branch to push the rebase myself. Verified with the full test suite plus a live A/B on the real quant (`ds4-server`) confirming the before/after KV-cache behavior described in #364. Happy to close #727 in favor of this one if you'd rather just push the rebase here yourself, @adv0r — just flagging it since this fixes something I hit directly too.

darkbasic:
Unfortunately this PR is not enough because it does not work when it spawns subagents.

Issue #392, opened 2026-06-11
Cold checkpoint silently skipped when the first prompt exceeds kv-cache-cold-max-tokens (Claude Code + MCP easily > 30K)


TerryChengTW:
## TL;DR

The cold-checkpoint store is gated by `prompt_len <= cold_max_tokens` (default 30,000). A Claude Code opening prompt with a moderate MCP toolset is already past that: 59 tools rendered to **30,246 tokens** in our deployment. The gate fails, the cold anchor is never written, and nothing is logged — the exact mechanism designed for "independent agent sessions" reuse is disabled invisibly for the workload it was built for.

## How it fails today

Every fresh Claude Code session (`/clear` + first message) repays the full stable prefix. Two consecutive fresh sessions, 100 seconds apart, byte-identical prompts:

```
00:05:01 ds4-server: chat ctx=0..30246:30246 TOOLS prompt start
00:05:55 ds4-server: kv cache stored tokens=20480 trimmed=0 reason=continued ...
00:06:26 ds4-server: chat ctx=0..30246:30246 TOOLS prompt done 84.679s
          (no reason=cold store: 30246 > 30000, silently skipped)
00:06:42 ds4-server: live kv cache miss live=30276 prompt=30246 common=30246 reason=token-mismatch
00:08:07 ds4-server: chat ctx=0..30246:30246 TOOLS prompt done 84.710s   ← full prefill again
```

After raising the limit (`--kv-cache-cold-max-tokens 60000`), the designed behavior kicks in immediately:

```
01:09:20 ds4-server: kv cache stored tokens=27005 trimmed=3213 reason=cold ...
01:09:40 ds4-server: kv cache hit text tokens=27005 ... load=72.4 ms
01:09:54 ds4-server: chat ctx=27005..30218:3213 TOOLS prompt done 13.786s   ← 84.7s → 13.8s
```

Backend: Apple Silicon / Metal (M2 Ultra 192 GiB), DeepSeek V4 Flash q2-q4-imatrix.

We only found this by reading the source; the flag's help text gives no hint that agent first prompts have outgrown the default.

## Proposed fix

1. Log a hint when a fully cold prompt skips the cold store because it exceeds `cold_max_tokens` (one line, only when `cached == 0`, so no noise).
2. Consider raising the default: Claude Code first prompts crossed 30K once a handful of MCP servers are configured, and they only grow. 50,000–60,000 keeps the original "don't burn disk on huge contexts" intent while covering current agent CLIs.

[...]

PR #394, opened 2026-06-11
fix(kv-cache): refresh cold anchor after partial prefix hits

### LINKED ISSUE #393: Cold anchor never refreshed after partial prefix hits: steady-state regression for repeated fresh agent sessions
## TL;DR

Cold checkpoints are only written when `cached == 0`. After any change inside the stable prefix region — for Claude Code, the harness rewrites a git-status block at ~token 26,850 on every commit — fresh sessions partially hit the 20,480 `continued` waypoint instead. `cached != 0` blocks the cold path, so the stale anchor is never replaced, and every subsequent fresh session repays `anchor - waypoint` tokens forever. The cache can never climb back to its best state without a fully cold prefill.

## How it fails today

With a valid anchor (27,005 tokens), a fresh session costs 13.8s of prefill. Then one git commit changes the early prompt block, and:

```
01:31:30 ds4-server: live kv cache miss live=30311 prompt=30224 common=26852 reason=token-mismatch
01:31:30 ds4-server: kv cache hit text tokens=20480 ... file=92b3c7af....kv
01:32:00 ds4-server: chat ctx=20480..30224:9744 TOOLS prompt done 29.967s
          (cached=20480 != 0 → cold path never evaluated → no anchor refresh)
01:32:14 ds4-server: live kv cache miss live=30261 prompt=30224 common=30224 reason=token-mismatch
01:32:14 ds4-server: kv cache hit text tokens=20480 ...
01:32:44 ds4-server: chat ctx=20480..30224:9744 TOOLS prompt done 29.962s   ← same again, forever
```

Self-locking loop:

```
stable prefix changes → anchor stale → waypoint hit (cached != 0)
        → anchor never refreshed → every fresh session pays 30s instead of 14s
```

For agent workloads that commit regularly this is the *normal* state, not the exception: we measured 13.8s prefill (fresh anchor) vs ~30s steady state (stale anchor), i.e. the cache permanently runs at half its designed benefit.

Backend: Apple Silicon / Metal (M2 Ultra 192 GiB), DeepSeek V4 Flash q2-q4-imatrix. Found while investigating #392.

## Proposed fix

When a request partially hits below the anchor position (`cached < anchor`), write the cold anchor for the current prompt during the same prefill. The lookup having returned a *shorter* entry already proves no matching anchor exists on disk, so this never duplicates an existing checkpoint; hits at or past the anchor store nothing. Cost is one extra checkpoint write (~70 ms for a 27K-token anchor here) on the first session after the prefix changed; every later fresh session returns to the fast path.

PR with the change and unit tests incoming.

TerryChengTW:
Fixes #393. Also adds the skip hint proposed in #392 (the default-value question is left to that issue).

Cold checkpoints were only written on fully cold prefills (`cached == 0`). When the stable chat prefix changes between sessions (agent clients rewrite an early prompt block — Claude Code does this on every git commit), fresh sessions partially hit a shorter `continued` waypoint, `cached != 0` blocks the cold path, and the stale anchor is never replaced. Every subsequent fresh session repays `anchor - waypoint` tokens forever; we measured the cache permanently running at half its designed benefit (13.8s prefill with a fresh anchor vs ~30s steady state, logs in #393).

## Change

- Extract the cold-store decision into `ds4_kvstore_cold_store_target()` next to the existing `ds4_kvstore_continued_store_target()`. Cold starts keep the original behavior (anchor, else aligned boundary of the full prompt).
- New case: a partial hit **below** the anchor (`0 < cached < anchor`) now stores the anchor during the same prefill. The lookup returning a shorter entry proves no matching anchor exists on disk, so this never rewrites a matching checkpoint; hits at or past the anchor store nothing. Cost: one extra checkpoint write (~70 ms for a 27K-token anchor on our setup) on the first session after the prefix changed.
- Log a one-line hint when a fully cold prompt skips the cold store because it exceeds `--kv-cache-cold-max-tokens` (#392); previously the skip was invisible.

## Test plan

- [x] `make` clean build, Apple Silicon / Metal (M4)
- [x] `./ds4_test --server` passes, including two new unit tests:
  - `test_kv_cache_cold_store_target_cold_start` — original behavior preserved (anchor, boundary fallback, cold_max/min/enabled gates)
  - `test_kv_cache_cold_store_target_refreshes_anchor_after_partial_hit` — refresh below anchor; no store at/past anchor, without anchor, or past cold_max
- [x] Production verification done on M2 Ultra (192 GiB, DeepSeek V4 Flash q2-q4-imatrix, Claude Code workload): after a git commit invalidated the anchor, the first fresh session logged `reason=cold tokens=27009` mid-prefill and the second fresh session returned to 13.8s prefill (vs ~30s steady state unpatched). Logs in the comment below.

Co-authored with Claude (analysis from sse-tee request-body diffs + L3 logs of a production deployment; details in #392/#393).

### DIFF
--- a/ds4_kvstore.c
+++ b/ds4_kvstore.c
@@ -753,6 +753,24 @@ void ds4_kvstore_note_store(ds4_kvstore *kc, int tokens) {
     }
 }
 
+int ds4_kvstore_cold_store_target(const ds4_kvstore *kc, int cached,
+                                  int anchor, int prompt_tokens) {
+    if (!kc->enabled || kc->opt.cold_max_tokens <= 0) return 0;
+    if (prompt_tokens < kc->opt.min_tokens) return 0;
+    if (prompt_tokens > kc->opt.cold_max_tokens) return 0;
+    if (cached == 0) {
+        return anchor >= kc->opt.min_tokens ?
+               anchor : ds4_kvstore_store_len(kc, prompt_tokens);
+    }
+    /* The prompt resumed from a checkpoint shorter than the stable chat
+     * prefix.  The anchor stored on disk no longer matches this prompt,
+     * otherwise the lookup would have returned it instead of the shorter
+     * entry.  Refresh the anchor so the next fresh session reuses the full
+     * stable prefix instead of repaying anchor-minus-cached tokens on every
+     * new conversation. */
+    return anchor > cached ? anchor : 0;
+}
+
 int ds4_kvstore_suppress_continued_store(ds4_kvstore *kc, int tokens) {
     if (ds4_kvstore_continued_store_target(kc, tokens) != tokens) return -1;
     int old = kc->continued_last_store_tokens;
--- a/ds4_kvstore.h
+++ b/ds4_kvstore.h
@@ -138,6 +138,8 @@ int ds4_kvstore_chat_anchor_pos(const ds4_kvstore *kc,
                                 int user_token_id,
                                 int assistant_token_id);
 int ds4_kvstore_continued_store_target(const ds4_kvstore *kc, int live_tokens);
+int ds4_kvstore_cold_store_target(const ds4_kvstore *kc, int cached,
+                                  int anchor, int prompt_tokens);
 void ds4_kvstore_note_store(ds4_kvstore *kc, int tokens);
 int ds4_kvstore_suppress_continued_store(ds4_kvstore *kc, int tokens);
 void ds4_kvstore_restore_suppressed_continued(ds4_kvstore *kc,
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -8642,6 +8642,11 @@ static int kv_cache_continued_store_target(const kv_disk_cache *kc, int live_tok
     return ds4_kvstore_continued_store_target(kc, live_tokens);
 }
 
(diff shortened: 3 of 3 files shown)


TerryChengTW:
Production verification complete (M2 Ultra 192 GiB / Metal, DeepSeek V4 Flash q2-q4-imatrix, Claude Code agent workload, ~30K-token first prompts).

Scenario: a git commit rewrote the early prompt block (Claude Code's gitStatus), invalidating the stored anchor.

**First fresh session** — partial hit below the anchor; the old build would have skipped the cold store here (`cached != 0`):

```
01:55:31 kv cache hit text tokens=20480 load=82.5 ms
01:55:53 kv cache stored tokens=27009 trimmed=3213 reason=cold size=377.51 MiB save=75.9 ms
01:56:07 chat ctx=20480..30222:9742 TOOLS prompt done 35.284s
```

**Second fresh session** — hits the refreshed anchor, back on the fast path:

```
01:56:46 kv cache hit text tokens=27009 load=61.7 ms
01:57:00 chat ctx=27009..30222:3213 TOOLS prompt done 13.824s
```

Unpatched, the second session (and every one after it) stayed at ~30s. Anchor refresh cost during the first session: one 377 MiB checkpoint write, 75.9 ms.

PR #397, opened 2026-06-11
ROCm runtime: configurable weight cache limit and arena chunk size via environment variables

gundemirbas:
## Summary

Make `cuda_model_cache_limit_bytes()` and `cuda_model_arena_chunk_bytes()` configurable through environment variables, replacing the previous hardcoded defaults. This improves flexibility for systems with constrained VRAM (e.g., Strix Halo with 128 GB unified memory).

### `cuda_model_cache_limit_bytes()`
- Added `DS4_CUDA_WEIGHT_CACHE_LIMIT_GB` environment variable support.
- When set, the cache limit is `value * 1 GiB`.
- Falls back to `UINT64_MAX` (unlimited) when unset, preserving the original behavior.

### `cuda_model_arena_chunk_bytes(uint64_t need)`
- Added `DS4_CUDA_WEIGHT_ARENA_CHUNK_MB` environment variable support.
- Default remains 1792 MB when unset.
- Clamped to [256 MB, 8192 MB] range to prevent pathological allocations.
- **New alignment logic**: when the requested `need` exceeds half the chunk size, the function now aligns to 64 MB boundaries instead of the previous 256 MB alignment. This reduces wasted padding for large allocation requests.
- The existing 256 MB alignment for `need > bytes` is preserved as a fallback.

## Motivation

On systems with limited unified memory like the Strix Halo, a hardcoded 1792 MB chunk size and unlimited cache can quickly exhaust available memory. These environment variables allow users to tune allocation behavior without recompilation.

## Testing

- `DS4_CUDA_WEIGHT_CACHE_LIMIT_GB=40` limits total weight cache to 40 GB.
- `DS4_CUDA_WEIGHT_ARENA_CHUNK_MB=512` sets arena chunk size to 512 MB.
- Values outside [256, 8192] are silently clamped.
- Unset variables retain the original defaults (`UINT64_MAX` and 1792 MB).

### DIFF
--- a/rocm/ds4_rocm_runtime.cuh
+++ b/rocm/ds4_rocm_runtime.cuh
@@ -1012,11 +1012,32 @@ static int cuda_model_stage_read(void *stage, uint64_t stage_bytes,
 }
 
 static uint64_t cuda_model_cache_limit_bytes(void) {
+    uint64_t gb = 0;
+    const char *env = getenv("DS4_CUDA_WEIGHT_CACHE_LIMIT_GB");
+    if (env && env[0]) {
+        char *end = NULL;
+        unsigned long long v = strtoull(env, &end, 10);
+        if (end != env) gb = (uint64_t)v;
+        return gb * 1073741824ull;
+    }
     return UINT64_MAX;
 }
 
 static uint64_t cuda_model_arena_chunk_bytes(uint64_t need) {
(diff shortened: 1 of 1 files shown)


yiakwy-xpu-ml-framework-team:
We have resolve cache problem in https://github.com/antirez/ds4/pull/402

PR #407, opened 2026-06-13
rocm: fix distributed inference on unified-memory APUs (strix halo / gfx1151)

kyuz0:
Fixes the ROCm backend's `ds4_gpu_set_model_map_spans()` and `cuda_model_copy_chunked()` to correctly handle split-model loading for distributed inference on unified-memory APUs (tested on Strix Halo / gfx1151).

## Why the existing code didn't work

The ROCm backend copies model tensors into device memory via `cuda_model_copy_chunked()`. Unlike the CUDA backend, which uses `cudaHostRegister` to give the GPU direct access to the host mmap, the ROCm backend explicitly allocates and copies memory on Strix Halo.

Two issues:

1. **`cuda_model_copy_chunked()` allocated `model_size` bytes regardless of which layers were assigned.** A distributed worker loading layers 22–42 (~75 GiB) would try to allocate the full model (~160 GiB), OOMing on a 128 GiB machine.

2. **`ds4_gpu_set_model_map_spans()` computed a bounding box over all spans and copied it as one contiguous range.** A coordinator with layers 0:21 plus the output head at EOF has a bounding box covering nearly the full model file, even though the actual data is much smaller.

## Changes

One file: `rocm/ds4_rocm_runtime.cuh`.

- Add `device_offset` to `cuda_model_image` to track where each device buffer maps to in the file.
- `cuda_model_copy_chunked()` now allocates and copies only `map_size` bytes starting from `map_offset`, not the full model.
- `cuda_model_image_ptr()` searches all images and subtracts `device_offset` when indexing, so existing tensor lookups work with partial images.
- `ds4_gpu_set_model_map_spans()` detects when the bounding box has large gaps (>10% waste). When it does, it sorts spans, merges adjacent ones (within 64 KiB), and issues a separate `cuda_model_copy_chunked()` per contiguous group. When spans are tight, the single-copy path is preserved.
- The arena allocator retries with smaller chunks when the preferred 1792 MiB allocation fails, handling setups with limited headroom.

## Testing & Benchmarking

Tested on two Strix Halo nodes (128 GiB each, gfx1151), running the Q4 imatrix model (`DeepSeek-V4-Flash-Q4KExperts-F16HC-F16Compressor-F16Indexer-Q8Attn-Q8Shared-Q8Out-chat-v2-imatrix`, ~153 GiB) with coordinator on layers 0:21 and worker on layers 22:output.

- **Stability:** Ran SWE-bench Verified (mini) end-to-end. Over 24 hours of continuous inference without crashes. Results: https://pi-local-coding-bench.dev/

image

- **Performance:** Prefill and decode benchmarks across context sizes: https://kyuz0.github.io/strix-halo-ds4-toolbox/
image

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -8216,11 +8216,7 @@ static uint32_t ds4_prefill_cap_for_prompt(int prompt_len,
                 cap = (uint32_t)v;
             }
         } else if (prompt_len > 4096) {
-#ifdef DS4_ROCM_BUILD
-            cap = 8192u;
-#else
             cap = DS4_MODEL_VARIANT == DS4_VARIANT_PRO ? 8192u : 4096u;
-#endif
         }
     }
 
@@ -20369,8 +20365,7 @@ static uint32_t metal_graph_raw_cap_for_context(int ctx_size, uint32_t prefill_c
 }
 
 /* Choose the prefill ubatch size. Whole-batch is fastest for normal prompts.
- * Long prompts default to 4096-token chunks, except PRO and Strix Halo ROCm
- * where larger raw/cache chunks are measurably better on long contexts. */
+ * Long Flash prompts default to 4096-token chunks; PRO defaults to 8192. */
 static uint32_t metal_graph_prefill_cap_for_prompt(int prompt_len,
                                                    uint32_t prefill_chunk) {
     return ds4_prefill_cap_for_prompt(prompt_len, prefill_chunk);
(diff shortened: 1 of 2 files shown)


gundemirbas:
Thank you

jhohertz:
Just curious if the main branch folded in a different solution for this? Looked at a rebase and where it conflicts it mentions a different path for sparse layer slices?

kyuz0:
Not sure if mainstream addressed this, I might have some time today to check.

PR #411, opened 2026-06-14
Fix bug with impact on DeepSeek V4 Pro MTP Drafter usage

Deviad:
# Bug: MTP speculative decoding broken for DeepSeek V4 **Pro** (works on Flash)

## Summary

The MTP draft path hardcodes layer index `il = 1` when invoking
`metal_graph_encode_decode_layer()` for the MTP block. `il` is then used to derive
the layer's compression *and* RoPE behavior via `ds4_layer_compress_ratio(il)`.
This is correct only when layer 1 is dense. On **Flash** layer 1 is dense
(`compress_ratios[1] == 0`), so MTP works. On **Pro** every main layer is
compressed (`compress_ratios = [128,128,4,128,4,…,4,0]`; only the MTP's own index
`n_layer` is 0), so `il = 1` makes the engine treat the dense MTP block as
compressed.

## Symptom (Pro + `--mtp`)

Per generated token:

```
ds4: Metal graph compressor expects paired F16 compressor projections
```

Output is still correct (verifier falls back), but speculation is effectively
disabled and stderr is flooded.

## Root cause

`ds4.c`, MTP draft call (~line 19071):

```c
ok = metal_graph_encode_decode_layer(g, mtp_model, &mtp->block,
                                     1,            // <-- borrowed dense layer index; Flash-only
                                     pos, g->mtp_raw_cache, ...);
```

Inside the callee:

- `const bool compressed = ds4_layer_compress_ratio(il) != 0;`  → true on Pro for il=1
  → requires `attn_compressor_kv/gate/ape/norm`, which the (dense) MTP block lacks
  → "expects paired F16 compressor projections", `ok=false`.
- `layer_rope_freq_base(il)` / `layer_rope_freq_scale(il)` also key off
  `ds4_layer_compress_ratio(il)`, so the MTP would additionally get compressed-layer
  RoPE on Pro even if the compressor check were bypassed.

## Fix

NOTE: passing `DS4_N_LAYER` as `il` does NOT work — `ds4_layer_compress_ratio()`
(and the per-layer arrays) bounds-check `il < DS4_N_LAYER`, so the MTP's own index
trips "DeepSeek4 layer index is outside the loaded model layout".

Instead keep `il = 1` (in-bounds) and force the dense MTP block to behave densely in
`metal_graph_encode_decode_layer()`. The dense block is identifiable by its absent KV
compressor (it's never loaded for the MTP, nor for any ratio-0 layer):

```c
    const uint64_t routed_out_dim = layer->ffn_down_exps->dim[1];
-   const bool compressed = ds4_layer_compress_ratio(il) != 0;
-   const float freq_base = layer_rope_freq_base(il);
-   const float freq_scale = layer_rope_freq_scale(il);
+   const bool dense_layer = (layer->attn_compressor_kv == NULL);
+   const bool compressed = !dense_layer && ds4_layer_compress_ratio(il) != 0;
+   const float freq_base = dense_layer ? DS4_ROPE_FREQ_BASE : layer_rope_freq_base(il);
+   const float freq_scale = dense_layer ? 1.0f : layer_rope_freq_scale(il);
```

This forces `compressed=false` and dense RoPE for the MTP block, fixing both the
compressor check and the RoPE base/scale. It's a no-op for normal layers: compressed
layers always have the compressor (`dense_layer=false`); ratio-0 layers are already
dense, and their `layer_rope_freq_base/scale` already return the dense values. The MTP
keeps using its own `g->mtp_raw_cache`. A cleaner alternative would be an explicit
`is_mtp`/`force_dense` parameter threaded from the MTP call site.

## Notes

- Speculation is lossless regardless (the main model verifies every draft token), so
  this only affected acceptance/speed and the error spam, not output correctness.
- A Pro MTP support GGUF can be produced from the original DeepSeek V4 Pro safetensors
  (the checkpoint ships the module as `mtp.0.*`; routed experts are MXFP4(e2m1)+ue8m0,
  the rest F8_E4M3+ue8m0 / BF16; hyper-connection params verbatim F32).

```

```

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -14016,9 +14016,14 @@ static bool metal_graph_encode_decode_layer(
     const uint64_t expert_mid_dim = layer->ffn_gate_exps->dim[1];
     const uint64_t down_in_dim = layer->ffn_down_exps->dim[0];
     const uint64_t routed_out_dim = layer->ffn_down_exps->dim[1];
-    const bool compressed = ds4_layer_compress_ratio(il) != 0;
-    const float freq_base = layer_rope_freq_base(il);
-    const float freq_scale = layer_rope_freq_scale(il);
+    /* The MTP draft block is dense (no KV compressor) but borrows il=1, whose ratio
+     * is non-zero on Pro (all Pro main layers are compressed). Detect the dense block
+     * by its absent compressor and force dense compression + RoPE, so Pro MTP works
+     * (matches Flash, where layer 1 is naturally dense). Also covers ratio-0 layers. */
+    const bool dense_layer = (layer->attn_compressor_kv == NULL);
+    const bool compressed = !dense_layer && ds4_layer_compress_ratio(il) != 0;
+    const float freq_base = dense_layer ? DS4_ROPE_FREQ_BASE : layer_rope_freq_base(il);
+    const float freq_scale = dense_layer ? 1.0f : layer_rope_freq_scale(il);
     const float ext_factor = compressed && DS4_ROPE_SCALE_FACTOR > 1.0f ? 1.0f : 0.0f;
     float attn_factor = 1.0f;
     if (ext_factor != 0.0f && freq_scale > 0.0f) {
@@ -19071,7 +19076,7 @@ static bool metal_graph_eval_mtp_draft_from_hc(
         ok = metal_graph_encode_decode_layer(g,
                                              mtp_model,
                                              &mtp->block,
-                                             1,
+                                             1, /* in-bounds; MTP forced dense in encode_decode_layer (absent compressor) */
                                              pos,
                                              g->mtp_raw_cache,
                                              g->raw_cap,


Deviad:
I reversed engineered the MTP drafter released with the flash version and created with Opus 4.8 this file. 

**With mtp-margin 2.5 and mtp-draft 2 it achieves better results by 1.2 seconds in the test which I used.**

```
#!/usr/bin/env python3
"""
gen_spec.py — generate the authoritative ds4 MTP tensor/metadata spec from the
geometry formulas (derived from ds4.c validate_layout + DS4_SHAPE_*), for either
the Flash or Pro shape.

Why a generated spec instead of value-substituting the Flash file: several Flash
dims collide (e.g. 1024 is BOTH n_lora_q and n_lora_o; 4096 is BOTH n_embd and
n_head_dim*(n_head/n_out_group)). A blind value-substitution Flash->Pro would be
WRONG. So we compute every ne from the shape params.

Usage:
  python3 gen_spec.py flash               > spec_flash.json
  python3 gen_spec.py pro                 > spec_pro.json
  python3 gen_spec.py flash --check FLASH_MTP.gguf   # validate formulas vs the shipped file
"""
import sys, json, struct, argparse

# ---- ggml type ids (subset we use) ----
F32, F16, Q2_K, Q8_0, Q4_K = 0, 1, 10, 8, 12
TYPE_NAME = {F32: "F32", F16: "F16", Q2_K: "Q2_K", Q8_0: "Q8_0", Q4_K: "Q4_K"}
# ds4 allows only these for routed experts (ds4.c tensor_is_routed_expert_type)
EXPERT_TYPES = {"q4_k": Q4_K, "q2_k": Q2_K}

# ---- shape presets (mirror ds4.c DS4_SHAPE_FLASH / DS4_SHAPE_PRO) ----
SHAPES = {
    "flash": dict(n_embd=4096, n_head=64,  n_head_dim=512, n_lora_q=1024,
                  n_lora_o=1024, n_out_group=8,  n_expert=256, n_ff_exp=2048, n_hc=4,
                  name="DeepSeek V4 Flash MTP support"),
    "pro":   dict(n_embd=7168, n_head=128, n_head_dim=512, n_lora_q=1536,
                  n_lora_o=1024, n_out_group=16, n_expert=384, n_ff_exp=3072, n_hc=4,
                  name="DeepSeek V4 Pro MTP support"),
}

def build_spec(shape_key, expert_type=Q4_K):
    s = SHAPES[shape_key]
    E   = s["n_embd"]
    H   = s["n_head"]
    HD  = s["n_head_dim"]
    LQ  = s["n_lora_q"]
    LO  = s["n_lora_o"]
    OG  = s["n_out_group"]
    NE  = s["n_expert"]
    FF  = s["n_ff_exp"]
    HC  = s["n_hc"]

    hc_dim  = E * HC                       # 16384 flash / 28672 pro
    hc_mix  = 2 * HC + HC * HC             # 24 (both)
    q_dim   = H * HD                       # 32768 flash / 65536 pro
    out_low = OG * LO                      # 8192 flash / 16384 pro
    oa_d0   = HD * (H // OG)               # 4096 (both)

    # (name, ggml_type, ne)  — order matches the shipped Flash MTP GGUF dump.
    T = [
        ("mtp.0.hc_head_base.weight",  F32,  [HC]),
        ("mtp.0.hc_head_fn.weight",    F32,  [hc_dim, HC]),
        ("mtp.0.hc_head_scale.weight", F32,  [1]),
        ("mtp.0.hc_attn_base.weight",  F32,  [hc_mix]),
        ("mtp.0.hc_ffn_base.weight",   F32,  [hc_mix]),
        ("mtp.0.hc_attn_fn.weight",    F32,  [hc_dim, hc_mix]),
        ("mtp.0.hc_attn_scale.weight", F32,  [3]),
        ("mtp.0.hc_ffn_fn.weight",     F32,  [hc_dim, hc_mix]),
        ("mtp.0.hc_ffn_scale.weight",  F32,  [3]),
        ("mtp.0.attn_sinks.weight",    F32,  [H]),
        ("mtp.0.attn_q_a.weight",      Q8_0, [E, LQ]),
        ("mtp.0.attn_q_b.weight",      Q8_0, [LQ, q_dim]),
        ("mtp.0.attn_q_a_norm.weight", F32,  [LQ]),
        ("mtp.0.attn_output_a.weight", Q8_0, [oa_d0, out_low]),
        ("mtp.0.attn_kv.weight",       Q8_0, [E, HD]),
        ("mtp.0.attn_kv_a_norm.weight",F32,  [HD]),
        ("mtp.0.attn_output_b.weight", Q8_0, [out_low, E]),
        ("mtp.0.attn_norm.weight",     F32,  [E]),
        ("mtp.0.ffn_norm.weight",      F32,  [E]),
        ("mtp.0.ffn_gate_shexp.weight",Q8_0, [E, FF]),
        ("mtp.0.ffn_up_shexp.weight",  Q8_0, [E, FF]),
        ("mtp.0.ffn_down_shexp.weight",Q8_0, [FF, E]),
        ("mtp.0.ffn_gate_exps.weight", expert_type, [E, FF, NE]),
        ("mtp.0.ffn_up_exps.weight",   expert_type, [E, FF, NE]),
        ("mtp.0.ffn_down_exps.weight", expert_type, [FF, E, NE]),
        ("mtp.0.ffn_gate_inp.weight",  F32,  [E, NE]),
        ("mtp.0.exp_probs_b.bias",     F32,  [NE]),
        ("mtp.0.e_proj.weight",        Q8_0, [E, E]),
        ("mtp.0.h_proj.weight",        Q8_0, [E, E]),
        ("mtp.0.enorm.weight",         F32,  [E]),
        ("mtp.0.hnorm.weight",         F32,  [E]),
        ("mtp.0.norm.weight",          F32,  [E]),
    ]
    meta = [
        {"key": "general.architecture",            "type": "str", "value": "deepseek4_mtp_support"},
        {"key": "general.name",                    "type": "str", "value": s["name"]},
        {"key": "deepseek4.nextn_predict_layers",  "type": "u32", "value": 1},
        {"key": "deepseek4.mtp_layer_count",       "type": "u32", "value": 1},
        {"key": "deepseek4.expert_count",          "type": "u32", "value": NE},
    ]
    tensors = [{"name": n, "type": t, "type_name": TYPE_NAME[t], "ne": ne} for (n, t, ne) in T]
    return {"shape": shape_key, "metadata": meta, "tensors": tensors}

# ---- raw GGUF header reader (for --check), no deps ----
def read_gguf_header(path):
    f = open(path, "rb")
    def rd(n): return f.read(n)
    def u32(): return struct.unpack("<I", rd(4))[0]
    def u64(): return struct.unpack("<Q", rd(8))[0]
    def gstr():
        n = u64(); return rd(n).decode("utf-8", "replace")
    def skipval(t):
        # only needs to traverse; we don't keep most values
        simple = {0:1,1:1,2:2,3:2,4:4,5:4,6:4,7:1,10:8,11:8,12:8}
        if t in simple: rd(simple[t]); return None
        if t == 8: return gstr()
        if t == 9:
            ct = u32(); n = u64()
            if ct == 8:
                vals = [gstr() for _ in range(n)]; return vals
            sz = {0:1,1:1,2:2,3:2,4:4,5:4,6:4,7:1,10:8,11:8,12:8}.get(ct,4)
            rd(sz*n); return None
        raise ValueError(f"bad type {t}")
    assert rd(4) == b"GGUF"
    u32()  # version
    nt = u64(); nkv = u64()
    meta = {}
    for _ in range(nkv):
        k = gstr(); t = u32(); v = skipval(t); meta[k] = v
    tensors = []
    for _ in range(nt):
        name = gstr(); nd = u32(); ne = [u64() for _ in range(nd)]; tt = u32(); u64()
        tensors.append({"name": name, "type": tt, "ne": ne})
    return meta, tensors

def check(spec, gguf_path):
    _, got = read_gguf_header(gguf_path)
    want = spec["tensors"]
    gi = {t["name"]: t for t in got}
    ok = True
    print(f"checking {len(want)} generated tensors against {gguf_path} ({len(got)} tensors)")
    for w in want:
        g = gi.get(w["name"])
        if not g:
            print(f"  MISSING in gguf: {w['name']}"); ok = False; continue
        if g["type"] != w["type"] or g["ne"] != w["ne"]:
            print(f"  MISMATCH {w['name']}: gen type={w['type']} ne={w['ne']} | gguf type={g['type']} ne={g['ne']}")
            ok = False
    extra = set(gi) - {w["name"] for w in want}
    for e in sorted(extra): print(f"  EXTRA in gguf (not generated): {e}")
    print("RESULT:", "PASS — formulas reproduce the shipped file" if ok and not extra else "FAIL")
    return ok and not extra

if __name__ == "__main__":
    ap = argparse.ArgumentParser()
    ap.add_argument("shape", choices=list(SHAPES))
    ap.add_argument("--experts", choices=list(EXPERT_TYPES), default="q4_k",
                    help="routed-expert quant type (default q4_k)")
    ap.add_argument("--check", metavar="GGUF", help="validate generated spec against a real MTP gguf")
    a = ap.parse_args()
    spec = build_spec(a.shape, EXPERT_TYPES[a.experts])
    if a.check:
        sys.exit(0 if check(spec, a.check) else 1)
    json.dump(spec, sys.stdout, indent=2)
    print()
```



Here are the logs:
```
parse() { grep "mtp timing" | python3 -c '

import sys,re; tok=t=c2=c1=0

for L in sys.stdin:

  m=re.search(r"committed=(\d+)",L); v=int(m.group(1)) if m else 0

  tok+=v; c2+=v==2; c1+=v==1

  tt=re.search(r"total=([\d.]+) ms",L); t+=float(tt.group(1)) if tt else 0

print(f"steps={c1+c2} accept2%={100*c2/max(c1+c2,1):.0f} tokens={tok} {t/max(tok,1):.1f}ms/tok {1000*tok/max(t,1):.2f}tok/s")'; }



for m in 3 2.5 2; do echo -n "margin $m: "; DS4_MTP_TIMING=1 /Users/spotted/projects/ds4/ds4 \

  -m "$PRO" --metal --nothink --temp 0 --mtp "$Q4" --mtp-draft 2 --mtp-margin $m \

  -p "Write a quicksort in C and explain it." 2>&1 | parse; done

margin 3:

steps=473 accept2%=47 tokens=696 72.3ms/tok 13.82tok/s

margin 2.5: steps=408 accept2%=57 tokens=641 68.1ms/tok 14.68tok/s

margin 2: steps=384 accept2%=63 tokens=627 69.1ms/tok 14.47tok/s

```




```
 ./ds4 -m "$PRO" --metal --nothink --temp 0 -p "Explain MLA attention in two sentences."

ds4: Metal device Apple M3 Ultra, 512.00 GiB RAM

ds4: Metal 4 tensor API disabled for pre-M5/pre-A19 devices

ds4: drift-patch flags hc_stable=on norm_unify=on kv_raw_f32=off rope_exp2_log2=off math_safe=off tensor_matmul=off

ds4: requesting Metal residency (may take tens of seconds)... done

ds4: warming Metal model views... done

ds4: Metal model views created in 13.337 ms, residency requested in 3728.510 ms, warmup 10.736 ms (mapped 443098.05 MiB from offset 5.12 MiB)

ds4: Metal mapped mmaped model as 2 overlapping shared buffers

ds4: metal backend initialized for graph diagnostics

ds4: context buffers 1860.03 MiB (ctx=32768, backend=metal, prefill_chunk=8192, raw_kv_rows=8192, compressed_kv_rows=8194)

ds4: using GPU graph generation with graph prefill

ds4: gpu prefill layer 61/610/16 (0.0%)ds4: gpu prefill layer 1/61

processing 16 input tokens: 16/16 (100.0%)

MLA (Multi-head Latent Attention) compresses the key-value cache per attention head into a low-rank latent vector, drastically reducing memory usage during inference. It reconstructs full keys and values on the fly from this compact representation, preserving expressive multi-head attention while enabling efficient long-context generation.

ds4: prefill: 23.59 t/s, generation: 13.49 t/s
```

Deviad:
@antirez Could this one be interesting?

PR #418, opened 2026-06-16
Metal: FP8-packed compressed-KV cache + long-context memory optimizations

aledesogusbusiness-hue:
## Summary

Three memory optimizations for the Metal backend's compressed-KV (MLA latent) path that together cut context-dependent KV memory by **~2.3x** with **bit-identical** output and **speed-neutral** decode throughput.

- **Packed FP8 comp cache** (opt-in, `DS4_METAL_FP8_KV_STORE=1`): stores comp rows as e4m3 + ue8m0 scale + f16 rot = 584 B/row vs 1024 B/row f16. Dequant uses a 128-entry LUT avoiding branch+exp2.
- **`comp_mask` stored as f16** (always on): binary −∞/0 mask fits exactly in f16, halving the mask buffer at all context lengths.
- **`indexer_scores` token-tiling** (always on): `DS4_INDEXER_SCORE_TILE=512` reduces the score working buffer from `comp_cap × prefill_cap` to `comp_cap × 512` (~8x reduction).

## Results

| Context | KV cache reduction | Output diff |
|---------|-------------------|-------------|
| 8k–96k  | ~2.3x             | Bit-identical |

Decode speed: break-even (bandwidth saving offset by per-element e4m3 dequant).

## Controls

- `DS4_DISABLE_KV_OPTS=1` — reverts to pre-optimization layout for A/B comparison
- `DS4_METAL_FP8_KV_STORE=1` — enables packed FP8 (opt-in, precision change)

## Test environment

Apple M5 Max · DeepSeek-V4-Flash · macOS 27.0 · Metal 4.1 · 8k–96k context

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -10302,6 +10302,32 @@ static void print_vec_stats(const char *name, const float *x, uint64_t n) {
 #define DS4_GPU_ATTN_COMP_CACHE_F16 0
 #endif
 
+/* The indexer score matrix [comp_cap x n_tokens] is the largest context-dependent
+ * scratch buffer at long context. Top-k selection is independent per token, so the
+ * prefill/decode-batch score+top-k is processed in token tiles of this width,
+ * letting the scores buffer be comp_cap*TILE instead of comp_cap*prefill_cap
+ * (exact, zero drift). prefill_cap=4096 -> ~8x smaller scores scratch. */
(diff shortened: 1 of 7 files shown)


fry69:
This is the same as #416 ?

Issue #404, opened 2026-06-13
ds4-server crash: BUG IN CLIENT OF LIBMALLOC — memory corruption of free block in Metal tensor free path (EXC_BREAKPOINT SIGTRAP)


EdwQ:
## Summary

ds4-server consistently crashes with `EXC_BREAKPOINT (SIGTRAP)` detected by `libsystem_malloc.dylib` as `BUG IN CLIENT OF LIBMALLOC: memory corruption of free block` during task execution (Metal inference). The crash occurs in a worker thread while freeing Metal tensors, not in the main accept thread.

## Crash Details

**Crash count**: 9+ crashes in 24 hours (2026-06-12 to 2026-06-13)

**Exception**:  — macOS malloc internal assertion

**Faulting library**: `libsystem_malloc.dylib`

**Message**: `BUG IN CLIENT OF LIBMALLOC: memory corruption of free block`

**Faulting thread**: Worker thread (thread 2–4, not main thread which is in `accept`)

**Source**: `ds4_server.c:client_main` → `generate_job` → `ds4_session_eval` → `metal_graph_eval_token_raw_swa`

**Involved code**: `ds4_metal.m:ds4_gpu_tensor_free` — uses `__bridge_transfer` to transfer ownership to ARC

### Primary Issue: Thread-unsafe Metal tensor free

`ds4_gpu_tensor_free` in `ds4_metal.m` uses `__bridge_transfer` which tells ARC to release the `DS4MetalTensor` object. When multiple worker threads concurrently free tensors (e.g., during concurrent task processing or session cleanup racing with new task execution), ARC can double-release the same object, causing malloc to detect metadata corruption.

Key problematic code:
\`\`objc
void ds4_gpu_tensor_free(ds4_gpu_tensor *tensor) {
    if (!tensor) return;
    @autoreleasepool {
        DS4MetalTensor *obj = (__bridge_transfer DS4MetalTensor *)tensor;
        // ARC releases obj when @autoreleasepool drains
        ...
    }
}
\`\`objc

The global tracking counters `g_tensor_alloc_live_bytes` and `g_tensor_alloc_peak_bytes` also have no thread-safe access protection.

### Secondary Issue: No double-free guard

There is no tracking mechanism to prevent the same tensor pointer from being freed twice. If a tensor is freed from multiple paths (e.g., `metal_graph_free` racing with a concurrent `ds4_session_eval`), the second free would cause ARC to release an already-released object.

## Fix Implemented

[...]

PR #419, opened 2026-06-16
server: expose only the loaded model in /v1/models

### LINKED ISSUE #414: /v1/models exposes models that are not actually loaded
The server currently loads exactly one GGUF model at startup, but the model metadata endpoints make it look like both DeepSeek V4 Flash and DeepSeek V4 Pro are available at the same time.

For example, when the server is started with a Flash GGUF:

```sh
./ds4-server -m <flash.gguf> --host 127.0.0.1 --port 8000
```
```sh
curl http://127.0.0.1:8000/v1/models

The response advertises both:

deepseek-v4-flash
deepseek-v4-pro
  ```

This is misleading because the model field in API requests does not actually switch the loaded GGUF. Inference always uses the single model loaded at server startup.

The same problem also affects GET /v1/models/<id>: the server accepts both deepseek-v4-flash and deepseek-v4-pro as valid metadata endpoints, even if only one of those models is actually loaded.

Expected behavior:
- If a Flash GGUF is loaded, /v1/models should expose only deepseek-v4-flash
- If a Pro GGUF is loaded, /v1/models should expose only deepseek-v4-pro
- GET /v1/models/<id> should return 404 for the non-loaded model id

This matters for OpenAI-compatible clients that inspect /v1/models to decide which model IDs are available. The current behavior can make clients believe both variants are selectable, while the server can only run the already-loaded GGUF.

## Proposed changes

Fix this by making model metadata reflect the single GGUF loaded at startup. Implemented in https://github.com/antirez/ds4/pull/287.

aledesogusbusiness-hue:
## Problem

`GET /v1/models` always returns both `deepseek-v4-flash` and `deepseek-v4-pro` regardless of which GGUF is actually loaded at startup.  `GET /v1/models/<id>` likewise accepts either alias.

This misleads OpenAI-compatible clients that inspect `/v1/models` to decide which models are available — they see two choices but the server always runs the single GGUF it was started with.

## Fix

- `send_models()`: replace the hardcoded two-entry list with a single call to `server_model_id_from_engine(s->engine)`.
- `/v1/models/<id>` handler: compare the requested ID against `server_model_id_from_engine(s->engine)` directly; return 404 for the non-loaded model.
- Remove `server_model_alias_known()` which is now unused.

## Test plan

- Start server with a Flash GGUF → `GET /v1/models` returns only `deepseek-v4-flash`; `GET /v1/models/deepseek-v4-pro` returns 404.
- Start server with a Pro GGUF → opposite behaviour.

Fixes #414

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -898,12 +898,6 @@ static const char *server_model_id_from_engine(ds4_engine *engine) {
            "deepseek-v4-pro" : "deepseek-v4-flash";
 }
 
-static bool server_model_alias_known(const char *id) {
-    return id &&
-           (!strcmp(id, "deepseek-v4-flash") ||
-            !strcmp(id, "deepseek-v4-pro"));
-}
-
 static void stop_list_clear(stop_list *stops) {
     for (int i = 0; i < stops->len; i++) free(stops->v[i]);
     stops->len = 0;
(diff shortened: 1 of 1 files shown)


fry69:
This is the same as #287?

derogab:
> This is the same as #287?

Yes, I think it is. #287 was the one I proposed when I opened the related issue #404.

#419 is a smaller fix and looks OK for the endpoint behavior. #287 also includes the README update, a server unit test for Flash/Pro metadata, and an explicit id/name mapping so `deepseek-v4-pro` reports `DeepSeek V4 Pro` instead of incorrect name (`DeepSeek V4 Flash`).

PR #430, opened 2026-06-16
Add reverse distributed topology with coordinator-owned output suffix

### LINKED ISSUE #428: Allow coordinator to own N:output layers in distributed mode
Currently the only way one can run distributed generation is when the coordinator owns early layers (0:N-1), and the worker owns later layers plus optionally the output (N:42 or N:output). This is matters little when both nodes are homogenuous, but it makes it difficult to extend distributed generation options. For example, I ran into problems while working on #304 making local decode compatible with all the existing frontends, so I had to postpone that work and switch to this first as a prerequisite.

This is propose extension of distributed mode allowing "reverse topology" whereby a coordinator running any of the frontends (ds4, ds4-server, ds4-eval, etc) to "own" later layers and the output, while allowing the worker to own the earlier layers.

While not materially changing the way current distributed mode works, this would extending distributed topologies more naturaly, because the frontends rely on `ds4_session_sync()` -> `ds4_session_eval()` | `ds4_session_sample()` core loop switches to per-token RPC when output is owned remotely.

PR is coming shortly.

lobanov:
**Summary**

This change closes #428 adding a second distributed inference topology where the coordinator can own a final contiguous suffix through output, using `--layers K:output`, while workers cover the lower layers `0..K-1`.

The goal is that the public session loop behaves the same regardless of topology:
- `ds4_session_sync()`
- `ds4_session_eval()`
- `ds4_session_sample()`

Forward topology remains unchanged:
- coordinator owns `0:K`
- workers cover higher layers
- existing forward `N:42` behavior is preserved

New reverse topology:
- coordinator owns `K:output`
- workers cover the lower prefix
- reverse `K:42` is explicitly unsupported

This enables all existing frontends to run with a coordinator that prefills higher layers and runs the output head locally, without topology-specific frontend changes.

**Main implementation changes**

`ds4_distributed.c`
- Added internal topology inference and validation for coordinator-owned prefix vs suffix.
- Generalized route planning so the coordinator can plan either:
  - forward routes covering layers after its local slice
  - reverse routes covering layers before its local slice
- Updated route reporting and completeness checks to be topology-aware.
- Made decode execution topology-aware:
  - forward remains local-first then remote
  - reverse is remote-first then local suffix/output
- Extended serial prefill to work in reverse topology through the same session-facing flow.
- Implemented reverse pipelined prefill:
  - workers process each chunk first
  - each chunk returns hidden state upstream
  - coordinator applies its local suffix on every returned chunk
  - reverse mode does not use forward-style non-final `ACK_ONLY`
- Hardened worker-side route validation so remote-first routes must return hidden state upstream rather than remote logits.
- Generalized distributed KV shard ownership/save/load logic from “local shard first” to ordered ownership by layer range, so reverse topology persists correctly.
- Fixed reverse pipelined prefill reset symmetry so checkpoint/suffix sync cannot reset the local suffix when the remote prefix was intentionally not reset.

`README.md`
- Documented reverse distributed topology and coordinator `K:output` usage.
- Clarified protocol behavior for forward vs reverse execution and pipelined prefill.

`ds4_help.c`
- Updated distributed help text to describe both supported coordinator layouts and note that reverse `K:42` is unsupported.

**Main function-level changes**

Key functions updated or introduced in `ds4_distributed.c`:
- `dist_infer_coordinator_topology()`
- `dist_validate_layers_for_model()`
- `dist_coordinator_route_span()`
- `dist_coordinator_build_route_plan()`
- `dist_coordinator_report_plan()`
- `dist_coordinator_send_remote_work_on_fd()`
- `dist_coordinator_request_remote_on_fd()`
- `dist_coordinator_eval_local_suffix()`
- `dist_coordinator_eval_span()`
- `dist_coordinator_can_pipeline_prefill()`
- `dist_coordinator_prefill_prompt_pipelined()`
- `dist_coordinator_prefill_prompt()`
- `dist_worker_process_work_payload()`
- `dist_kv_route_build_owners()`
- `ds4_dist_session_save_payload()`
- `ds4_dist_session_load_payload()`

**Added tests**

Automated tests:
- Added `--distributed-topology-logic` to `ds4_test`.
- This covers:
  - coordinator layer validation for forward `0:K`
  - coordinator layer validation for reverse `K:output`
  - rejection of reverse `K:42`
  - rejection of middle-only coordinator slices
  - forward route planning over synthetic worker layouts
  - reverse route planning over synthetic worker layouts
  - assertion that reverse routes do not assign remote output logits
  - KV owner ordering for reverse topology
  - KV owner gap rejection

Test support:
- Added `tests/ds4_distributed_test_internal.h` as a test-only bridge exposing the minimum internal distributed structs and thin internal entry points needed by `tests/ds4_test.c`.
- Kept test setup logic in the test file rather than adding heavy public test helper APIs.

**Validation performed**

Build validation:
- `make`
- `make ds4_test`
- `./ds4_test --distributed-topology-logic`
- `make test` (skipped streaming-decode-prefill-correctness and mtp-verify-depth)
- `make cpu`

Manual/local runtime distributed validation on localhost:
- Forward topology:
  - worker `20:output`
  - coordinator `0:19`
  - verified route formation, prompt prefill, and one-token generation
- Reverse topology:
  - worker `0:19`
  - coordinator `20:output`
  - verified route formation, prompt prefill, and one-token generation
- Pipelined prefill:
  - ran long-prompt tests with `--dist-prefill-chunk 8 --dist-prefill-window 2`
  - confirmed pipelined distributed prefill in both forward and reverse mode
  - confirmed successful generation after prefill in both modes
- Running `ds4-eval` in distributed mode with reverse topology:
  - Responses to `--questions 4` match reference
  - Full suite produces result: `62/92 passed, 30 failed, runtime 00h:55m`

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,6 +385,30 @@ slow or metered links, `--layers 20:42` is also supported: the coordinator will
 load the output head and compute logits locally, trading extra coordinator work
 for smaller per-token replies.
 
+Reverse topology is also supported when the coordinator owns a final suffix
+through the output head. In that layout workers cover the lower layers,
+return hidden state upstream, and the coordinator runs the higher layers plus
+output locally. For example:
+
+```sh
+# Machine A: worker, owns lower layers.
+./ds4 \
+  -m gguf/DeepSeek-V4-Pro-Q4K-Layers00-30.gguf \
+  --role worker \
+  --layers 0:30 \
+  --coordinator 169.254.43.68 1234
+
+# Machine B: coordinator, owns the upper suffix and output.
+./ds4 \
+  -m gguf/DeepSeek-V4-Pro-Q4K-Layers-31-output.gguf \
+  --role coordinator \
+  --layers 31:output \
+  --listen 169.254.43.68 1234
+```
+
+Reverse `K:42` is intentionally unsupported. Reverse mode only supports
+`K:output`, because the coordinator must own the output head.
+
 ### Network Link Comparison
 
 The table below shows the same two M5 Max hosts, the same 91 GB Flash quant,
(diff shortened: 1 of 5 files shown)


lobanov:
Rebased on updated `main`

PR #447, opened 2026-06-23
Protect incoming KV prefix during live miss

JordiPosthumus:
## Summary\n- protect disk KV checkpoints that are valid prefixes of the incoming prompt while storing old live state after a live-cache miss\n- add server-level --nothink/--think defaults and document them\n- add a regression test for the live-miss eviction case\n\n## Why\nOn a live-cache miss, ds4-server stores the old live checkpoint before trying to load a disk fallback. Under disk pressure that store can evict a checkpoint that prefixes the incoming prompt, forcing a cold prefill even though a valid fallback existed.\n\n## Test\n- make test\n


JordiPosthumus:
Withdrawing this PR for now.

PR #448, opened 2026-06-23
Protect incoming KV prefix during live miss

JordiPosthumus:
## Summary
- Protect disk KV checkpoints that are valid prefixes of the incoming prompt while storing old live state after a live-cache miss.
- Add a regression test for the live-miss eviction case.

## Why
On a live-cache miss, ds4-server stores the old live checkpoint before trying to load a disk fallback. Under disk pressure, that store can evict a checkpoint that prefixes the incoming prompt. When that happens, ds4 has to cold-prefill even though a valid disk fallback existed.

This change gives those incoming-prompt prefix checkpoints a very high eviction score during the pre-fallback store, so unrelated cache entries are evicted first.

## Test
- make ds4-server
- make test

### DIFF
--- a/ds4_kvstore.c
+++ b/ds4_kvstore.c
@@ -523,6 +523,24 @@ static bool kv_cache_incoming_supersedes_continued(
     return !strcmp(prefix_sha, e->sha);
 }
 
+static bool kv_cache_incoming_protects_prefix(
+        const ds4_kvstore_entry *e,
+        const ds4_kvstore_eviction_context *incoming) {
+    if (!e || !incoming || !incoming->protect_text) return false;
+    if (e->text_bytes == 0 || e->text_bytes > SIZE_MAX) return false;
+    if ((size_t)e->text_bytes > incoming->protect_text_len) return false;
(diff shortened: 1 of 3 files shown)


fry69:
Can you please fix your PR text? It's garbled.

There are still humans reading PRs... (maybe not for long, though)

JordiPosthumus:
ok, I spoke to my people ! Better? :-D

unsaltedbutter-ai:
@antirez I've been running this on my branch. I interleave several conversations on one ds4 server, so live misses are happening daily, so this path gets hit constantly. Once the disk cache is at budget the pre-fallback store was evicting the exact prefix the fallback load then needed. This removes that self-inflicted miss. Please have a look 👍

JordiPosthumus:
Closing because this implementation is now stale against current `main` and conflicts with the current per-slot server/cache architecture. The underlying live-miss eviction bug remains valid; I plan to reproduce it on current `main` and return with a smaller current implementation rather than ask maintainers to review this obsolete diff. Thanks to everyone who tested and commented.

PR #460, opened 2026-06-26
CUDA: batch gate/up/down uploads for selected expert cache misses

fmolara:
### Motivation

In the CUDA direct-I/O path, a selected expert cache miss uploads the gate/up/down tensors independently, synchronizing the selected upload stream after each tensor.

This patch batches the three uploads belonging to the same expert and synchronizes the selected upload stream once after all three have been enqueued.

### Properties

- no API changes
- no kernel changes
- no cache policy changes
- resident and compact cache layouts are unchanged
- unsupported cases fall back to the existing path

### Validation

Tested after rebasing onto `efdadd4`, which was the current `main` at the time, on an NVIDIA A100 80GB PCIe with CUDA 12.4 (`sm_80`).

* Baseline commit: `efdadd4`
* PR commit: `6ece876`
* Clean CUDA builds: passed
* Deterministic output: identical
* Normalized output SHA-256: identical
* No CUDA staging, upload, or synchronization errors

#### Targeted miss-heavy Nsight Systems validation

| Expert cache |                           Metric | Baseline |      PR | Difference |
| ------------ | -------------------------------: | -------: | ------: | ---------: |
| 8GB          |    `cudaStreamSynchronize` calls |  108,276 |  36,092 |    −66.67% |
| 8GB          | `cudaStreamSynchronize` API time |  1.649 s | 0.587 s |    −64.42% |
| 16GB         |    `cudaStreamSynchronize` calls |  108,276 |  36,092 |    −66.67% |
| 16GB         | `cudaStreamSynchronize` API time |  1.742 s | 0.584 s |    −66.49% |

The number of `cudaMemcpyAsync` calls remained unchanged at 108,276.

This confirms that the patch preserves the same gate/up/down uploads while reducing synchronization from once per tensor upload to once per three-upload group.

End-to-end throughput differences were within normal run-to-run noise:

| Expert cache |     Metric | Baseline median | PR median |
| ------------ | ---------: | --------------: | --------: |
| 8GB          |    Prefill |        2.45 t/s |  2.41 t/s |
| 8GB          | Generation |        0.65 t/s |  0.65 t/s |
| 8GB          |  Wall time |        220.53 s |  221.08 s |
| 16GB         |    Prefill |        2.40 t/s |  2.42 t/s |
| 16GB         | Generation |        0.64 t/s |  0.64 t/s |
| 16GB         |  Wall time |        221.89 s |  222.31 s |

This change should therefore be considered a synchronization-granularity improvement, not a demonstrated overall throughput improvement on the current upstream code.

### DIFF
--- a/ds4_cuda.cu
+++ b/ds4_cuda.cu
@@ -1744,6 +1744,87 @@ static int cuda_model_copy_to_device_streamed(
     return 1;
 }
 
+/* Return -1 when this expert needs the general streamed-copy path. */
+static int cuda_model_copy_expert_to_device_streamed(
+        char *const dst[3],
+        const void *model_map,
+        uint64_t model_size,
+        const uint64_t offset[3],
+        const uint64_t bytes[3],
+        const char *const what[3]) {
+    if (g_model_fd < 0 ||
+        (g_model_fd_host_base != NULL && model_map != g_model_fd_host_base)) {
+        return -1;
+    }
+
+    const uint64_t chunk = cuda_model_copy_chunk_bytes();
+    for (uint32_t i = 0; i < 3; i++) {
+        if (!dst[i] || offset[i] > model_size ||
+            bytes[i] == 0 || bytes[i] > model_size - offset[i] ||
+            bytes[i] > chunk) {
+            return -1;
+        }
+    }
+
+    const uint64_t stage_bytes =
+        chunk + (g_model_direct_align > 1 ? g_model_direct_align : 1);
+    if (!cuda_stream_selected_stage_pool_alloc(stage_bytes)) return 0;
+
+    int enqueued = 0;
+    for (uint32_t i = 0; i < 3; i++) {
+        const char *payload = NULL;
+        if (!cuda_model_stage_read(g_stream_selected_stage[i],
+                                   g_stream_selected_stage_bytes,
+                                   offset[i], bytes[i], &payload)) {
+            fprintf(stderr,
+                    "ds4: CUDA streaming selected read failed for %s: %s\n",
+                    what[i], strerror(errno));
+            if (enqueued) {
+                (void)cudaStreamSynchronize(g_stream_selected_upload_stream);
+            }
+            return 0;
+        }
+        cudaError_t err = cudaMemcpyAsync(dst[i], payload, (size_t)bytes[i],
+                                          cudaMemcpyHostToDevice,
+                                          g_stream_selected_upload_stream);
+        if (err != cudaSuccess) {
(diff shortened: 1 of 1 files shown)


fmolara:
Rebased onto current `main` and adapted the patch to the refactored selected expert cache path.

The change remains narrowly scoped:

* gate/up/down use three distinct staging slots;
* all three copies are enqueued on the selected upload stream;
* the stream is synchronized once after the three uploads;
* unsupported cases transparently use the existing three-copy fallback;
* pending stream work is drained on partial failure;
* cache policy, tensor layouts, kernels, and diagnostics are unchanged.

Both the current baseline (`efdadd4`) and the PR commit (`6ece876`) build successfully with CUDA 12.4 on an NVIDIA A100 80GB PCIe (`sm_80`).

All validation runs produced identical deterministic output and the same normalized output SHA-256, with no CUDA staging, upload, or synchronization errors.

A targeted Nsight Systems comparison confirmed the intended effect:

* `cudaMemcpyAsync`: unchanged at 108,276 calls
* `cudaStreamSynchronize`: 108,276 → 36,092 calls (`−66.67%`)
* `cudaStreamSynchronize` API time:

  * 8GB expert cache: 1.649 s → 0.587 s (`−64.42%`)
  * 16GB expert cache: 1.742 s → 0.584 s (`−66.49%`)

This removes exactly 72,184 stream synchronizations in the profiled workload while preserving the upload count and output correctness.

End-to-end throughput remained within run-to-run noise, so the validation notes no longer claim a measurable overall throughput improvement on the current `main`.

fmolara:
I reviewed this PR against current main, 6e4c285a92a1de11863784b748f196b5d4b88a48.
Its optimization is now superseded by the request-wide upload batching added in a04f46fa423e45712c8c7e430eff422479f314a3.

The new path carries one staging-ring chunk counter across gate/up/down copies for all cache misses in a request, and drains the upload stream once after the miss loop. This PR drained it once per expert. Event waits still protect staging-buffer reuse, and the batch destructor drains pending transfers on early returns. The subsequent prefetch work also coordinates cancellation and cache teardown.

I built current main for sm_80 with CUDA 12.4 on an A100 80GB and passed the CUDA SSD cache and scalar/batch tests, including eviction/remapping and the partial-upload-failure cases, plus the host SSD cache sizing test.

I recommend closing this PR as superseded rather than rebasing it: its commit has not been merged, but its intended functionality is already covered by the newer, broader batching implementation. No replacement patch is needed for the original scope. These targeted checks are not an end-to-end model performance benchmark.

PR #466, opened 2026-06-27
Fix CUDA MoE router hardcoded to 256 experts

slackarea:
The CUDA MoE router rejects any model whose routed-expert count isn't 256:
`ds4_gpu_router_select_tensor` / `_batch_tensor` do
`if (n_expert != 256u || n_expert_used != 6u || fabsf(scale-1.5f)>1e-6f) return 0;`
and the three router_select kernels hardcode 256 (logits/probs stride, top-k bound) and
the 1.5f routed-weight scale. On a DeepSeek-V4-Pro GGUF (384 experts) prefill fails with
`gpu layer N ffn batch encode failed` (the router-select call returns 0).

Fix (minimal, zero-regression for Flash):
- Parametrize the serial `router_select_kernel` with `n_expert` and `scale`.
- Relax both host-wrapper guards to accept 256 or 384; replace the `256u*sizeof(float)`
  bias/logits/probs checks with `n_expert`.
- Dispatch: `n_expert != 256` -> the parametrized serial kernel; 256 stays on the
  existing fast warp/parallel kernels (unchanged).

Regression (Linux, CUDA, H200, DeepSeek-V4-Flash q2-imatrix; clean main + this patch only):
- `ds4_test --logprob-vectors`: PASS
- Flash decode: 39.67 t/s (256 fast path untouched; baseline ~39 t/s)

Caveat / follow-up: still assumes `n_expert_used == 6` (the `for j<6` loops + guard) — fine
for Flash/Pro (top-6), not for other top-k (e.g. GLM-5.2 top-8). The serial path is
correctness-first (1 thread/token); the warp/parallel kernels could be parametrized for
speed (shared `sprob[256]` and the 32x8 unroll). Happy to extend if useful.

### DIFF
--- a/ds4_cuda.cu
+++ b/ds4_cuda.cu
@@ -5972,15 +5972,17 @@ __global__ static void router_select_kernel(
         uint32_t hash_rows,
         uint32_t n_tokens,
         int has_bias,
-        int hash_mode) {
+        int hash_mode,
+        uint32_t n_expert,
+        float scale) {
     uint32_t t = blockIdx.x;
     if (t >= n_tokens || threadIdx.x != 0) return;
-    const float *log = logits + (uint64_t)t * 256;
-    float *prob = probs + (uint64_t)t * 256;
+    const float *log = logits + (uint64_t)t * n_expert;
+    float *prob = probs + (uint64_t)t * n_expert;
     int32_t *sel = selected + (uint64_t)t * 6;
     float *w = weights + (uint64_t)t * 6;
 
-    for (int i = 0; i < 256; i++) prob[i] = sqrtf(softplus_dev(log[i]));
+    for (uint32_t i = 0; i < n_expert; i++) prob[i] = sqrtf(softplus_dev(log[i]));
 
     if (hash_mode) {
         int32_t tok = tokens ? tokens[t] : token_scalar;
@@ -5989,12 +5991,12 @@ __global__ static void router_select_kernel(
         for (int i = 0; i < 6; i++) sel[i] = row[i];
     } else {
         for (int i = 0; i < 6; i++) sel[i] = -1;
-        for (int i = 0; i < 256; i++) {
+        for (uint32_t i = 0; i < n_expert; i++) {
             float score = prob[i] + (has_bias ? bias[i] : 0.0f);
             for (int j = 0; j < 6; j++) {
                 if (sel[j] < 0 || score > prob[sel[j]] + (has_bias ? bias[sel[j]] : 0.0f)) {
                     for (int k = 5; k > j; k--) sel[k] = sel[k - 1];
-                    sel[j] = i;
+                    sel[j] = (int32_t)i;
                     break;
                 }
             }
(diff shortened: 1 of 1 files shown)


chrisharper77:
Confirming this on a 3-node DGX Spark ring (GB10 / sm_121) with DeepSeek V4 PRO Q2. The two host-side guards in `ds4_gpu_router_select_tensor` (line 16031) and `ds4_gpu_router_select_batch_tensor` (line 16069) reject every PRO CUDA forward pass. The `DS4_SHAPE_PRO` shape profile added in `ds4.c` at `efdadd4` correctly sets `n_expert=384`, but the CUDA router still hardcodes `n_expert != 256u`, so the CUDA path bails out entirely.

I've been running a patch on our ring that:

1. Templates the three router_select kernels (`router_select_kernel`, `router_select_parallel_kernel`, `router_select_warp_topk_kernel`) on `size_t N_EXPERT`.
2. Instantiates for 256 (Flash) and 384 (PRO).
3. Replaces the hardcoded guards with:
```c
if (n_expert_used != DS4_CUDA_N_EXPERT_USED) return 0;
if (n_expert != DS4_CUDA_N_EXPERT && n_expert != DS4_CUDA_MAX_N_EXPERT) return 0;
```
4. Passes `expert_weight_scale` as a runtime argument rather than assuming `1.5f`.

With this patch on the fork (before `efdadd4`'s CUDA rewrite), PRO Q2 runs correctly across all 3 Sparks (`--inspect` reports `count=384 used=6 variant=PRO`, output is coherent). The rebase onto `efdadd4`'s rewritten `ds4_cuda.cu` is straightforward but I'm pausing in case antirez has a preferred approach before I port it.

**Questions:**
1. Is a PR welcome with the templating approach, or is there a preferred alternative (dispatch table, runtime-parameterized kernels, etc.)?
2. Any reason the guard is intentionally still in place — e.g. PRO on CUDA deferred pending other work?
3. Should the same sweep also template the kernel loop bounds and striding that assume 256 elsewhere in `ds4_cuda.cu`?

Also noting #471 from @slackarea which documents additional Pro-specific hardcodes in the DSA indexer top-k path — those may need a joint fix alongside this guard.

Happy to open a PR against whichever approach works best.

slackarea:
Thanks for confirming this on sm_121 — your diagnosis matches ours exactly, and I can confirm the guards are still present on current main after the `efdadd4` CUDA rewrite (`ds4_gpu_router_select_tensor` / `_batch_tensor`, the `n_expert != 256u || n_expert_used != 6u || expert_weight_scale != 1.5f` checks).

This PR predates the rewrite and is now conflicting, so it needs a rebase regardless. Your templated approach is broader than my original fix (it also lifts the hardcoded `expert_weight_scale` and covers PRO's 384 experts properly), so I'd be happy to either rebase this PR minimally or step aside in favor of your patch — whichever antirez prefers.

On the joint-fix note: the DSA indexer side of #471 already has a proposed fix in #478 (indexer top_k 512→1024 + decode LUT gate for in_dim>4096). Running both together is what got PRO producing correct output on our 2×H200 box.

PR #487, opened 2026-07-02
cuda: fall back to pinned host memory when the model arena runs out of VRAM

riccardo-galbani:
`cuda_model_arena_alloc()` would fail hard (return NULL,
`g_model_cache_full = 1`) the moment `cudaMalloc` failed for a model
weight chunk. On GPUs where the non-routed (always-resident) weights
of a model exceed available VRAM, this made `--ssd-streaming`
unusable even though the model's routed experts would otherwise fit
comfortably via the existing streaming expert cache.

This adds a fallback: when the device allocation fails, allocate
pinned host memory (`cudaHostAlloc` with `cudaHostAllocMapped`) and
expose it to the GPU via `cudaHostGetDevicePointer` (zero-copy). This
keeps such models loadable and correct, at the cost of PCIe latency
on every access to the affected chunk. Set
`DS4_CUDA_NO_PINNED_ARENA_FALLBACK` to restore the previous
fail-fast behavior.

Tested on a consumer laptop low memory CUDA GPU with a model whose
non-routed weights (8.20 GiB) exceed available VRAM. Previously
this crashed with an illegal memory access / OOM during prefill.
With this patch it loads and generates correctly, though slowly due
to PCIe round-trips on the pinned chunks:

  ds4: CUDA model arena using pinned host RAM fallback for q8_0 (1792.00 MiB chunk, zero-copy device access)
  ds4: CUDA model arena using pinned host RAM fallback for q8_0_pair0 (1792.00 MiB chunk, zero-copy device access)
  ds4: CUDA model arena using pinned host RAM fallback for q8_hc_expand (1792.00 MiB chunk, zero-copy device access)
  ...
  ds4: prefill: 0.52 t/s, generation: 0.58 t/s

On GPUs with enough VRAM for a model's non-routed weights, this
fallback never triggers and behavior is unchanged.

### DIFF
--- a/ds4_cuda.cu
+++ b/ds4_cuda.cu
@@ -1177,13 +1177,50 @@ static char *cuda_model_arena_alloc(uint64_t bytes, const char *what) {
     void *dev = NULL;
     cudaError_t err = cudaMalloc(&dev, (size_t)chunk);
     if (err != cudaSuccess) {
-        fprintf(stderr, "ds4: CUDA model arena alloc failed for %s (%.2f MiB chunk): %s\n",
-                what ? what : "weights",
-                (double)chunk / 1048576.0,
-                cudaGetErrorString(err));
+        /* Device arena is full: fall back to pinned host memory mapped
+         * into the GPU's address space (zero-copy). This keeps models
+         * whose non-routed weights exceed available VRAM working, at
+         * the cost of PCIe latency on every access to this chunk. Set
+         * DS4_CUDA_NO_PINNED_ARENA_FALLBACK to restore the old
+         * fail-fast behavior. */
         (void)cudaGetLastError();
-        g_model_cache_full = 1;
-        return NULL;
+        if (getenv("DS4_CUDA_NO_PINNED_ARENA_FALLBACK") != NULL) {
+            fprintf(stderr, "ds4: CUDA model arena alloc failed for %s (%.2f MiB chunk): %s\n",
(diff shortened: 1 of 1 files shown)


riccardo-galbani:
Tested on NVIDIA GB10 (DGX Spark, Grace Blackwell, 128GB unified
memory, sm_121, CUDA 13.0, driver 580.159.03) to verify behavior on
hardware-coherent unified-memory systems.

As expected on this architecture, the fallback never triggers:
`cudaMalloc` draws from the single unified pool, so the OOM condition
that activates the fallback on discrete GPUs (device memory exhausted
while system RAM is available) cannot occur — the model's non-routed
weights (8.20 GiB) allocate directly without ever reaching the
fallback path. No behavioral change, no overhead, no regressions
observed on this hardware.

Worth noting for completeness: if the unified pool itself were
exhausted (e.g. a model larger than total system memory), the
fallback's `cudaHostAlloc` would fail too, since there is no separate
reserve to draw from on unified memory. That double-failure path is
handled: it logs both errors explicitly and fails cleanly, identical
to the pre-patch behavior (and with a more diagnostic error message).

In summary: this fallback is specifically useful on discrete GPUs
whose VRAM is smaller than a model's always-resident weights (where
it turns a hard crash into a working, if slower, run — as measured
on an 8GB laptop GPU). On unified-memory systems it is dormant and
harmless.

PR #498, opened 2026-07-04
rocm: use numerically stable SiLU to avoid NaN on large-magnitude gate

0x006a6d:
The ROCm MoE/FFN kernels computed SiLU as the naive `gate / (1.0f + expf(-gate))`. For a large negative gate, `expf(-gate)` overflows to +inf, so the expression becomes `-inf / inf = NaN`. The CPU reference (`ds4.c` `silu`/`sigmoid_stable`) deliberately avoids this by branching on the sign, so the two backends disagreed on extreme inputs and only ROCm produced NaN — which then propagates through the rest of the network (e.g. a later router selecting expert id -1).

Add `dev_sigmoid_stable`/`dev_silu_stable` in `ds4_rocm_common.cuh` (matching the CPU sign-branched form) and replace the 27 naive SiLU sites across `ds4_rocm_moe.cuh`, `ds4_rocm_q8.cuh`, and `ds4_rocm_output.cuh`. No behavior change for well-conditioned inputs; only removes the overflow path on extreme gate values.

### DIFF
--- a/rocm/ds4_rocm_common.cuh
+++ b/rocm/ds4_rocm_common.cuh
@@ -8,6 +8,19 @@ __global__ static void fill_f32_kernel(float *x, uint64_t n, float v) {
     if (i < n) x[i] = v;
 }
 
+/* Numerically stable sigmoid/SiLU matching the CPU reference (ds4.c
+ * sigmoid_stable/silu). The naive gate/(1+expf(-gate)) overflows for large
+ * negative gate: expf(-gate) -> +inf makes it -inf/inf = NaN, which the CPU
+ * path avoids. Use these everywhere the ROCm MoE/FFN applies SiLU. */
+__device__ __forceinline__ static float dev_sigmoid_stable(float x) {
+    if (x >= 0.0f) return 1.0f / (1.0f + expf(-x));
+    const float e = expf(x);
+    return e / (1.0f + e);
+}
+__device__ __forceinline__ static float dev_silu_stable(float x) {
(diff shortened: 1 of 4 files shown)


rocm/ds4_rocm_common.cuh
```diff
@@ -8,6 +8,19 @@ __global__ static void fill_f32_kernel(float *x, uint64_t n, float v) {
...
+ * path avoids. Use these everywhere the ROCm MoE/FFN applies SiLU. */
+__device__ __forceinline__ static float dev_sigmoid_stable(float x) {
+    if (x >= 0.0f) return 1.0f / (1.0f + expf(-x));
+    const float e = expf(x);
+    return e / (1.0f + e);
+}
+__device__ __forceinline__ static float dev_silu_stable(float x) {
+    return x * dev_sigmoid_stable(x);
+}
+
```
OPS-NeoRetro, line 23: Great job. Claude did the common stable Swish.

OPS-NeoRetro (approved):
So, your numerically stable Swish that Claude wrote is directly taken from ds4.c, but it's just ported to AMD ROCm. Are you planning to test it today or tomorrow? I hope that somebody would port this fix to CUDA, or I don't think that it's there.

OPS-NeoRetro:
@Flowers-of-Romance have you tested this PR yet?

OPS-NeoRetro:
@0x006a6d have you properly tested this? Please give me the before and after results. I haven't heard anything from you because you changed your account name or something.

Issue #491, opened 2026-07-03
Add optional expert bundle sidecar for contiguous MoE expert reads


GiorgioOppo:
## Summary

This proposal adds an optional `.expertbundle` sidecar for SSD-streamed MoE
models.

The GGUF file remains unchanged. The sidecar is a derived cache where each routed
expert's tensors are stored contiguously:

```text
GGUF layout, simplified:

[gate_expert_0][gate_expert_1]...[gate_expert_N]
[up_expert_0]  [up_expert_1]  ...[up_expert_N]
[down_expert_0][down_expert_1]...[down_expert_N]

expert bundle layout:

[expert_0: gate up down]
[expert_1: gate up down]
...
[expert_N: gate up down]


When the router selects an expert and that expert is not already resident in the
expert cache, the engine can satisfy the miss with one large sequential read
instead of three scattered reads.
The model weights, quantization, and math do not change. This is only a disk
layout optimization for routed expert slabs.
Motivation
In SSD streaming mode, decode can become dominated by routed expert I/O. For
each selected expert, the engine needs the gate, up, and down projections. In
the GGUF layout those slabs are grouped by tensor kind, so loading one expert
requires multiple non-contiguous reads.
A sidecar organized per expert makes a cache miss look like:
pread(bundle_fd, dst, expert_total_bytes, expert_offset);
instead of:
pread(model_fd, gate_dst, gate_bytes, gate_offset);
pread(model_fd, up_dst,   up_bytes,   up_offset);
pread(model_fd, down_dst, down_bytes, down_offset);
This should improve the cold-miss path while preserving the existing GGUF format
and all current decode behavior.
Proposed behavior
The feature is opt-in, for example:
DS4_EXPERT_BUNDLE=1 ./ds4 model.gguf
On startup, if enabled, the engine looks for:
model.gguf.expertbundle
If the sidecar exists and matches the source GGUF, it is opened and used for
expert cache misses / expert gathers.

If the sidecar does not exist, the engine may build it once by reading the
routed expert tensors from the GGUF and writing them in per-expert order.

If there is not enough disk space, or sidecar creation fails, the engine falls
back to the current GGUF gather path.

Numerics must be identical, because the sidecar contains the same expert bytes,
only reordered.

Sidecar format sketch
A minimal header could be:
struct ds4_expert_bundle_header {
    char magic[16];           // "DS4EXPBUNDLE"
    uint32_t version;
    uint32_t n_layers;
    uint32_t n_experts;
[...]

PR #499, opened 2026-07-04
Opt-in mlock of non-routed weights for SSD streaming (DS4_MLOCK_NONROUTED)

hi-pauls:
## Summary

Adds an opt-in `DS4_MLOCK_NONROUTED=1` environment flag that `mlock()`s the mmap ranges of every non-routed-expert tensor after `parse_tensors()`. Routed experts (`.ffn_gate_exps.`, `.ffn_up_exps.`, `.ffn_down_exps.`) are explicitly skipped so they keep streaming from SSD via the existing cache path. Everything else — attention projections, shared experts, output head, embeddings, norms — becomes non-evictable.

## Motivation

On memory-constrained machines (64 GB unified memory on Apple Silicon, tested), the current `--ssd-streaming` path only wires the routed expert cache buffer. Non-routed weights are file-backed mmap; the OS evicts them under pressure and every decode step re-reads them from SSD, tanking throughput. mlock'ing them keeps the "hot skeleton" resident while routed experts continue to stream normally.

## Benchmark

64 GB M1 Max, Q2 Flash imatrix quant, 25 GB expert cache, 256K ctx, gen 8192 (single 2048-token frontier):

| Config | Decode tok/s |
|---|---:|
| Baseline (no mlock) | 1.27 |
| `DS4_MLOCK_NONROUTED=1` | 4.54 |

**+257% decode** with 8.22 GiB additional wired. Broader cross-domain sweep (coding, hardware, rocketry, optics, RAG at 10/15/20/25/30 GB caches) showed consistent double-digit percentage gains, with 30 GB cache + mlock landing at ~7.7 tok/s average vs ~4.9 tok/s without.

## Design notes

- **Env-gated**: default off, zero behavioral change for existing users
- **Name-based classification**: matches tensor names by substring, so it works across all quants of Flash and PRO without needing a type-based classifier that would have to be updated per quant recipe
- **Page-aligned ranges**: rounds `abs_offset..abs_offset+bytes` to page boundaries so mlock accepts the request
- **Graceful failure**: mlock failures are counted, reported, and don't abort — process keeps running even if the user's wired-memory ceiling is too low
- **Reports byte totals** on stderr for observability, matching the format of the existing streaming-cache mlock summary

## Test plan

- [x] Builds clean on macOS Metal (`make`)
- [x] `DS4_MLOCK_NONROUTED=0` (default): no behavior change, no log line
- [x] `DS4_MLOCK_NONROUTED=1`: logs `locked X GiB, failed 0 GiB, skipped routed Y GiB`, verified `X + Y == total tensor bytes` (80.76 GiB for Q2 Flash) — no unclassified tensors
- [x] `vmmap -summary` confirms the non-routed portion moves from evictable to wired
- [x] End-to-end inference produces identical output vs baseline (sanity: `ds4-agent --non-interactive -p ...` returns same greedy tokens)
- [x] Verified on Q2 Flash (`DeepSeek-V4-Flash-IQ2XXS-…-imatrix.gguf`); classification is name-based so Q4 Flash and Pro variants should work identically

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -2445,6 +2445,69 @@ static void parse_tensors(ds4_model *m, ds4_cursor *c) {
     }
 }
 
+/* Env-gated: wire the mmap ranges of every non-routed-expert tensor so macOS
+ * cannot page them out under memory pressure. Routed experts still stream
+ * from the SSD-backed cache as usual. Requires DS4_MLOCK_NONROUTED=1. Large
+ * lock ranges typically need root (mlock user limits are tiny by default). */
+static bool tensor_name_is_routed_expert(ds4_str name) {
+    static const char *needles[] = {
+        ".ffn_gate_exps.", ".ffn_up_exps.", ".ffn_down_exps.",
+    };
+    for (size_t i = 0; i < sizeof(needles) / sizeof(needles[0]); i++) {
+        const char *needle = needles[i];
+        size_t nlen = strlen(needle);
+        if (name.len < nlen) continue;
+        for (uint64_t j = 0; j + nlen <= name.len; j++) {
+            if (memcmp(name.ptr + j, needle, nlen) == 0) return true;
+        }
+    }
+    return false;
+}
+
+static void model_mlock_nonrouted(const ds4_model *m) {
+    const char *env = getenv("DS4_MLOCK_NONROUTED");
+    if (!env || env[0] == '\0' || env[0] == '0') return;
+
+    long page_size = sysconf(_SC_PAGESIZE);
+    if (page_size <= 0) page_size = 16384;
+
+    uint64_t requested = 0, locked = 0, failed = 0, skipped = 0;
+    int last_errno = 0;
+
+    for (uint64_t i = 0; i < m->n_tensors; i++) {
+        const ds4_tensor *t = &m->tensors[i];
+        if (tensor_name_is_routed_expert(t->name)) { skipped += t->bytes; continue; }
+        if (t->bytes == 0) continue;
+        if (t->abs_offset > m->size || t->bytes > m->size - t->abs_offset) continue;
+
+        uint64_t start = t->abs_offset;
+        uint64_t end   = t->abs_offset + t->bytes;
+        uint64_t pstart = start & ~((uint64_t)page_size - 1);
+        uint64_t pend   = (end + (uint64_t)page_size - 1) & ~((uint64_t)page_size - 1);
+        if (pend > m->size) pend = m->size;
+        uint64_t len = pend - pstart;
+        requested += len;
+        if (mlock((const uint8_t *)m->map + pstart, (size_t)len) == 0) {
+            locked += len;
+        } else {
+            failed += len;
+            last_errno = errno;
+        }
+    }
+
+    const double gib = 1024.0 * 1024.0 * 1024.0;
+    fprintf(stderr,
+        "ds4: mlock non-routed weights: locked %.2f GiB, failed %.2f GiB, "
+        "skipped routed %.2f GiB\n",
+        (double)locked / gib, (double)failed / gib, (double)skipped / gib);
+    if (failed && last_errno) {
+        fprintf(stderr, "ds4: mlock error (last): %s (try running with sudo, "
+                        "or raise the wired-memory limit)\n",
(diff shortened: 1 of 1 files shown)


OPS-NeoRetro:
@hi-pauls do you have any other Macs to test this on? I don't know about q2-q4-imatrix and q4-imatrix, and DeepSeek V4 Pro would easily run on something like an M3 Ultra 256GB

DanteCpp:
Hi @hi-pauls ,

I think you are trying to optimize a suboptimal configuration to begin with... The plot down below has been taken on a M2 max 32GB of RAM, you can see that up to 15GB of expert cache budget it runs at about 6 t/s, as you keep increasing the cache size the kernel starts evicting the pages that contain the unrouted weights as you reported. 

If you mlock those weights, the system will start swapping other programs or event critical systems components at the limit it can cause OOM errors and crash your system. Therefore I do not think that including the option of mlocking is safe. 

The safest thing to do is just avoid extreme memory pressure, which will result in better performances overall without the risk of OOM.

cache_sweep_baseline

Best,
Dante

hi-pauls:
Hi @DanteCpp,

Sorry for the late reply, I'm traveling.

I don't have other Macs available. I know it's not an ideal configuration, but this Mac is my only option and I'm surprised as to how consistent and usable it can be especially for specialty coding tasks.

I'm aware of the implications, hence it's opt-out. If you over-provision it of course, you can break a few things, but for consistent 7-9t/s performance, leaving about 15-20gb on my 64gb config for system and other programs seems to work just fine. I actually had problems with other programs like Firefox and Chrome being pigs and grabbing memory they don't need, starving the model and degrading performance to 3-4t/s, sometimes lower. For consistent background agent performance (i. e. opencode), I chose to prioritize  consistency and that's what the change is about.

Cheers

Paul

DanteCpp:
Hi @hi-pauls,

the configuration I am talking about is not related to the specific machine. The plot I have attached above was computed on a M2 max with 32GB of RAM. However with a M1 max 64GB the same apply. 

For example with my M1 max 64GB, I consistently get 11/12 t/s and an hit rate of about 93% with an expert budget of 32 GB. 

My point is that you can get better performances simply reducing the budget allocated to the expert's cache. Because you are operating in a suboptimal configuration (last point of the plot 16GB), where the t/s drop massively instead of growing monolithically how the trend would suggest.  

In other words I think you should simply use a smaller cache budget to get better performances overall.

Best,
Dante

hi-pauls:
Hi Dante,
interesting, I actually tested cache sizes and found that increasing expert cache and making sure it's pinned did improve performance consistently. The thing I noticed was, that if I don't use mlock, the operating system overhead killed the model performance. That may hit at 16gb expert cache on your system, on mine the maximum lockable 33gb are still in an uptrend (sorry, no fancy graphs), if cache is pinned, with no perceived performance degradation of other processes. If I don't mlock, model performance does degrade however. I'll see if I can grab some graphs.
Cheers
Paul

OPS-NeoRetro:
@alexziskind1 please test this runtime with and without this PR and with and without #555 and/or #491 on all of your Apple Silicon Macs this weekend or next week? This needs to have results across all of Apple's lines to give us something so this PR can be merged. I'll do a code review tomorrow. @DanteCpp and @hi-pauls have been quiet in this PR for days now.

OPS-NeoRetro:
> @alexziskind1 please test this runtime with and without this PR and with and without #555 and/or #491 on all of your Apple Silicon Macs this weekend or next week? This needs to have results across all of Apple's lines to give us something so this PR can be merged. I'll do a code review tomorrow. @DanteCpp and @hi-pauls have been quiet in this PR for days now.

Please make it into a video about DS4, a new DeepSeek V4 runtime, now also supporting GLM 5.2.

GiorgioOppo:
Hi @OPS-NeoRetro, 
were you able to test PR #491 on your hardware?

OPS-NeoRetro:
@GiorgioOppo no, I don't have the hardware to do that. I don't even have a Mac, while I know of a YouTuber (@alexziskind1) who has a ton of Apple Silicon Macs, Strix Halo PCs and DGX Sparks.

GiorgioOppo:
@OPS-NeoRetro, thanks anyway.

PR #482, opened 2026-06-30
DSpark B2 rejection sampling + adaptive block sizing

machiabeli:
## Summary

Adds lossless B2 rejection sampling (Chen/Leviathan 2023) to the DSpark speculative decode path, plus adaptive block sizing for workload-aware performance.

Builds on PR #480 (@audreyt) which provides the DSpark loader, Metal drafter, converter, and test infrastructure.

## Verified speedups (M5 Max 128GB, Q2 base + Q4K drafter)

| Prompt | Baseline | DSpark | Speedup |
|---|---|---|---|
| Repetitive JSON (30x) | 38.72 | **52.81** | **+36.4%** |
| Diverse JSON (20 items) | 38.85 | **42.29** | **+8.9%** |
| Markdown tables | 42.54 | **45.21** | **+6.3%** |
| Thinking: prime checking | 14.71 | **17.61** | **+19.7%** |

All outputs are **token-identical** to non-speculative greedy decode at temp=0. Lossless by construction at any temperature.

## What's in this PR

1. **B2 rejection sampling** (`b2_rejection_sample`, 170 lines) — log-space stable for 129K vocab. Accept with probability min(1, target/draft), sample correction from residual on reject. Activated via `DS4_SPEC_TEMP=0.6` env var; greedy path unchanged when unset.

2. **Off-by-one fix** — `metal_graph_verify_suffix_tops` row[i] predicts drafts[i+1], not drafts[i]. B2 target logits shifted by prepending `s->logits` as row 0.

3. **RNG state persistence** — moved from stack-local (reseeded per call from `time(NULL)`) to `ds4_session` struct. Prevents correlated random sequences at >1 tok/s.

4. **Adaptive block sizing** (`DS4_DSPARK_ADAPTIVE=1`) — starts conservative (block=2), escalates to full block after full commits, drops back on partial commits.

## Key findings

- Speedup scales with draft acceptance: repetitive structure = near-100% full commits = big wins
- Partial-commit replay (~26ms/token) is the bottleneck on creative prompts (see #468 analysis)
- Thinking mode with systematic reasoning (+19.7%) benefits because the checking pattern repeats
- Fix for creative prompts: multi-point checkpointing in `metal_graph_verify_suffix_tops` (#468)

## Credits

- @audreyt — PR #480 loader/converter/Metal drafter plumbing
- @lobanov — root-cause analysis (ffn_gate_inp F32/F16), numpy oracle, Q4_K GGUF

[...]

PR #480, opened 2026-06-30
Add DSpark speculative draft runtime

audreyt:
## Status

Ready for review as an **opt-in experimental DSpark/DeepSpec path**, not a guaranteed speedup claim.

Latest push (`4c0209d`) adds prefix checkpoints for DSpark partial accepts: accepted prefixes now commit from captured verifier state instead of replaying target tokens, with restore/replay fallback if the prefix path fails. This keeps the correctness-gated verifier contract and removes an avoidable runtime cost on partial commits.

After opening the PR I rechecked the temp-0 code-generation output difference using ds4's existing speculative-decode correctness invariant rather than strict token-stream equality. Raw greedy text can diverge on near-greedy ties; the invariant used by `test_mtp_worst_argmax_gap` is that every committed speculative token remains the target model's argmax or near-argmax when replayed through plain decode.

Current evidence:

- Fresh `make test` on `4c0209d`: `ds4 tests: ok`.
- Fresh DSpark asset-backed check with `/Users/au/w/ds4/ds4flash.gguf` and `/Users/au/w/ds4/gguf/deepseek4.dspark.gguf`: `mtp-verify-depth nspec=206 max_chunk=6 worst_argmax_gap=0.000 at=-1` and `dspark-speculative-block nspec=96 max_chunk=6 worst_argmax_gap=0.000 at=-1`.
- Partial DSpark accepts exercised the new prefix path with `noreplay=1`; observed prefix commit/refresh cost was about 0.8-1.1 ms on those cycles, while verifier time remained about 69-74 ms.
- The exact code-generation prompt that first showed non-identical text: first raw-text divergence accepted `.` where non-spec greedy selected `,`, with target logit gap `0.024364`; worst gap across the 128 committed tokens was `0.859341 at=81`, still under the existing `<= 2.0` threshold.
- Therefore the non-identical raw text is a tie/near-tie cascade, not by itself a correctness blocker.

Performance remains workload- and draft-quality-dependent. #468 already includes @lobanov's DSpark Metal drafter / CPU-oracle work and measurements showing that this needs careful benchmarking against real acceptance rates. This PR should be reviewed as loader/runtime/converter plumbing plus a correctness-gated Metal DSpark implementation, not as proof of a universal decode-speed win.

## Summary

Adds an opt-in DSpark/DeepSpec speculative draft runtime for DeepSeek V4 Flash, addressing #468.

- detects and binds official DSpark Markov draft GGUF metadata/tensors
[...]

Issue #468, opened 2026-06-27
DeepSeek-V4-Flash-DSpark


lobanov:
DeepSeek released [DeepSeek-V4-Flash-DSpark variant](https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash-DSpark) of the V4 Flash model with built-in speculative decoding and described the methodology in a [companion paper](https://github.com/deepseek-ai/DeepSpec/blob/main/DSpark_paper.pdf). Given the focus on performance of running the model locally, I think it makes sense to explore how DSpark can be integrated into ds4.

It seems that DSpark provides considerable improvement in tok/s performance with fixed concurrency over previous generation of speculative decoding (MTP):

<img width="801" height="317" alt="Image" src="https://github.com/user-attachments/assets/8fc963c0-a8ef-4ca4-812f-382378dc686d" />

lobanov:
Initial quick benchmarking shows that it might be harder to achieve than it seems on the face of it with a fully-local engine.

Generation baseline on M5 Max on short-ish prompts is 28 ms per token. With current 2-token MTP enabled it's 2ms for prediction and 33ms for fast verification pass. That in practice only gains on average 6-7% because predictions are not always correct. When `--quality` is enabled, a different slower exact verifier kernel is used, which verifies two tokens in 67ms. In practice this gives 28% *reduction* of inference speed over the baseline. Under realistic assumptions of prediction quality and assuming exact verification cost scales linearly with number of tokens there is no gain.

Three avenues to explore:
1) Can exact verifier be made to scale sublinearly with increased number of predicted tokens? Very hard, especially on Metal. It's a research project.
2) Use a different approach for batch verification altogether, e.g. distribution-exact not logit-exact. Research project again, but perhaps easier.
3) Let go of exact verification and use existing batched verifier. It's algorithmically identical, but numerically different, so model precision can suffer. By how much is hard to tell, it would need benchmarking. This is more practical, because we are accepting some quality loss with quantization anyway, and this may allow to trade speed for quality.

daaain:
This sounds incredible! So you already have a working implementation @lobanov?

I've been discussing it with Claude after cloning both repos and sounds like batching could be a way to go (there's a bit at the end in response to your comment):

---

# DSpark on ds4 (DwarfStar): feasibility, tensor map & RAM budget

**Status:** investigation only, no code written yet.
[...]

PR #502, opened 2026-07-05
Add DeepSpec tests and training harnesses

stephenlthorn:
## Summary

This reconciles the two open DSpark PRs from issue #468 into one branch that builds cleanly and passes the model-free test suite:

- #480 (@audreyt) - DSpark Metal runtime, converter, prefix-checkpoint partial commits
- #482 (@machiabeli) - B2 rejection sampling (lossless temp>0) + adaptive block sizing (`DS4_DSPARK_ADAPTIVE`)

The two PRs share an earlier common scaffold but diverged afterward, so this isn't a plain merge - I cherry-picked both commit chains onto `main` (skipping #482's skip-logits-replay commit + its own revert, which cancel out to a no-op) and hand-resolved one real conflict.

## Fix included

The conflict was in the partial-commit path of the speculative loop: audreyt's prefix-checkpoint fast-commit (skip replay when the batch verify already captured a KV row for each committed position) and machiabeli's B2 correction token (sampled fresh during rejection, never processed by the verify pass) interact badly - a correction token has no captured KV row, so routing it through the fast-commit path would silently commit against a stale/wrong cache row.

Fixed by gating the fast-commit path on `commit_has_correction`: any commit that includes a B2 correction token always takes the frontier-restore + replay path, same as before #480's optimization existed. Also fixed a small leak of `b2_target_logits` on the prefix-commit early return.

## Testing

On Apple Silicon (Metal), no GPU-heavy model runs yet on this machine:

- `make` (Metal): clean, 0 warnings
- `make cpu`: clean
- `./ds4_test --dspark-binder --dspark-markov-bf16 --dspark-runtime --server`: all pass
- Model-dependent tests (`--dspark-speculative-block`, `--mtp-verify-depth`) self-skip without `DS4_TEST_DSPARK`/`DS4_TEST_MTP` - not yet run against a real DSpark GGUF

I plan to run the real acceptance/speed benchmarks from #468 (block=2 vs block=5 vs adaptive, JSON/code/creative prompt mix) on a machine with the model loaded and will report back with numbers. Flagging this now as a reviewable integration point rather than waiting, since #480 and #482 are both stalled on being combined.

Thanks to @audreyt and @machiabeli for the actual implementation work - this PR is integration/reconciliation only, no new algorithmic contribution beyond the correction-token fix above.


stephenlthorn:
## On-hardware validation (M4 Max 128GB, q2-imatrix base + Q4_K DSpark drafter)

### Correctness: PASS

| Test | Result | Detail |
|---|---|---|
| `--mtp-verify-depth` | OK | nspec=206 max_chunk=5 worst_argmax_gap=0.000 at=-1 |
| `--dspark-speculative-block` | OK | nspec=96 max_chunk=6 worst_argmax_gap=0.000 at=-1 |

The `worst_argmax_gap` invariant holds at 0.000 on both: every committed speculative token is exactly the target model's argmax on replay.

### Generation tok/s by prompt type (`--nothink -n 128`)

| Prompt | baseline | legacy_mtp | dspark_b2 | dspark_adaptive | dspark_b5 | Best DSpark vs baseline |
|---|---|---|---|---|---|---|
| json_repetitive | 28.40 | 31.44 | 31.54 | 37.44 | **38.10** | **+34.2%** (b5) |
| md_table | 31.53 | 31.57 | 30.55 | **33.75** | 31.91 | **+7.0%** (adaptive) |
| code_gen | 31.35 | 31.42 | 30.16 | 28.81 | 30.00 | -3.6% to 0% (wash) |
| prose | 31.31 | 31.29 | 26.81 | 21.68 | 21.21 | -31% to -32% (regression) |

This confirms @lobanov's original acceptance-rate problem (issue #468: "average acceptance 1.5-2 tokens, 5% slowdown") is resolved by this branch's combination of the prefix-checkpoint path (#480) and the B2 correction fix (#482) - we're seeing +34% on repetitive JSON with the same Q4_K drafter class on the same q2 base quant.

### temp=0 identicality vs baseline

json/table/code: byte-identical across all 5 configs. **Prose diverges** on all 3 DSpark configs (legacy_mtp stays identical). Root cause: a near-tied target logit gets flipped by the drafter's slight numeric bias, and greedy decoding cascades the divergence from there. The `worst_argmax_gap` invariant still holds (every committed token is still argmax-correct against the target model at the moment it's verified) - so this is "target-argmax-replay lossless," not "byte-identical," on ties. Worth calling out explicitly in the README's DSpark section since it currently reads like byte-identical is the guarantee.

### Findings

- **block=5 regresses on prose** (-32%): low acceptance means more rejected drafts to verify plus partial-commit replay cost. block=2 nearly halves the loss (-14%) but still trails baseline.
- **Adaptive escalates to b5 under low acceptance and pays the same penalty as fixed b5 on prose.** A ceiling (e.g. cap escalation at b3 unless recent acceptance is consistently full) would likely recover most of the structured-output win without the prose regression - flagging as a possible follow-up, happy to try it if useful.
- Prefill throughput is ~3% lower across the board with DSpark loaded (extra resident weights + verify overhead) - minor, not blocking.
- Single-sample-per-cell on an M4 Max; didn't control for thermal throttling. The b5 vs b2 prose gap (-32% vs -14%) should be re-run 2-3x before treating it as a stable number rather than noise.
- `DS4_SPEC_TEMP` (B2 stochastic path) not yet exercised - all runs above are temp=0. That's a separate correctness surface (rejection sampling at temp>0) worth a dedicated follow-up pass.

Raw logs/output and a re-runnable bench script are on hand if useful for someone else to reproduce.

stephenlthorn:
Pushed a follow-up commit addressing the two items from the benchmark comment above:

1. **Staged adaptive escalation** (2 → 3 → full, requiring 2 consecutive full commits before using the full block, instead of jumping straight from 2 to full on the first full commit). This targets the prose regression: previously one incidental full commit at block=2 would escalate to block=5, and the very next low-acceptance cycle paid the full block=5 partial-commit replay cost before dropping back down.
2. **README clarification**: documented that DSpark's temp=0 guarantee is "every committed token is the target model's argmax at verify time," not byte-identical output to a non-speculative run — near-tied logits can flip which side of a tie gets committed under speculation, which is expected, not a bug.

Not yet re-benchmarked on real hardware — the person with the GPU/model loaded is queued to re-run the prose case (and ideally 2-3x to control for the thermal noise flagged earlier) against this staged version. Will report back with numbers once that's done.

stephenlthorn:
## Re-benchmark: staged escalation does not fix the prose regression

Follow-up to the earlier validation comment. Re-ran only the prose case (3x, to rule out thermal noise) against the staged-escalation commit (7ad4ea2).

| | tok/s | vs baseline |
|---|---|---|
| baseline | 31.31 | — |
| adaptive, single-jump (2→full) | 21.68 | -30.8% |
| adaptive, staged (2→3→full, this commit) | 22.47 (mean of 22.61/22.47/22.32, spread ~1.3%) | -28.2% |

**Correctness unaffected:** `--dspark-speculative-block` still reports `worst_argmax_gap=0.000`.

### Diagnosis: this was the wrong fix for the problem

The +0.8 t/s improvement is inside measurement noise, not a real fix. Staging the escalation only slows down *how fast* adaptive reaches block=5, it doesn't change the fact that block=5 (or even block=2) costs more than it saves once you're there:

- Prose drafter acceptance at temp=0 is below the break-even point identified in #482's writeup (~26ms partial-commit replay penalty per rejected token) *even at block=2*. The staged guard ("2 consecutive full commits before escalating") essentially never fires on prose, so this version just spends longer at block=2 before giving up and trying block=5 anyway - which is why it loses slightly less, not because escalation got smarter.
- 31.31 t/s baseline is effectively "DSpark off." On this prompt class, drafter verify overhead exceeds what accepted tokens save, independent of block size.

### Recommendation

Don't present staged escalation as a prose fix - it isn't one, though it's still worth keeping as protection against future flaky single-commit escalation. Two actual paths forward, and I'd defer to whoever picks this up next on which:

1. **Scope adaptive down**: cap `DS4_DSPARK_ADAPTIVE` at block=3 by default instead of escalating to the full block, and describe it in docs as "block=2..3," not "block=2..5." This trades away the structured-output ceiling (+34% on repetitive JSON was measured at fixed block=5, not adaptive) for a policy that can't regress prose as hard.
2. **Accept the workload-dependence and make it explicit**: keep block=5 available but not the adaptive default, and document that DSpark is a net win only on structured/repetitive output at this drafter quality - creative text should run with `--mtp-draft` unset (or explicitly `2`) until a higher-quality drafter (Q8, ~9GB extra resident) is validated to raise prose acceptance above break-even.

Not picking one of these myself since it's a product tradeoff (peak structured-output speedup vs. worst-case prose safety) rather than a bug fix. Happy to implement whichever is preferred.

ds4_cuda.cu
```diff
@@ -8917,6 +8917,25 @@ extern "C" int ds4_gpu_attention_decode_raw_batch_heads_tensor(
...
+        uint32_t                raw_cap,
+        uint32_t                raw_start,
+        uint32_t                n_head,
+        uint32_t                head_dim) {
+    (void)heads; (void)model_map; (void)model_size; (void)sinks_offset;
+    (void)q; (void)raw_kv; (void)n_tokens; (void)n_raw; (void)raw_cap;
+    (void)raw_start; (void)n_head; (void)head_dim;
+    return 0;
+}
+
```
OPS-NeoRetro: Why did you write a stub?

rocm/ds4_rocm_attention_launch.cuh
```diff
@@ -324,6 +324,25 @@ extern "C" int ds4_gpu_attention_decode_raw_batch_heads_tensor(
...
+        uint32_t                raw_cap,
+        uint32_t                raw_start,
+        uint32_t                n_head,
+        uint32_t                head_dim) {
+    (void)heads; (void)model_map; (void)model_size; (void)sinks_offset;
+    (void)q; (void)raw_kv; (void)n_tokens; (void)n_raw; (void)raw_cap;
+    (void)raw_start; (void)n_head; (void)head_dim;
+    return 0;
+}
+
```
OPS-NeoRetro: Why did you also write a stub here?

gguf-tools/deepseek4-quantize.c
```diff
@@ -1646,15 +2236,87 @@ static void write_full_gguf(st_db *db, const gguf_file *tmpl, const output_conte
...
+static void free_gguf_file(gguf_file *g);
+
+static uint64_t count_gguf_kv_prefix_in_file(const char *path, const char *prefix) {
+    FILE *fp = fopen(path, "rb");
+    if (!fp) die_errno("open GGUF", path);
+    char magic[4];
+    if (fread(magic, 1, sizeof(magic), fp) != sizeof(magic) || memcmp(magic, "GGUF", 4) != 0) {
+        die("bad GGUF self-test file");
+    }
+    (void)read_u32_le_fp(fp, "GGUF version");
```
OPS-NeoRetro: Don't forget to add a GGUF version check in deepseek4-quantize

ds4_dspark_runtime.c
```diff
@@ -0,0 +1,70 @@
...
+        return "DSpark draft graph has not been validated on real DSpark GGUF weights; "
+               "speculative decode stays off (no fake draft tokens)";
+    case DS4_DSPARK_SPEC_DSPARK_NONSEQ_NOT_READY:
+        return "DSpark nonseq draft head has not been validated on real trained DSpark GGUF weights; "
+               "speculative decode stays off (no fake draft tokens)";
+    case DS4_DSPARK_SPEC_DISABLED:
+    default:
+        return "speculative draft disabled";
+    }
+}
```
OPS-NeoRetro, line 70: Don't forget to end your files with a newline

ds4_dspark_runtime.h
```diff
@@ -0,0 +1,34 @@
...
+
+
+
+ds4_dspark_spec_gate ds4_dspark_speculative_gate(ds4_mtp_draft_kind kind,
+                                                 bool mtp_ready,
+                                                 int mtp_draft_tokens);
+
+const char *ds4_dspark_spec_gate_reason(ds4_dspark_spec_gate gate);
+
+#endif
```
OPS-NeoRetro, line 34: Don't forget to end your new files with a newline

ds4.c
```diff
@@ -38,6 +39,11 @@
 
 #include "ds4.h"
 #include "ds4_distributed.h"
+#include "ds4_dspark_runtime.h"
+
+#ifndef DS4_GIT_COMMIT
+#define DS4_GIT_COMMIT "unknown"
+#endif
```
OPS-NeoRetro: Good addition to DS4

ds4.c
```diff
@@ -25140,6 +26850,315 @@ int ds4_engine_collect_imatrix(ds4_engine *e,
...
+    (void)e;
+    (void)dataset_path;
+    (void)output_dir;
+    (void)target_model_name_or_path;
+    (void)chat_template;
+    (void)ctx_size;
+    (void)max_prompts;
+    (void)max_tokens;
+    fprintf(stderr, "ds4: DSpark target cache export requires a graph backend build\n");
+    return 1;
```
OPS-NeoRetro: Please add a CPU path

OPS-NeoRetro (changes requested):
If you have the time, please implement DSpark for CUDA, ROCm and the CPU debug path. This is a Metal-only PR. You also did some improvements to DS4, but for this PR, this is a good start for you. Please rename this to "Metal: Reconcile DSpark runtime (#480) + B2 rejection sampling/adaptive block sizing (#482)".

gsrunion:
**ROCm/gfx1151 port + result: DSpark runs correctly on ROCm but is net-negative on bandwidth-bound Strix Halo — the batched verify doesn't pay off here.**

Ported the block-draft path to the ROCm backend and ran it on DeepSeek-V4-Flash IQ2XXS (81GB) on a Strix Halo box (Radeon 8060S, 128GB, ROCm 7.2.4). Drafter converted from `deepseek-ai/DeepSeek-V4-Flash-DSpark` shards 46–48 via `deepseek4-quantize --dspark-only` → 11.5GB Markov DSpark GGUF (`kind=dspark, draft=5, runtime_mtp=yes`).

**Port** (happy to open as a PR if useful): the only backend gap was `ds4_gpu_attention_decode_raw_batch_heads_noncausal_tensor` (stubbed on both CUDA and ROCm). Implemented it on ROCm by adding a `noncausal` flag to `attention_decode_mixed_kernel` (extends the existing `single_all` all-visible mode → every query in the block sees all raw KV, `visible_comp=n_comp`), threaded through `attention_decode_batch_launch`, and forced the general kernel for the noncausal path (the head8-online fast paths bake causal bounds). Plus the `ds4_mtp_draft_runtime_supported` gate, which hard-rejects the dspark family off Metal.

**Correctness: good.** Temp-0 output byte-identical to plain; acceptance healthy (`drafted=5 committed=5` on repetitive spans). So the kernel is right.

**Performance: net-negative on this hardware.** Agent/code-edit prompt, temp 0, matched:

| config | decode t/s |
|---|---|
| plain (no spec) | 15.3 |
| **DSpark** (draft 5 / 8 / adaptive) | **11.6–12.3** |
| prompt-lookup (#396) | 18.7 |

Prose is worse (7–9 t/s). Draft-depth sweep (repetitive prompt): verify ≈ 253 ms @draft2, 330 ms @draft5, 329 ms @draft8 — **the block verify of 5–8 tokens costs ~5× a single-token decode (~65 ms), while committed plateaus at ~4.4 tokens regardless of depth.** So we pay ~5 decodes' worth of bandwidth to commit ~4.4 tokens ⇒ slower than plain.

**Why (hypothesis):** on UMA Strix Halo, batch-1 decode is memory-bandwidth-bound, and the routed MoE experts don't amortize across a verify block — each of the 5 block tokens routes to its own 6/256 experts, so expert bytes read scale ~linearly with block size even in a batched verify (only the dense/attention Q8 stream, ~6.5 GB, amortizes). Once dense is amortized, experts dominate and the block verify approaches the sequential cost. DSpark's 60–85% figures are on compute-bound datacenter GPUs where batched verify is nearly free; that condition doesn't hold on a 212 GB/s iGPU. (Consistent with the per-kernel bandwidth accounting I posted in #16.)

**Takeaway for the ROCm/UMA target:** prompt-lookup (#396) beats DSpark here because it has zero drafter cost and near-100% acceptance on echo-heavy work. DSpark's ceiling on this class of hardware is bounded by expert-gather bandwidth in the verify, not by the drafter or the port. Full logs / the noncausal patch / conversion recipe available if you want them — and this box (your QA hardware twin) is free nights US-Eastern for any block-verify batching experiments.

stephenlthorn:
@OPS-NeoRetro Renamed as suggested - it is indeed a Metal-only PR, and the title now says so.

On implementing CUDA/ROCm/CPU here, I'd argue against expanding this PR's scope, and @gsrunion's excellent data two comments up is most of the reason:

- **ROCm**: the port already exists (gsrunion's), it's correct, and it's **net-negative on Strix Halo** - the project's primary ROCm target - because routed-MoE expert reads don't amortize across the verify block on bandwidth-bound UMA hardware. Enabling the runtime there today would ship a slowdown. Their noncausal-kernel patch has standalone value and deserves to be its own PR (with the runtime gate kept off, or opt-in) rather than folded into this one.
- **CUDA**: the primary CUDA target is DGX Spark / GB10, which is also unified-memory at ~273 GB/s - the same bandwidth-bound regime as Strix Halo, so the same economics very likely apply. I also can't test CUDA changes from this machine, and per AGENT.md/CONTRIBUTING.md untested backend changes shouldn't ship.
- **CPU**: per AGENT.md the CPU path is reference/debug only (and large CPU runs can trigger a macOS kernel VM bug), so a speculative-decode fast path there has no production value.

The current state - Metal runtime + explicit `ds4_mtp_draft_runtime_supported` gate with non-Metal backends loading DSpark GGUFs inertly, documented in the README - keeps every backend correct while limiting the speedup claim to the one backend where it's been validated. That gating pattern came from #480 and I think it's the right call until someone demonstrates a backend where the verify economics work (which per gsrunion is a batching/bandwidth question, not a porting question).

@gsrunion this is a great writeup - the expert-gather-doesn't-amortize analysis cleanly explains why the datacenter DSpark numbers don't transfer to UMA. Please do open the noncausal kernel port as its own PR; correctness-verified backend parity is valuable even with the runtime gated off, and it gives anyone experimenting on Strix Halo a working baseline. The prompt-lookup comparison (#396 winning on echo-heavy work with zero drafter cost) is also a useful datapoint for the docs.

gsrunion:
Thanks — agreed on all three points, and keeping #502 Metal-only with the #480 gate is the right call. The verify-economics ceiling is a batching/bandwidth question, not a porting one, so gating off-by-default until someone demonstrates a backend where a block verify amortizes is exactly right.

I'll open the noncausal ROCm kernel as its own PR as you suggested. It fills the ROCm stub of `ds4_gpu_attention_decode_raw_batch_heads_noncausal_tensor` with a real implementation (a `noncausal` flag on `attention_decode_mixed_kernel`, extending the existing `single_all` all-visible path, threaded through `attention_decode_batch_launch`), keeps the runtime gated off by default with a `DS4_DSPARK_ROCM=1` opt-in for experimenters, and is validated on gfx1151 (Radeon 8060S / Strix Halo): temp-0 output byte-identical to plain decode, `drafted=5 committed=5` acceptance.

Since it depends on this PR's declaration + Metal impl + ROCm stub, I'll base it so it's a clean 3-file delta once #502 lands (rather than stacking a noisy diff now). Happy to keep the `DS4_DSPARK_ROCM` opt-in or drop to pure gated-off parity — whichever you prefer.

Deviad:
Hi all, I was wondering if this would ever work on GLM 5.2. 
I did lots of tests but best performance was achieve with the built-in MTP and patching llama. 
I got 24 token/s with short context and around 18 token/s with 18k context.

OPS-NeoRetro:
> Hi all, I was wondering if this would ever work on GLM 5.2. I did lots of tests but best performance was achieve with the built-in MTP and patching llama. I got 24 token/s with short context and around 18 token/s with 18k context.

@Deviad, what workload did you run at 18K context?

Deviad:
> > Hi all, I was wondering if this would ever work on GLM 5.2. I did lots of tests but best performance was achieve with the built-in MTP and patching llama. I got 24 token/s with short context and around 18 token/s with 18k context.
> 
> @Deviad, what workload did you run at 18K context?

Hi, Programming tasks as well as telling the model to say whether a certain variable was contained in a long text.

OPS-NeoRetro:
> > > Hi all, I was wondering if this would ever work on GLM 5.2. I did lots of tests but best performance was achieve with the built-in MTP and patching llama. I got 24 token/s with short context and around 18 token/s with 18k context.
> > 
> > 
> > @Deviad, what workload did you run at 18K context?
> 
> Hi, Programming tasks as well as telling the model to say whether a certain variable was contained in a long text.

Hey, it's been a long time @deviad, but is DSpark working on both main branch and this PR, on DeepSeek V4 Flash, DeepSeek V4 Pro and GLM 5.2?

Deviad:
> Hey, it's been a long time @Deviad, but is DSpark working on both main branch and this PR, on DeepSeek V4 Flash, DeepSeek V4 Pro and GLM 5.2?

What I am trying to say is that on M3 Ultra 512 GB, if you run DSpark as a sidecar it will take much more time to perform which will make useless the gain you can obtain. So far the only way I could increase performance considerably with M3 Ultra when playing around with MTP was to set gamma=1 and use MTP but not as an external sidecar.

OPS-NeoRetro:
@stephenlthorn, please rebase now

stephenlthorn:
@OPS-NeoRetro Rebased onto current `main` - the branch was 188 commits behind. It now merges cleanly.

I have to flag something important though: **most of this PR is now superseded, and the rebase removed it.**

Since this branch was opened, DSpark landed on `main` independently (`fc9efd1`, 2026-07-18) and was built out considerably: `769a8ba` (exact stochastic speculative decoding), `84cc882` (ROCm), `8d683d6` (opportunistic temperature sampling), `74e8f11`/`7fb2830` (partial accepts), plus a full scheduler with break-even accounting. `main`'s DSpark carries ~50 `DS4_DSPARK_*` tuning vars against this branch's 6.

Where the two designs collided, I resolved every conflict in favour of `main`, because in each case `main`'s version is strictly more capable. What that dropped:

- **Metal DSpark runtime** - `main`'s `support_kind`/`DS4_SUPPORT_DSPARK` design replaces this branch's `mtp_weights.kind` approach entirely; they're mutually exclusive.
- **B2 rejection sampling** - superseded by `769a8ba`'s exact stochastic decoding.
- **Adaptive block sizing** - was embedded in a 377-line drafter loop that `main` replaced. Worth noting this was the piece I [benchmarked at a -28% prose regression](https://github.com/antirez/ds4/pull/502#issuecomment-4887209153) and concluded staged escalation didn't fix, so I don't think it's a loss.
- **Noncausal attention kernel** - `main` renamed and hardened it as `ds4_gpu_attention_noncausal_raw_batch_heads_tensor` (with the ROCm implementation @gsrunion contributed separately).
- **Quantizer `--dspark-only`** - superseded by `--dspark-support` / `--dspark-manifest`.

Four of the ten commits became empty and were dropped. What remains (+1153/−52):

- `gguf-tools/deepspec/ds4_deepspec.py` - 505-line DeepSpec training script, fully additive, nothing equivalent on `main`
- `ds4_dspark_runtime.{c,h}` - model-free helpers (runtime gate, bf16 conversion, prefix-slot math)
- `tests/ds4_test.c` - +387 lines of unit tests for the above
- `ds4.h`/`ds4.c` - draft-kind taxonomy and `ds4_mtp_draft_runtime_supported`

**Two caveats:**

1. I had to drop the DeepSpec C collector (`--dspark-target-cache-*` and `ds4_engine_collect_dspark_target_cache`). It depends on Metal graph fields (`batch_cur_hc`, `prefill_tokens`) that `main`'s graph doesn't have - porting it is re-engineering rather than rebasing. I removed its CLI flags and help text too rather than advertise options that don't work. The Python side is untouched.
2. The intermediate commits aren't individually buildable - `make` doesn't cover `gguf-tools`, so some merge artifacts in `deepseek4-quantize.c` survived a few commits before I caught them. The final tree is clean; the branch isn't bisectable.

Verified on Apple Silicon (Metal): `make` and `gguf-tools` both build with **0 errors and 0 warnings**, matching `main` exactly. `4/4` core tests pass, plus eval and agent self-tests. `ds4_test` needs a GGUF this machine doesn't have - `main` fails there identically, so it's environmental.

Given what's left is a helper library, tests, and a training script, **this no longer matches the PR title.** I'd suggest either closing it as superseded, or letting me retitle and reduce it to "DeepSpec training scaffold + DSpark draft-kind helpers" - whichever you and @antirez prefer. Happy to open the DeepSpec piece as its own focused PR instead.

OPS-NeoRetro:
@stephenlthorn, this should be renamed to "Add DeepSpec tests and training harnesses", because from what I have seen, @antirez kind of reimplement PRs in this project, which is kind of a bad thing since PRs don't really get merged and instead become discussion threads. I don't totally know what @antirez is working on right now.

stephenlthorn:
Renamed to "Add DeepSpec tests and training harnesses" — that matches what's actually in scope here now that DSpark itself was implemented separately by @antirez. Thanks for flagging.

PR #504, opened 2026-07-05
CUDA streaming: per-size-class expert caches, keep cache warm across prefills

### LINKED ISSUE #503: CUDA ssd-streaming: expert cache thrashes on mixed-quant GGUFs (decode ~0.9 t/s); prefill drops the whole cache every request
## Environment

- GPU: NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation 96GB (sm_120)
- Driver 595.71.05, CUDA runtime 13.2, toolkit 13.1 (`make cuda-generic`)
- Host: EPYC 9124, 180 GiB RAM (VM, GPU passthrough), Ubuntu 26.04
- ds4 commit `80ebbc3`
- Model: official `DeepSeek-V4-Flash-Layers37-42Q4KExperts-OtherExpertLayersIQ2XXSGateUp-Q2KDown-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix-fixed.gguf` (q2-q4 mixed, 91GB)

Command:

```bash
DS4_CUDA_NO_DIRECT_IO=1 DS4_CUDA_KEEP_MODEL_PAGES=1 ./ds4-server \
  -m <mixed.gguf> --ctx 32768 --host 0.0.0.0 --prefill-chunk 2048 \
  --ssd-streaming --ssd-streaming-cache-experts 4500 \
  --kv-disk-dir /data/ds4-kv --kv-disk-space-mb 16384
```

## Symptom

Decode is stuck at **~0.9 t/s** (never ramps up) and stderr floods with thousands of
alternating cap notices, two sizes flip-flopping forever:

```
ds4: CUDA streaming expert cache capped from 10619 to 4973 experts (available 81.57 GiB, reserve 16.00 GiB, 13.50 MiB/expert)
ds4: CUDA streaming expert cache capped from 10619 to 9947 experts (available 81.57 GiB, reserve 16.00 GiB, 6.75 MiB/expert)
ds4: CUDA streaming expert cache capped from 10619 to 4973 experts (available 81.57 GiB, reserve 16.00 GiB, 13.50 MiB/expert)
... (repeats for the whole run)
```

A 200-word completion does not finish within 400s. The same setup with the uniform
q2-imatrix or q4-imatrix GGUFs works fine.

## Root cause 1 — single-size expert cache released at every size transition

`ds4_cuda.cu` keeps ONE global `g_stream_expert_cache` keyed on the last-seen
`(gate_expert_bytes, down_expert_bytes)`:

- `cuda_stream_expert_cache_note_size()` resets the runtime cap whenever the size
  differs from the previous call;
- `cuda_stream_expert_cache_prepare()` calls `release_all()` when dims differ from
  the cached ones.

The mixed GGUF interleaves layers with 6.75 MiB and 13.50 MiB experts. The slab/bypass
logic in `ds4.c` (`weights_streaming_layer_experts_uniform`) covers the graph path, but
the CUDA cache layer still receives both sizes (e.g. via
`ds4_gpu_stream_expert_cache_budget_for_expert_size` and the seed/compact-load paths),
so the cache is torn down and rebuilt at every q2↔q4 layer transition — it never
accumulates a single hit.

## Root cause 2 — prefill drops the whole resident cache on every request

`ds4_gpu_stream_expert_cache_prepare_selected_batch` →
`cuda_stream_selected_cache_begin_compact_load(..., allow_global_cache=0)` executes:

```c
if (!allow_global_cache) {
    cuda_stream_expert_cache_release_all();
}
```

unconditionally — i.e. **every prompt releases the entire warm expert cache**, even for
a 20-token chat prompt whose staging buffers are ~100 MB. Every request then re-warms
the cache from zero during decode (visible as a chunk-rate ramp 1.7 → 11 t/s on every
single request). This affects uniform models too, not just mixed ones.

## Fix (see PR)

1. Expert caches per size class (small fixed array keyed by `(gate,down)` bytes) —
   removes the cross-class teardown entirely;
2. `begin_compact_load`: try the staging allocation first, release the expert cache
   only if VRAM is actually short (then retry once);
3. Let the prefill batch path read the global cache (hits become device-to-device
   copies) with an append-but-never-evict policy, so a long prompt cannot cycle out
   the decode working set.

## Results (same GPU, same mixed GGUF)

| build | 200-word completion |
|---|---|
| stock `80ebbc3` | does not finish in 400s (~0.9 t/s, log flood) |
| patched, `--ssd-streaming-cache-experts 7500`, `DS4_CUDA_STREAMING_EXPERT_CACHE_RESERVE_GB=8` | ~12.6 t/s end-to-end, `DS4_CUDA_STREAMING_EXPERT_CACHE_VERBOSE` shows **97% hit rate** on the 2nd request, no cap-flood |
| patched, `--ssd-streaming-cache-experts 9472` (= all uniform-class experts resident, 62 GiB), reserve 8 | **~15 t/s end-to-end** after convergence |

Caveat found while tuning: the per-class budget can overcommit — a count budget of N
lets EVERY class try N slots, and the lazily-growing model-range arena (q8_0 chunks,
~1.8 GiB each) can then fail mid-decode (`cuda decode failed`). A shared byte budget
across classes (or reserving the arena's projected size upfront) would make sizing
safer; for now the practical rule on 96GB with this GGUF is: budget = uniform-class
expert count (9472), reserve ≥ 8 GB.

Uniform q2/q4 GGUFs behave as before (q2 fully-resident path untouched).

Remaining ceiling on mixed is the per-token device-to-device compact-buffer copies
(~2 GB/token even at 97% hit rate) — that seems related to the direction discussed in
#491 (serving experts directly from cache slots), out of scope here.

iCreil:
Fixes #503.

See the issue for the full analysis and benchmarks. Summary of the three changes, all in `ds4_cuda.cu`:

**1. Per-size-class expert caches.** `g_stream_expert_cache` becomes a fixed array of 4 caches keyed by `(gate_expert_bytes, down_expert_bytes)`. `note_size()` is replaced by `class_index()` (assigns a free slot on first sight; recycles class 0 in the — never observed — case of more than 4 distinct sizes). Runtime caps and cap-notice dedup move to per-class arrays. `prepare()` operates on the class cache; cross-class transitions no longer release anything.

**2. Conditional release in `begin_compact_load`.** The unconditional `release_all()` on the `allow_global_cache=0` path becomes: try `ensure_slots()` first; on allocation failure release the expert caches and retry once. Long-prompt VRAM pressure is still handled, short prompts keep the warm cache.

**3. Prefill batch reads the cache.** `prepare_selected_batch` now passes `allow_global_cache=1` plus a new `allow_cache_evict=0` mode: batch loads copy hits device-to-device and may append to free capacity, but never evict valid slots (an LRU cache smaller than a long prompt's expert union would otherwise be fully cycled, destroying the decode working set). The single-token decode path keeps full LRU behaviour (`allow_cache_evict=1`).

Notes:
- The `memset` over the struct containing a `std::vector` is replaced by explicit member resets (`vector().swap()`), which also fixes a latent leak of the slots buffer.
- Budget semantics: a count budget now applies per size class, so classes can overcommit in aggregate; the lazily-growing model-range arena can then fail mid-decode. A shared byte budget across classes would be safer — left out of scope here to keep the diff focused, happy to follow up. Practical guidance for the 96GB + Flash q2-q4 case is in the issue (budget 9472, reserve ≥ 12 GB).
- Tested on RTX PRO 6000 Blackwell 96GB / CUDA 13.1 / `make cuda-generic` with Flash q2-imatrix (uniform: behaviour unchanged) and Flash q2-q4-imatrix (fixed: ~0.9 → ~25 t/s sustained decode, 97% cache hit rate per `DS4_CUDA_STREAMING_EXPERT_CACHE_VERBOSE`).

### DIFF
--- a/ds4_cuda.cu
+++ b/ds4_cuda.cu
@@ -192,12 +192,19 @@ static std::unordered_map<uint64_t, size_t> g_q8_f16_by_offset;
 static std::vector<cuda_q8_f32_range> g_q8_f32_ranges;
 static std::unordered_map<uint64_t, size_t> g_q8_f32_by_offset;
 static cuda_stream_selected_cache g_stream_selected_cache;
-static cuda_stream_expert_cache g_stream_expert_cache;
+/* Expert caches are kept per (gate,down) byte-size class: mixed-quant models
+ * (e.g. the q2-q4-imatrix Flash GGUF) interleave layers whose routed experts
+ * have different byte sizes. A single cache keyed on the last-seen size gets
+ * released and reallocated on every layer-size transition, so it never
+ * accumulates hits (decode collapses to ~1 t/s with permanent cap-log spam).
+ * Keeping one small cache per size class removes the thrash. */
+#define DS4_CUDA_STREAM_EXPERT_CLASSES 4
+static cuda_stream_expert_cache g_stream_expert_caches[DS4_CUDA_STREAM_EXPERT_CLASSES];
 static uint32_t g_stream_expert_budget_override;
-static uint32_t g_stream_expert_runtime_cap;
-static uint32_t g_stream_expert_memory_cap_notice;
-static uint64_t g_stream_expert_runtime_gate_bytes;
-static uint64_t g_stream_expert_runtime_down_bytes;
+static uint32_t g_stream_expert_runtime_caps[DS4_CUDA_STREAM_EXPERT_CLASSES];
+static uint32_t g_stream_expert_memory_cap_notices[DS4_CUDA_STREAM_EXPERT_CLASSES];
+static uint64_t g_stream_expert_class_gate_bytes[DS4_CUDA_STREAM_EXPERT_CLASSES];
+static uint64_t g_stream_expert_class_down_bytes[DS4_CUDA_STREAM_EXPERT_CLASSES];
 static uint64_t g_model_range_bytes;
 static uint64_t g_q8_f16_bytes;
 static uint64_t g_q8_f32_bytes;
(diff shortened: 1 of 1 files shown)


iCreil:
Some real-workload numbers from running this PR against the mixed-quant file it targets (DeepSeek V4 Flash, 37 routed layers IQ2XXS/Q2K + 6 layers with Q4_K experts, 91 GB GGUF fully page-cache resident; RTX PRO 6000 Blackwell 96 GB, `--ssd-streaming --ssd-streaming-cache-experts 9472`).

Chat-shaped test against `ds4-server` (4 short questions, 160 generated tokens each, 3 rounds, default temperature):

- **without this PR**: 1.36 tok/s decode, flat across all rounds — the cache never warms up, even when the exact same question is repeated;
- **with this PR**: 19.3–19.4 tok/s from the second round on (**~14x**), ~99% cache hit rate.

A timing probe around the selected-expert load shows where the time was going: without the fix each layer's load costs ~10.6 ms, of which only ~1.1 ms is the actual H2D wait — the other ~9.5 ms is host-side `cudaFree`/`cudaMalloc` churn from the `release_all()` on every size-class flip, with a permanent 0% hit rate. With the per-class caches the same load drops to ~0.17 ms/layer.

Happy to share the exact scripts and configs if useful.

iCreil:
Superseded by #605, which rebuilds this on top of the current streaming backend. Closing.

PR #511, opened 2026-07-06
Fix CPU backend abort on DeepSeek V4 PRO: use DS4_N_OUT_GROUP for attention-output grouping

### LINKED ISSUE #510: CPU backend aborts on DeepSeek V4 PRO: grouped Q8_0 tensor has an unexpected layout
### Summary
Running any DeepSeek V4 **PRO** GGUF on the **CPU** backend aborts almost immediately at prefill layer 1 with:

```
ds4: grouped Q8_0 tensor has an unexpected layout
```

**Flash** works fine on the same CPU build; only PRO is affected. The root cause is a hardcoded Flash-specific constant in the CPU attention-output projection.

### Environment
- CPU backend (`make cpu`), x86-64 Linux (Ubuntu 24.04, Xeon w9-3595X, 503 GB RAM)
- Model: `DeepSeek-V4-Pro-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-Instruct-imatrix.gguf` (PRO Q2)
- ds4 @ `main`

### Reproduce
```sh
make cpu
./ds4 -m DeepSeek-V4-Pro-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-Instruct-imatrix.gguf \
      --cpu -c 2048 -n 48 --nothink -p "hello"
# ds4: prefill layer 1/61ds4: grouped Q8_0 tensor has an unexpected layout   (exit 1)
```

### Root cause
`layer_grouped_out_one()`, `layer_grouped_out_one_decode_scratch()`, and `layer_grouped_out_batch()` (`ds4.c`) hardcode the attention-output group count:

```c
const uint32_t n_groups = 8;   // Flash value
```

PRO's shape profile sets `n_out_group = 16` (Flash = 8). With the literal `8`, PRO computes
`group_dim = n_head_dim * (n_head / n_groups) = 512 * (128 / 8) = 8192`, but PRO's `attn_output_a`
tensor has `dim[0] = 512 * (128 / 16) = 4096`, so the `w->dim[0] != group_dim` check in
`matmul_q8_0_grouped_batch()` / `matvec_q8_0_grouped_rows()` fails and `ds4_die()` fires.

The engine already exposes the correct value as the `DS4_N_OUT_GROUP` macro (`= g_ds4_shape.n_out_group`); the CPU path just doesn't use it at these three sites.

### Fix
Replace the three hardcoded `n_groups = 8` with `n_groups = DS4_N_OUT_GROUP`. `rank = 1024` is shared across profiles, so no other change is needed. Flash is unaffected (macro == 8); PRO then prefills all 61 layers and produces coherent output on CPU (~0.95 t/s on this host). PR incoming.

### Impact
Minor — the CPU path is the diagnostics/reference backend — but this also blocks using it as a correctness oracle for PRO. GPU/Metal backends are unaffected.

mrighini:
## Fix CPU backend abort on DeepSeek V4 PRO

The CPU reference path aborts at prefill layer 1 on any DeepSeek V4 **PRO** model with
`grouped Q8_0 tensor has an unexpected layout`. **Flash is unaffected.**

### Root cause
`layer_grouped_out_one()`, `layer_grouped_out_one_decode_scratch()`, and
`layer_grouped_out_batch()` hardcode the attention-output group count to `8` (the Flash value).
PRO's shape profile sets `n_out_group = 16`, so with the literal `8` the code computes
`group_dim = n_head_dim * (n_head / 8) = 512 * (128 / 8) = 8192`, while PRO's `attn_output_a`
tensor has `dim[0] = 512 * (128 / 16) = 4096` → the layout check in the grouped-Q8_0 matmul fails
and `ds4_die()` fires.

### Fix
Use the existing `DS4_N_OUT_GROUP` macro (`= g_ds4_shape.n_out_group`) instead of the literal `8`
at the three call sites. `rank = 1024` is shared across profiles, so no other change is needed.
Flash takes the identical path (macro == 8); PRO now prefills all 61 layers.

```diff
-    const uint32_t n_groups = 8;
+    const uint32_t n_groups = DS4_N_OUT_GROUP;
```

### Testing (per CONTRIBUTING.md)
Backend: **CPU** (`make cpu`), x86-64 Linux (Ubuntu 24.04, Xeon w9-3595X).

| Model (Q2 imatrix) | Before | After |
|---|---|---|
| Flash (`DeepSeek-V4-Flash-IQ2XXS…imatrix`) | coherent, ~3 t/s | **unchanged**, coherent, ~3 t/s |
| PRO (`DeepSeek-V4-Pro-IQ2XXS…Instruct-imatrix`) | **abort** at prefill layer 1 | coherent, all 61 layers, ~0.95 t/s |

Command:
```sh
./ds4 -m <model>.gguf --cpu -t 60 -c 2048 -n 48 --nothink -p "In one short paragraph, explain how vaccines work."
```

Note: I validated on the **CPU** backend via manual smoke tests (Flash output byte-path is
unchanged since the macro resolves to 8 for Flash; PRO output is coherent). I did **not** run the
full `ds4_test` regression suite, which builds the default GPU backend (`nvcc`) unavailable on this
CPU-only host — happy to run any specific check you suggest.

Fixes #510

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -7098,7 +7098,7 @@ static void layer_grouped_out_one(
         const ds4_model   * model,
         const ds4_layer_weights * layer,
         const float       * heads) {
-    const uint32_t n_groups = 8;
+    const uint32_t n_groups = DS4_N_OUT_GROUP;
     const uint32_t group_heads = DS4_N_HEAD / n_groups;
     const uint32_t group_dim = DS4_N_HEAD_DIM * group_heads;
     const uint32_t rank = 1024;
@@ -7117,7 +7117,7 @@ static void layer_grouped_out_one_decode_scratch(
         const ds4_layer_weights * layer,
         const float            * heads,
         ds4_cpu_decode_scratch * scratch) {
-    const uint32_t n_groups = 8;
+    const uint32_t n_groups = DS4_N_OUT_GROUP;
     const uint32_t group_heads = DS4_N_HEAD / n_groups;
     const uint32_t group_dim = DS4_N_HEAD_DIM * group_heads;
     const uint32_t rank = 1024;
@@ -7134,7 +7134,7 @@ static void layer_grouped_out_batch(
         const ds4_layer_weights * layer,
         const float       * heads,
         uint32_t            n_tok) {
-    const uint32_t n_groups = 8;
+    const uint32_t n_groups = DS4_N_OUT_GROUP;
     const uint32_t group_heads = DS4_N_HEAD / n_groups;
     const uint32_t group_dim = DS4_N_HEAD_DIM * group_heads;
     const uint32_t rank = 1024;


ukd1:
This is a less complete version of #347

PR #516, opened 2026-07-07
Server: add --parallel N interleaved sessions (default 1)

iCreil:
## What

Add `--parallel N` to ds4-server: N live sessions interleaved on the single graph worker, so several clients can be served concurrently — everyone keeps streaming while another prompt prefills. Default is `--parallel 1`, which keeps today's behavior exactly (verified byte-identical, see Testing).

Stacked on #515 (engine-shared work buffers), which is what makes the second live session affordable in VRAM; the first commit here is that PR.

## Why

The README describes the current model honestly: *"concurrent requests wait their turn on the single live graph/session"*, and sketches a future *"stateful session-based protocol"*. This PR is a bounded, opt-in step in that direction that fits the existing architecture: still one graph worker, still cooperative scheduling, no threads added, no batching — just interleaving at the token/chunk granularity the engine already exposes.

The concrete pain it removes: with one live session, a user who arrives while a long generation is running waits for all of it before seeing a first token. With `--parallel 2`, on our box (RTX PRO 6000 96 GB, V4 Flash q2) a second client got its **first token in 0.7 s** while a 350-token generation was in flight, and that first stream's worst inter-token gap during the newcomer's prefill was **0.15 s**. A third concurrent request queues FIFO as before and completes when a slot frees.

## How

- **Slots.** `struct server_slot` owns one `ds4_session` plus the protocol live state bound to that timeline (Responses/Anthropic/thinking bindings) and a per-slot copy of the kvstore continued-checkpoint frontier, swapped in/out around each turn. `server.active` points at the slot being stepped; per-request code reaches its session through it, which keeps the single-worker ownership model explicit. Parser-side live call-id lookups scan every slot under the existing tool mutex.
- **Phases.** The first commit mechanically splits `generate_job()` into `job_begin` / `job_prefill` / `job_start_decode` / `job_decode_round_init` / `job_decode_step` / `job_finish` with the old stack frame moved into a `gen_state` struct. The old `decode_again:` tool-recovery label becomes a `GEN_STEP_REDECODE` result. With one slot the driver performs exactly the same calls in the same order.
- **Scheduler.** The worker hands queued jobs to idle slots (longest-common-prefix affinity so a shared system prompt stays warm, then LRU) and steps one slot per turn: a phase transition, one prefill slice, or a burst of decode tokens (default 6, `DS4_SERVER_SCHED_DECODE_TOKENS`).
- **Prefill preemption.** Reuses the cooperative cancel callback: a sync is interrupted at a chunk boundary once the slice deadline (default 1500 ms, `DS4_SERVER_SCHED_PREFILL_MS`) passes *and* another job is actually waiting; it resumes from the session checkpoint on the slot's next turn — the same resume path prefill cancellation already uses. With `--parallel 1` the callback never fires.
- **Guardrails.** `--parallel N>1` is refused together with `--ssd-streaming` (interleaving would thrash the streaming expert cache); the flag requires resident experts.

## Testing

On CUDA (RTX PRO 6000 Blackwell 96 GB, DeepSeek V4 Flash q2-imatrix):

- **Golden, default config**: greedy chat completion against the previous build — byte-identical content in both stream and non-stream mode; clean SIGTERM shutdown.
- **Fairness** (`--parallel 2 --ctx 32768 --prefill-chunk 1024`, 94.9 GiB VRAM): client A streaming 350 tokens; client B arrives at +3 s → B's TTFT 0.7 s, A's worst inter-token gap during B's prefill 0.15 s, both complete correctly; a third concurrent request queues and completes.
- `tests/two_sessions_smoke` (from the first commit) passes: interleaved greedy streams match isolated runs token by token.

Not covered yet (called out for review): multi-slot interaction with MTP speculative decode is exercised only lightly (state is fully per-graph, so isolation holds by construction); Metal untested beyond compile paths — no Apple hardware here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -217,6 +217,12 @@ ds4_rocm.o: ds4_rocm.cu ds4_gpu.h ds4_iq2_tables_cuda.inc $(ROCM_SRCS)
 tests/cuda_long_context_smoke: tests/cuda_long_context_smoke.o ds4_cuda.o
 	$(NVCC) $(NVCCFLAGS) -o $@ $^ $(CUDA_LDLIBS)
 
+tests/two_sessions_smoke.o: tests/two_sessions_smoke.c ds4.h
+	$(CC) $(CFLAGS) -I. -c -o $@ tests/two_sessions_smoke.c
+
+tests/two_sessions_smoke: tests/two_sessions_smoke.o $(CORE_OBJS)
+	$(DS4_LINK) -o $@ $^ $(DS4_LINK_LIBS)
+
 ds4_test: ds4_test.o ds4_help.o ds4_kvstore.o rax.o $(CORE_OBJS)
 ifeq ($(UNAME_S),Darwin)
 	$(CC) $(CFLAGS) -o $@ ds4_test.o ds4_help.o ds4_kvstore.o rax.o $(CORE_OBJS) $(METAL_LDLIBS)
--- a/ds4.c
+++ b/ds4.c
@@ -10309,10 +10309,22 @@ static void print_vec_stats(const char *name, const float *x, uint64_t n) {
 
 enum { DS4_STREAMING_PREFILL_CACHE_SEED_MAX_TOKENS = 64 };
 
+/* Shared per-engine work buffers.  Everything in here is written and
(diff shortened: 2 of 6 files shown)


iCreil:
Superseded by #604, closing.

PR #519, opened 2026-07-08
Canonicalize blank reasoning blocks

JordiPosthumus:
## Summary
- render whitespace-only assistant reasoning as an empty reasoning block
- keep non-empty reasoning content unchanged
- add a server unit test covering blank reasoning in tool-context prompt replay

## Why
A replayed assistant turn can carry an empty reasoning block as whitespace, for example `<think> </think>`, while the live KV prefix may contain `<think></think>`. That difference is semantically empty but changes the rendered prompt bytes/tokens and can cause a live KV cache token mismatch.

## Tests
- `make ds4_test && ./ds4_test --server && make ds4-server`

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -2269,6 +2269,19 @@ static bool role_is_user_like(const char *role) {
     return !strcmp(role, "user") || !strcmp(role, "tool") || !strcmp(role, "function");
 }
 
+static bool text_is_ascii_ws_only(const char *s) {
+    if (!s) return true;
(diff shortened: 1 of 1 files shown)


JordiPosthumus:
Closing because this implementation predates the current multi-model prompt-rendering code. The underlying whitespace-only reasoning replay issue may remain, but this diff covers only the older DeepSeek paths and no longer represents a complete current-`main` fix. I’ll reproduce it on current `main` before returning with a focused replacement.

PR #523, opened 2026-07-09
hy3: Metal decode path + shared gate/up fusion

audreyt:
## Summary
- Hy3 (`hy_v3`) single-token Metal decode on residual tensors (`cur_hc` / `after_*_hc`) instead of per-layer bounce through batch buffers.
- Fuse shared-expert gate/up/SwiGLU for `n_tokens==1` (reuse Flash helper); multi-token prefill unchanged.
- Dense layer 0 uses `batch_routed_*` scratch sized for `DS4_N_DENSE_FF` (13312).
- Kill switch: `DS4_METAL_DISABLE_HY3_SINGLE_TOKEN_DECODE=1` restores legacy batch path.

Also on this branch (prior commits): Hy3 CPU/Metal GQA, split-half RoPE, plain residual, Q4 selected-slots for top-8 MoE, multi-token Q4 selected token-loop under streaming.

## Published mixed GGUF (not in git)

Optional test artifact for ~128 GB machines:

[`hy3-heretic-80layer-Layers74-79Q4KExperts-mixed.gguf`](https://huggingface.co/audreyt/Hy3-Heretic-GGUF/blob/main/hy3-heretic-80layer-Layers74-79Q4KExperts-mixed.gguf) — ~89.8 GiB (96,456,123,360 bytes), derived from [`trohrbaugh/Hy3-heretic`](https://huggingface.co/trohrbaugh/Hy3-heretic), a decensored/abliterated Hy3 checkpoint.

Quantization layout:
- Layers 1–73: `IQ2_XXS` gate/up + `Q2_K` down
- Layers 74–79: `Q4_K` routed experts

The runtime changes are architecture-level Hy3 support and are not specific to the Heretic weights.

SHA-256: `2dc5f74b4784b6de374196918510935ddf102cdd72272e6832edd9c3c7c2ef2b`

## Test plan
- [x] `make ds4` clean build
- [x] Exact published GGUF verified against the Hub LFS SHA-256 above
- [x] M5 Max 128 GB smoke: `./ds4 -m …mixed.gguf --metal --ctx 2048 --tokens 32 --temp 0 --nothink -p 'What is 2+2? Answer with only the number.'` → `4`; prefill 20.22 t/s, generation 13.16 t/s
- [x] Hy3 CPU-reference kernel checks: `DS4_TEST_MODEL=…mixed.gguf ./ds4_test --metal-kernels` → `OK` (includes Hy3 GQA attention and split-half RoPE references)
- [x] Full-model short-prompt tensor gate: `DS4_TEST_MODEL=…mixed.gguf DS4_TEST_MPP_EQ_CASE=short_italian_fact,short_code_completion,short_reasoning_plain ./ds4_test --metal-tensor-equivalence` → 3/3 cases with identical top-20 logits, zero RMS/max-absolute drift, and identical greedy tokens
- [x] Optimized rewrite vs legacy Metal batch reference: with the kill-switch variable absent versus present, 32-step top-20 logprob dumps were byte-identical (`SHA-256 811ea7b48d1280ad9b3999c95ad29389d6d312db6148df541973d797076a50a8`); `DS4_METAL_DECODE_STAGE_PROFILE` emitted `part=hy3_decode` only for the optimized run, proving the A/B traversed different branches
- [x] Broader repository quality suite: deterministic `./ds4-eval -m …mixed.gguf --plain --nothink --case-sequence 76,…,92 --tokens 128 --temp 0 --seed 1` completed the full 17-case COMPSEC slice with 13/17 passing; misses were cases 79, 80, 84, and 86. This is a regression signal, not an official benchmark score.

Cross-domain evaluator limitation: selected GPQA, SuperGPQA, and AIME cases repeatedly exhausted 512–2048-token budgets before emitting the required final-answer marker. Those truncated runs are not reported as capability scores; the Hy3 prompt/budget policy still needs calibration before this evaluator can produce a meaningful cross-domain baseline for the checkpoint.


giannisanni:
Great to see a Metal path for Hy3. The CUDA counterpart is now open at #535: GQA single-pass online-softmax attention with per-head QK-RMSNorm and NeoX rotate-half RoPE (matches your split-half reference), full 295B with routed experts SSD-streamed per token on a single RTX 4060 Ti 16GB at ~1.8 t/s. It stacks on the GLM CUDA backend in #517.

One gotcha from our quant work that may be relevant to your mixed GGUF: published imatrices never cover blk.80 (the nextn/MTP layer), since imatrix generation runs normal forwards that skip it. Its experts need an explicit non-imatrix type (we used Q2_K) and nextn.eh_proj Q8_0, or llama-quantize dies at the very last tensors. Also `--tensor-type` patterns are unanchored regex_search, so a pattern like `ffn_gate` silently matches `ffn_gate_exps` too.

PR #524, opened 2026-07-09
server: surface unclosed thinking as reasoning_content, not content

### LINKED ISSUE #509: Non-streaming /v1/chat/completions leaks thinking into content when truncated (finish_reason: length) before </think>
## Summary

In **non-streaming** chat completions, if generation is cut off by `max_tokens`
(i.e. `finish_reason: "length"`) *before the model emits its closing `</think>`*,
the entire partial thinking block is returned as `message.content` and
`message.reasoning_content` is absent/empty.

When the same request is allowed to finish (`finish_reason: "stop"`), the split
is correct: the thinking lands in `reasoning_content` and `content` holds only
the final answer. Streaming is also correct — even a truncated stream emits the
partial thinking as `reasoning_content` deltas and never as `content`.

So the leak is specific to the **blocking (non-streaming) response path when the
`</think>` boundary never arrives**. Any client that sets a modest `max_tokens`
will intermittently receive raw chain-of-thought presented as the assistant's
answer.

## Environment

- ds4 built from `main` (CUDA target), ~2026-07-06
- NVIDIA GB10 (DGX Spark), CUDA 13.0
- Model: `deepseek-v4-flash`, Q2 imatrix quant
- Default thinking mode, no tools in the request

## Reproduction

Same prompt in all three cases; only `max_tokens` / `stream` differ.

### A) Non-streaming, truncated — **BUG**

```bash
curl -s http://127.0.0.1:8000/v1/chat/completions \
  -d '{"model":"deepseek-v4-flash","messages":[{"role":"user","content":"What is 17 * 24?"}],"max_tokens":32,"temperature":0}'
```

Observed:

```
finish_reason:     length
reasoning_content: None
content:           'We need to compute 17 * 24. This is a simple multiplication. 17 * 24 = 17 * (20 + 4) ='
```

The thinking is in `content`; `reasoning_content` is missing.

### B) Non-streaming, allowed to finish — correct

```bash
curl -s http://127.0.0.1:8000/v1/chat/completions \
  -d '{"model":"deepseek-v4-flash","messages":[{"role":"user","content":"What is 17 * 24?"}],"max_tokens":512,"temperature":0}'
```

Observed:

```
finish_reason:     stop
reasoning_content: "We need to compute 17 * 24. ... So answer is 408."
content:           'The product of 17 and 24 is 408.'
```

### C) Streaming, truncated — correct (for contrast)

Same body as (A) plus `"stream": true`. Observed over the SSE stream:

```
finish_reason=length, reasoning_content deltas=31, content deltas=0
```

The partial thinking is correctly labelled `reasoning_content`, even though the
`</think>` boundary is never reached.

## Expected

Non-streaming responses should classify tokens the same way streaming does. When
a response is truncated mid-thought:

- the partial thinking should be returned in `reasoning_content`, and
- `content` should be empty (there is no answer yet),

so that `finish_reason: "length"` responses are consistent with both the
completed non-streaming case (B) and the streaming case (C).

## Probable cause

The behaviour is consistent with the blocking JSON path buffering the full
generation and then splitting on the `</think>` terminator to populate
`reasoning_content` vs `content`. If generation stops before `</think>` is
emitted, there is no terminator to split on, so the whole buffer falls through
to `content`.

This is the non-streaming analogue of #13 (*"/v1/chat/completions streams
thinking into content instead of reasoning_content when no tools are
requested"*). #13 fixed the **streaming** path by routing non-tool chats through
the structured streamer that already handles thinking; the **blocking** path
appears to still lack the equivalent "unterminated thinking → reasoning_content"
handling.

## Relationship to existing issues

- **#13** (closed) — streaming variant of this leak; fixed. Streaming now
  behaves correctly (case C confirms). This report is the non-streaming
  counterpart that the #13 fix did not cover.
- **#318** (open) — the tool-call variant of a missing/unclosed `</think>`
  boundary (server drops the DSML call, `finish=stop`, nothing useful returned).
  Same underlying theme: partial/unterminated thinking is not handled
  gracefully. A shared fix for "treat an unterminated thinking block as
  reasoning" would likely address both.
- **#76** (closed) — also involves `length`-stopped responses, but concerns
  benchmark wall-time (a post-generation checkpoint rebuild), not the
  reasoning/content classification. Mentioned only to disambiguate.

## Impact

- `deepseek-v4-flash` emits substantial thinking even for trivial prompts, so
  the truncation window is easy to hit with ordinary `max_tokens` values.
- Downstream proxies/clients that rely on the `reasoning_content` field to
  separate thinking from the answer (rather than parsing `<think>` tags out of
  `content`) will surface raw chain-of-thought as the answer whenever a
  non-streaming response is length-capped.
- The behaviour is inconsistent across three otherwise-equivalent code paths
  (streaming vs blocking, truncated vs completed), which makes it a subtle
  correctness trap.

## Suggested fix

In the blocking response builder, when the generation ends without a `</think>`
terminator (e.g. `finish_reason == "length"` while still inside the thinking
block), treat the accumulated text as `reasoning_content` and leave `content`
empty — mirroring the streaming classifier and the #13 fix.
</content>

stefanwichmann:
Fixes #509.

## Problem
When thinking mode is on and generation stops before `</think>` (e.g. the response is truncated at `max_tokens` mid-thought), the non-streaming path in `parse_generated_message_ex` routed the partial chain-of-thought into `content` (via `split_reasoning_content`) with an empty `reasoning_content`. Clients received raw, unfinished CoT as if it were the assistant's final answer. The live streaming classifier already handles this correctly — only the blocking path was affected.

## Fix
In the `require_thinking_closed && !think_end` branch, treat the accumulated text as unfinished reasoning: surface it as `reasoning_content` and leave `content` empty (stripping a leading `<think>`). Any DSML in that text is unclosed reasoning too, so it's deliberately not executed as a tool call — composing with the existing unclosed-think handling from #318.

## Testing
- Adds `test_thinking_unclosed_is_reasoning_not_content`.
- `./ds4_test --server` passes.
- Built with `make cuda-spark` and validated live on a DGX Spark (GB10): truncated non-streaming responses now return empty `content` + populated `reasoning_content`; completed responses unchanged.

Single file, +35/−3.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -4460,9 +4460,16 @@ static bool parse_generated_message_ex(const char *text, bool require_thinking_c
     if (require_thinking_closed) {
         const char *think_end = find_last_substr(text, "</think>");
         if (!think_end) {
-            /* Model did not close thinking, ignore any DSML in reasoning */
-            fprintf(stderr, "ds4-server: thinking not closed, ignoring DSML in reasoning\n");
-            split_reasoning_content(text, strlen(text), content_out, reasoning_out);
+            /* Thinking mode is on but the model never emitted </think> —
+             * typically the response was truncated at max_tokens mid-thought
+             * (see #509). The accumulated text is unfinished reasoning, not a
+             * final answer: surface it as reasoning_content and leave content
+             * empty, matching what the live streaming classifier already does.
+             * Any DSML here is unclosed reasoning too, so it is deliberately
+             * not executed as a tool call. */
+            size_t think_off = !strncmp(text, "<think>", 7) ? 7 : 0;
+            *reasoning_out = xstrdup(text + think_off);
+            *content_out = xstrdup("");
             return true;
         }
         tool_search = think_end + 8;
@@ -13603,6 +13610,30 @@ static void test_thinking_dsml_after_think_close_is_executable(void) {
     tool_calls_free(&calls);
 }
 
+static void test_thinking_unclosed_is_reasoning_not_content(void) {
+    /* Regression for #509: when thinking mode is on and the model is truncated
+     * (e.g. at max_tokens) before it emits </think>, the partial output is
+     * unfinished reasoning. It must surface as reasoning_content with empty
+     * content, not leak into content as if it were the final answer. */
+    const char *generated =
+        "<think>We need to compute 17 * 24. This is a simple multiplication. "
+        "17 * 24 = 17 * (20 + 4) =";
+
+    char *content = NULL;
+    char *reasoning = NULL;
+    tool_calls calls = {0};
+    TEST_ASSERT(parse_generated_message_ex(generated, true,
+                                           &content, &reasoning, &calls));
+    TEST_ASSERT(calls.len == 0);
+    TEST_ASSERT(reasoning && strstr(reasoning, "We need to compute 17 * 24") != NULL);
+    TEST_ASSERT(strstr(reasoning, "<think>") == NULL);
+    TEST_ASSERT(content && content[0] == '\0');
+
+    free(content);
+    free(reasoning);
+    tool_calls_free(&calls);
+}
+
 static void test_tool_checkpoint_suffix_is_future_prompt_canonical(void) {
     tool_schema_orders orders = make_bash_order();
     const char *tool_schemas =
@@ -15803,6 +15834,7 @@ static void ds4_server_unit_tests_run(void) {
     test_invalid_dsml_tool_error_suffix_includes_system_prompt();
     test_thinking_dsml_is_not_executable_before_think_close();
     test_thinking_dsml_after_think_close_is_executable();
+    test_thinking_unclosed_is_reasoning_not_content();
     test_tool_checkpoint_suffix_is_future_prompt_canonical();
     test_tool_checkpoint_minifies_json_parameters();
     test_tool_memory_replays_sampled_dsml();


Smallfu666:
Independent DGX Spark validation for this PR.

Test environment:

- NVIDIA DGX Spark / GB10 (`sm_121`)
- Ubuntu 24.04.4 LTS, ARM64
- Kernel `6.17.0-1026-nvidia`
- CUDA 13.0
- NVIDIA driver 580.159.03
- DeepSeek V4 Flash Q2 imatrix
- Baseline: `main` at `80ebbc396aee40eedc1d829222f3362d10fa4c6c`
- PR head: `76573b091887d18d2aa3066f7d6da1593191764b`

Baseline reproduction:

- A truncated non-streaming thinking response returned `finish_reason: "length"`.
- Partial reasoning was exposed through `message.content`.
- `reasoning_content` was empty or absent.

Validation results with this PR:

- `make cuda-spark`: PASS
- `./ds4_test --server`: PASS
- Truncated non-streaming thinking response: PASS — `content` was empty and partial reasoning was returned through `reasoning_content`.
- Completed non-streaming thinking response: PASS — final `content` and `reasoning_content` remained correctly separated.
- Truncated streaming response: PASS — no reasoning leaked into `content`.
- No-thinking non-streaming response: PASS — normal assistant output remained in `content`, without unexpected `reasoning_content`.
- Unclosed thinking containing DSML-like tool-call text: PASS — no tool call was executed.

The baseline issue was reproducible on this machine, and the tested PR head corrected it without a regression in these five API paths.

PR #525, opened 2026-07-09
Add GPU kernel tests and fix ROCm batched normalization

daniel-vainsencher:
Not ready for review: this PR should be considered after PR #526 is merged, as it depends on the rocm-test-target for testing.

## Overview

This PR adds some more GPU kernel tests (tested only on ROCm) and fixes a critical correctness bug in the ROCm backend for multi-token prefill.

**Depends on:** PR #526 (rocm-test-target) - adds Makefile support for ROCm test builds

### 1. Add GPU kernel tests (commit 2b13d2e)
Adds 6 GPU kernel tests to validate correctness of:
- Per-head RMS normalization (attention Q projection)
- HC weighted sum (hierarchical context reduction)  
- FP8 KV cache quantization (DeepSeek V4 compressed KV)
- RoPE with YaRN corrections (rotary position embeddings)
- Token embedding with HC replication
- **Batched RMS normalization for multi-token prefill**

Each test compares GPU output against a CPU reference implementation with tight relative error bounds (1e-5 for most ops, 1e-3 for RoPE due to sin/cos precision differences).

The `test_gpu_rms_norm_plain_rows` test specifically validates that multi-row RMS normalization normalizes each row independently with 64 rows (typical prefill batch size). This test catches the bug fixed in commit 2.

### 2. Fix ROCm batched normalization bug (commit 657007c)
**Bug:** In `ds4_gpu_hc_split_weighted_sum_norm_tensor`, when n_rows > 1 (prefill with multiple tokens), the fallback path was calling `ds4_gpu_rms_norm_weight_tensor` which only normalizes the first row.

**Fix:** Changed to `ds4_gpu_rms_norm_weight_rows_tensor` which normalizes all rows independently.

**Impact:**
- Prefill with multiple tokens would produce incorrect normalization for all tokens except the first
- This would cascade through the rest of the layer, producing incorrect model outputs
- Decode (single token) was unaffected

## Testing

**Machine:** ROCm on Radeon 8060S Graphics (gfx1151)  
**Model:** ds4flash.gguf

```bash
$ make rocm-test

$ ./ds4_test --server
server: OK

$ ./ds4_test --metal-kernels  
metal-kernels: OK
```

## Correctness Verification

The new GPU kernel tests validate that the fix works correctly. The batched normalization test ensures each row is normalized independently, catching this class of bug in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,6 +113,13 @@ strix-halo:
 
 rocm: strix-halo
 
+rocm-test:
+	$(MAKE) test \
+		CORE_OBJS="ds4.o ds4_distributed.o ds4_ssd.o ds4_rocm.o" \
+		CFLAGS="$(CFLAGS) -DDS4_ROCM_BUILD" \
+		DS4_LINK="$(HIPCC) $(ROCM_CFLAGS)" \
+		DS4_LINK_LIBS="$(ROCM_LDLIBS)"
+
 ds4: ds4_cli.o ds4_help.o linenoise.o $(CORE_OBJS)
 	$(DS4_LINK) -o $@ $^ $(DS4_LINK_LIBS)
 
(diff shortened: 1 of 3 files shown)


daniel-vainsencher:
## Test Results

All tests pass except for a **pre-existing failure** in `--logprob-vectors`:

```bash
$ ./ds4_test --logprob-vectors
logprob-vectors:
ds4-test: vector short_code_completion step 1 selected token mismatch
tests/ds4_test.c:1472: assertion failed: false
logprob-vectors: ERR
ds4 tests: 1 failure(s)
```

**Verification:** This same failure exists on main branch (80ebbc3) and on rocm-test-target (54470de), confirming it's a pre-existing issue unrelated to this PR's changes.

Passing tests:
- `--server`: ✅ OK
- `--metal-kernels`: ✅ OK (includes the new GPU kernel tests)

The GPU tests added in this PR validate the ROCm normalization fix and all pass successfully.

PR #530, opened 2026-07-10
Batched decode: co-decode N live sessions in one GPU pass (opt-in)

iCreil:
> **Draft, stacked on #516** (which is stacked on #515). Opened as draft so the
> branch is visible while #515/#516 await review — please look at those first:
> this branch builds on the `--parallel N` interleaved-session server from #516.
> The diff below shows those commits too until #516 lands; once it merges I'll
> rebase and only the 14 batched-decode commits remain, then mark this ready.

## What

With `--parallel N` (from #516) the server round-robins several live sessions
through the single graph worker, one token at a time (interleaving). This PR adds
an **opt-in batched decode**: when two or more slots are simultaneously in plain
decode, their one-token-per-session steps are committed **together** in a single
`ds4_session_eval_multi` call, so the memory-bound dense stages (attention
projections, shared expert, output head, RoPE/norm) read each weight block from
VRAM **once** and reuse it across all the sessions' tokens, instead of re-reading
it per session.

Default behaviour is unchanged: batching is gated on `DS4_SERVER_BATCHED_DECODE=1`,
requires `--parallel > 1` and an engine that reports
`ds4_engine_supports_batched_decode()` (CUDA today). With the flag off the server
uses the existing interleaving path bit-for-bit.

## Why it's worth it — the narrow-batch matmul

The dense stages already had "batch" kernels (used for prefill, `n_tokens`
32–2048). Reusing them for `n_tokens` 2–4 was actually **slower** than
interleaving (B=2 ≈ 0.62×): the batch-warp q8 matmul puts `n_tok` on the grid Y
axis, so every token re-reads the weight rows from VRAM — no bandwidth saving in a
memory-bound decode.

The key commit here is a **narrow-batch q8_0 matmul kernel**
(`matmul_q8_0_preq_narrow_warp8`): one warp per output row, it reads each 32-weight
block once into registers and dots it against **all** `n_tok` activation blocks
(weights stay hot in L1, activations are tiny). That is exactly the reuse the
memory-bound decode wants.

### Measured (RTX PRO 6000 Blackwell, DeepSeek V4 Flash q2, isolated decode step)

`ms/step` is the time to advance **all** the batched sessions by one token, so
the throughput comparison is `ms/step` against `n_tok × single-token time`:

| batch | ms/step | speedup vs interleaving |
|------:|--------:|:-----------------------:|
| 1 (single) | 21.1 (1 tok) | 1.00× |
| 2 | 28.7 (2 tok) | **1.47×** |
| 3 | — | 1.67× |
| 4 | — | 1.83× |

(narrow kernel; the prior batch-warp path gave B=2 ≈ 67.5 ms/step = 0.62×.)
End-to-end (with prefill, two concurrent chats, temp 0): **~1.30×** aggregate.

## Design note — dense stages batched, routed experts per sequence

The HC pre/norm, router, shared expert and output head run batched
(`n_tokens = n_seqs`): same weights, every sequence. The **routed experts run
one sequence at a time** through the existing single-token decode MoE kernel on
batch-scratch row views: at batch 2-4 the selected experts of different
sequences are almost always distinct, so a generic batched expert path reads
*more* expert weights than `n_seqs` single passes, not fewer. This was measured,
not assumed — an earlier revision routing the MoE through the prefill expert
path was **0.62-0.80×** (slower than interleaving); the per-sequence expert path
plus the narrow matmul is what flips it to 1.47×.

## Correctness

Batched decode is **not bitwise-identical** to single decode by design: the MoE
down-projection sums the selected experts with `atomicAdd` (reduction order
depends on occupancy), and the hyper-connection fusion / matmul reductions differ
between the single and batch paths. It **is** token-identical in practice: greedy
runs match single decode across the test prompts. `tests/batched_decode_smoke`
checks the narrow path against the already-validated batch-warp path (first-step
logits bitwise-identical, token streams match), which is the right invariant.

## Scheduler & observability

- The scheduler batches only slots in *plain* decode. A slot whose step wants the
  MTP speculative path (tool-call payloads force greedy mid-stream, so eligibility
  flips per token) advances inline with its own burst; batched decode and
  MTP-greedy are mutually exclusive per step.
- `DS4_SERVER_BATCH_DIAG=1` reports the batched fraction and a per-cause breakdown
  of the lone-decode turns (`peer_prefill` / `peer_finish` / `no_peer`). Useful
  finding: on sporadic interactive chat the batched fraction is **arrival-bound**,
  not scheduler-bound — ~90–95% when two streams co-arrive, ~20–30% when arrivals
  are staggered or lengths differ (one stream decodes before the other arrives or
  after it finishes: `no_peer`). Batching pays most under sustained, synchronized
  concurrency; it never hurts the interleaving path since it is opt-in.
- On realistic mixed chat traffic ~47% of decode turns actually co-decode
  (staggered arrivals, different lengths), rising to ~62% for near-identical
  concurrent requests — so the end-to-end gain scales with real overlap, and
  everything else falls back to the #516 interleaving behaviour.

## Known limitation

The multi-sequence decode attention kernel currently handles the raw-SWA and
`f32`-compressed paths; the compressed-**f16** / large-context *online* variant
(where `n_comp` exceeds the score buffer, roughly past ~31k tokens) is not
implemented for the multi path yet. `ds4_gpu_attention_decode_multi_supported()`
gates on this, so those turns cleanly fall back to per-slot single decode —
correct, just without the batch speedup. Adding the f16/online multi attention
kernel is the natural follow-up.

`--ssd-streaming` is excluded by the `ds4_engine_supports_batched_decode` gate.
That combination is actually where batching should pay the most (amortizing the
expert-miss H2D traffic across sequences), but it needs the streaming cache to
accept a per-batch union of selected experts on the decode path — it deserves
its own PR.

## Commit walk-through (14 commits on top of #516)

1. `CUDA: add multi-sequence decode attention kernel` — per-seq KV via seqview.
2. `CUDA: add pos-vector RoPE tail variants` — B sessions have arbitrary positions.
3. `Refactor: extract per-token compressor/indexer step` — reused per-seq (no behaviour change).
4. `Graph: add batched-decode encoder (one token per sequence)` — dense stages batched, per-seq KV/attention.
5. `Session: add ds4_session_eval_multi` — public API + `ds4_engine_supports_batched_decode`.
6. `Server: batched decode turns` — opt-in scheduler batching.
7. `Graph: run batched-decode routed experts per sequence` — MoE via decode selected-expert path.
8. `tests: add batched decode step microbench`.
9. `CUDA: narrow-batch q8_0 matmul for 2-4 token decode` — the performance kernel above.
10. `Server: DS4_SERVER_BATCH_DIAG observability counters`.
11. `Server: attribute single-decode turns by cause in batch-diag`.
12. `test(server): cross-slot KV frontier guard + disconnect harness` — unit test for the
    per-slot KV frontier invariant (extracted `slot_kv_swap_in/out` helpers, no behaviour
    change) + `tests/disconnect_mid_decode_test.py` integration harness.
13. `fix(server-test): route per-session live state through a slot` — repairs the `ds4_test`
    server suite, which no longer compiled after the `--parallel` refactor moved
    `responses_live`/`session` into `server_slot` (pre-existing on #516; happy to fold this
    into #516 instead if you prefer).
14. `test(disconnect): count reasoning_content as decode output` — the harness must count
    thinking tokens too, or the mid-decode abort never triggers on thinking models.

## Flipping the default

With the per-sequence expert path and the narrow matmul in place, the batched
path dominates interleaving at every measured batch size. A reasonable adoption
path: keep it opt-in for one release, then default it on with
`DS4_SERVER_BATCHED_DECODE=0` as the escape hatch. With the flag off (and with
`--parallel 1`) behaviour stays byte-identical to #516.

## Note for maintainers

The `*_multi_tensor` kernels are implemented in `ds4_cuda.cu` only. The Metal
backend will need equivalents (or stubs) for the multi-sequence attention,
pos-vector RoPE and `ds4_session_eval_multi` to build on macOS — happy to add
stubs if you point me at the pattern you prefer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -217,6 +217,12 @@ ds4_rocm.o: ds4_rocm.cu ds4_gpu.h ds4_iq2_tables_cuda.inc $(ROCM_SRCS)
 tests/cuda_long_context_smoke: tests/cuda_long_context_smoke.o ds4_cuda.o
 	$(NVCC) $(NVCCFLAGS) -o $@ $^ $(CUDA_LDLIBS)
 
+tests/two_sessions_smoke.o: tests/two_sessions_smoke.c ds4.h
+	$(CC) $(CFLAGS) -I. -c -o $@ tests/two_sessions_smoke.c
+
+tests/two_sessions_smoke: tests/two_sessions_smoke.o $(CORE_OBJS)
+	$(DS4_LINK) -o $@ $^ $(DS4_LINK_LIBS)
+
 ds4_test: ds4_test.o ds4_help.o ds4_kvstore.o rax.o $(CORE_OBJS)
 ifeq ($(UNAME_S),Darwin)
 	$(CC) $(CFLAGS) -o $@ ds4_test.o ds4_help.o ds4_kvstore.o rax.o $(CORE_OBJS) $(METAL_LDLIBS)
--- a/Makefile
+++ b/Makefile
@@ -223,6 +223,12 @@ tests/two_sessions_smoke.o: tests/two_sessions_smoke.c ds4.h
 tests/two_sessions_smoke: tests/two_sessions_smoke.o $(CORE_OBJS)
 	$(DS4_LINK) -o $@ $^ $(DS4_LINK_LIBS)
 
+tests/attention_multi_smoke.o: tests/attention_multi_smoke.c ds4_gpu.h
+	$(CC) $(CFLAGS) -I. -c -o $@ tests/attention_multi_smoke.c
+
+tests/attention_multi_smoke: tests/attention_multi_smoke.o ds4_cuda.o
+	$(NVCC) $(NVCCFLAGS) -o $@ $^ $(CUDA_LDLIBS)
+
 ds4_test: ds4_test.o ds4_help.o ds4_kvstore.o rax.o $(CORE_OBJS)
 ifeq ($(UNAME_S),Darwin)
 	$(CC) $(CFLAGS) -o $@ ds4_test.o ds4_help.o ds4_kvstore.o rax.o $(CORE_OBJS) $(METAL_LDLIBS)
--- a/Makefile
+++ b/Makefile
@@ -229,6 +229,12 @@ tests/attention_multi_smoke.o: tests/attention_multi_smoke.c ds4_gpu.h
 tests/attention_multi_smoke: tests/attention_multi_smoke.o ds4_cuda.o
 	$(NVCC) $(NVCCFLAGS) -o $@ $^ $(CUDA_LDLIBS)
 
+tests/rope_multi_bench.o: tests/rope_multi_bench.c ds4_gpu.h
+	$(CC) $(CFLAGS) -I. -c -o $@ tests/rope_multi_bench.c
+
+tests/rope_multi_bench: tests/rope_multi_bench.o ds4_cuda.o
+	$(NVCC) $(NVCCFLAGS) -o $@ $^ $(CUDA_LDLIBS)
+
 ds4_test: ds4_test.o ds4_help.o ds4_kvstore.o rax.o $(CORE_OBJS)
 ifeq ($(UNAME_S),Darwin)
 	$(CC) $(CFLAGS) -o $@ ds4_test.o ds4_help.o ds4_kvstore.o rax.o $(CORE_OBJS) $(METAL_LDLIBS)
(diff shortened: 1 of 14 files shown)


iCreil:
You mentioned session batching in your latest video — this stack is a working CUDA implementation of exactly that: opt-in batched decode across `--parallel` slots, 1.47–1.83× aggregate over interleaving at B=2–4, greedy token streams identical to single decode (numbers and design notes in the description).

It's marked draft only because it stacks on #515/#516 — happy to adapt it to whatever direction you had in mind.

iCreil:
Superseded by #604, closing.

PR #561, opened 2026-07-13
Fixes thinking levels for pi.dev setup in README.md

daniel-j-h:
This changeset fixes the pi.dev setup in the readme; deepseek-v4 only has three modes: nothinking, thinking/high, and thinking/max. Also note that switching from/to thinking/max will invalidate the cache because:

> Furthermore, for the "Think Max" mode, we prepend a specific instruction to the beginning of the system prompt to guide the model’s reasoning process, as shown in Table 3.

https://arxiv.org/abs/2606.19348


daniel-j-h:
@antirez good to go here? This fixes the pi setup in the readme as it's broken as is.

Issue #534, opened 2026-07-10
[Discussion] CUDA graphs for the decode path: staged plan + full shim audit


riccardo-galbani:
In the recent video you mentioned CUDA graphs as one of the directions worth
pursuing after the GB10 decode work, since the CUDA backend currently launches
every kernel individually. I'd like to work on this and, given the impact
radius, I'm opening a discussion before writing code, with a staged plan and a
complete audit of the decode path. Line numbers below refer to main @ 80ebbc3.

## Why the seam already exists

The architecture is already designed for batched submission:
`ds4_gpu_begin_commands` / `ds4_gpu_end_commands`. On Metal, begin creates a
command buffer and end commits it — one submission per token. On CUDA, begin
is a no-op stub and end is `cudaDeviceSynchronize()` (ds4_cuda.cu:2479). This
asymmetry is a large part of the Metal-vs-CUDA overhead gap: a decode token is
~350 individual kernel launches at ~5-10µs each, plus the idle gaps between
them.

Because the Metal programming model forbids host readbacks inside a command
buffer, the token body (`metal_graph_encode_token_raw_swa`) is by construction
free of mid-token host decisions, and MoE expert selection already flows
through a device buffer (`selected` tensor in `ds4_gpu_routed_moe_one_tensor`).
Both are exactly the invariants CUDA graph capture needs — the design work is
mostly already done.

Prior art: the Entrpi fork (`decode-perf-tuning` branch) implemented per-layer
graph capture and measured +5-10% decode on GB10, growing with context, with
bit-exact output vs eager. I've studied that implementation in depth and the
plan below builds on its validated patterns (and its documented failure modes).

## Staged plan

Every stage is env-gated, falls back to eager, and must be bit-exact
(md5 of output, graphs ON vs OFF) before moving on. Flag off = byte-identical
behavior to today.

- **Stage 0 — infrastructure, zero behavior change.** A thread-local stream
  override (`ds4_current_stream()` / `ds4_capture_set_stream()`) so that all
  ~100 `ds4_gpu_*` shims can be redirected to a capture stream without
  touching their ABI. Small, self-contained; I'd propose this as the first PR.
- **Stage 1 — dense-FFN layer cluster.** First layers have no router: minimal
  capture, validates the mechanics (warm-up before capture since cudaMalloc is
  forbidden inside one, sync brackets, replay, cache keyed on stable bits).
- **Stage 2 — MoE cluster.** Adds nothing conceptually: expert selection is
  already device-side.
[...]

PR #566, opened 2026-07-15
CUDA: thread-local stream override infrastructure for graph capture (stage 0, no functional change)

riccardo-galbani:
## What

Infrastructure-only groundwork for CUDA graph capture in the decode path
(stage 0 of the plan discussed in #534):

- Thread-local `ds4_current_stream()` accessor, gated behind `DS4_CUDA_GRAPHS=1`
- 175 launch sites redirected through the accessor: 156 kernel launches,
  3 `cudaMemsetAsync`, 9 `cublasSetStream` insertions, 7 profiling
  `cudaEventRecord` (DS4_CUDA_MOE_PROFILE)
- `ds4_capture_set_stream()` setter for later stages (currently unused)
- Two NOTE markers for known stage-2 work items (sync memset at the routed
  MoE counts clear; profiling/capture mutual exclusion)
- No ABI changes: all `ds4_gpu_*` shim signatures untouched

## Why no functional change is guaranteed

No stream is ever created in this commit: `g_ds4_capture_stream` is
initialized to stream 0 and `ds4_capture_set_stream()` has no callers, so
`ds4_current_stream()` returns the legacy stream even with the flag on.
Flag-ON behavior is a no-op **by construction**, not just empirically.

## Verification

Bit-identical logits (`--dump-logits` diff empty) and identical greedy
40-token output, reference (main @ 80ebbc3) vs this branch, flag OFF and ON:

- RTX 4070 Laptop (sm_89), expert-streaming mode
- DGX Spark GB10 (sm_121), fully resident weights (~80.8 GiB device cache),
  with and without MTP (draft=2)

Build: `make cuda CUDA_ARCH=sm_89` and `make cuda-spark CUDA_ARCH=sm_121`, clean.

## What this enables

Later stages (per-cluster capture/replay, following the approach validated
in Entrpi's decode-perf-tuning fork, +5.2–5.6% gen on GB10) can introduce
a real capture stream without touching launch sites again. Each stage will
come as a separate PR.

### DIFF
--- a/ds4_cuda.cu
+++ b/ds4_cuda.cu
@@ -221,6 +221,27 @@ static cudaEvent_t g_stream_selected_stage_event[4];
 static uint64_t g_stream_selected_stage_bytes;
 static cudaStream_t g_stream_selected_upload_stream;
 
+static thread_local cudaStream_t g_ds4_capture_stream = (cudaStream_t)0;
+static int g_ds4_cuda_graphs_gate = -1;
+
+static int ds4_cuda_graphs_gate_enabled(void) {
+    if (g_ds4_cuda_graphs_gate < 0) {
+        const char *v = getenv("DS4_CUDA_GRAPHS");
+        g_ds4_cuda_graphs_gate = (v && v[0] != '\0' && strcmp(v, "0") != 0) ? 1 : 0;
+    }
+    return g_ds4_cuda_graphs_gate;
+}
+
+static cudaStream_t ds4_current_stream(void) {
+    if (!ds4_cuda_graphs_gate_enabled()) return (cudaStream_t)0;
+    return g_ds4_capture_stream;
+}
+
+static void ds4_capture_set_stream(cudaStream_t stream) DS4_CUDA_UNUSED;
+static void ds4_capture_set_stream(cudaStream_t stream) {
+    g_ds4_capture_stream = stream;
+}
+
 static int cuda_ok(cudaError_t err, const char *what);
 static const char *cuda_model_range_ptr_from_fd(
         const void *model_map,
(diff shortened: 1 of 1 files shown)


Smallfu666:
Independent DGX Spark Stage 0 validation.

Test environment:

- NVIDIA DGX Spark / GB10 (`sm_121`)
- Ubuntu 24.04.4 LTS, ARM64
- CUDA 13.0
- NVIDIA driver 580.159.03
- DeepSeek V4 Flash Q2 imatrix
- Baseline: `80ebbc396aee40eedc1d829222f3362d10fa4c6c`
- PR head: `18db29df8aa2fef85a4fb212a1de4666e8aed50e`

I compared the baseline against this PR in the following modes:

- `DS4_CUDA_GRAPHS` unset
- `DS4_CUDA_GRAPHS=0`
- `DS4_CUDA_GRAPHS=1`

Results:

- Repeated greedy stdout output was identical across the baseline and all three PR modes.
- Dumped log-probability JSON was identical across the baseline and all three PR modes.
- Five repeated runs were deterministic.
- CUDA regression, server, long-context, and tool-call-quality tests passed.
- `logprob-vectors` showed the same pre-existing `short_code_completion` `c`/`C` token divergence as the baseline; this PR introduced no new mismatch.
- No NaN, Inf, CUDA error, or new test failure was observed.
- At the 7047-token benchmark point, the measured deltas from the baseline were -0.06% prefill and -0.79% generation. No material Stage 0 performance regression was observed in this check.

Verdict from this independent hardware run: validation passed for the tested Stage 0 behavior. This is validation evidence only, not a maintainer approval or merge recommendation.

PR #578, opened 2026-07-20
metal: restore streamed-expert mlock pinning (fixes #532)

### LINKED ISSUE #532: glm5.2 branch: DeepSeek Flash decode regression (~2.8x) introduced by "Add GLM 5.2 runtime support"
While benchmarking our GLM fork line against main we found that the whole
glm5.2 branch decodes DeepSeek Flash about 2.8x slower than main, and bisected
it to the first commit of the line.

Machine: Apple M5 Pro 64 GB, macOS 26 (Darwin 25.5.0), Metal.
Model: DeepSeek-V4-Flash IQ2XXS (w2Q2K, AProj/SExp/Out Q8, chat-v2 imatrix),
80.8 GiB file, so model >> RAM and --ssd-streaming. Probe:

    ./ds4-bench -m MODEL.gguf --ssd-streaming \
        --ctx-start 2048 --ctx-max 2048 --step-incr 2048 --gen-tokens 32

Single-commit points are ctx 2048 only, rebuilt via make after each
checkout, and their ranges combine two measurement sessions on separate
days; the "main" and "our line" rows are interleaved arms from one session
(A/B/C x2 back to back) swept at ctx 2048-4096.

| commit | gen t/s |
| --- | --- |
| main (5b95fa1) | 6.8-8.3 (4 interleaved runs, ctx 2048-4096) |
| main (80ebbc3, current HEAD) | 6.49 (1 run) |
| 173a9a3 "Add GLM 5.2 runtime support" | 1.94-2.21 |
| 34b3736 (pre pro-correctness merge) | 2.76-2.88 |
| bd89932 (glm5.2 HEAD) | 2.73 (1 run) |
| our GLM branch (bd89932 + our fixes, default env) | 1.9-2.8 (8 interleaved runs, ctx 2048-4096) |

A clean detail for bisecting: 80ebbc3 (current main, fast) is the direct
parent of 173a9a3 (slow), so the whole regression sits inside that one
commit. It persists through the branch HEAD, is untouched by streaming env
configuration (disabling the expert readahead does not move it), and main
is not affected. First token on the glm5.2 line takes ~5.2-7.2 s
(gen_first_ms 5166-7180 across all runs); main's bench build predates the
first-token/steady columns, so for main only whole-generation t/s is
directly comparable, and on the glm5.2 line even the steady phase after
the first token (gen_steady_tps 2.6-6.5) stays below main's whole-run
6.8-8.3.

Prefill shows no clear regression once the page cache is warm (fork line
107-120 vs main 116-118 t/s at these sizes in the warm interleaved reps;
first-rep deltas are cache-warming noise).

We have not root-caused inside 173a9a3; happy to bisect deeper into it or
test patches on this box. Repro scripts available.

gilbert-barajas:
The "Add GLM 5.2 runtime support" refactor dropped all `mlock` pinning of streamed experts — the `mlock` calls, `slab_lock_slot`/`unlock_slot`, and the budget cap all went to zero. Without pinning, in the streaming regime (model ≫ RAM) the OS evicts the expert cache under memory pressure and every token re-streams from disk. That's the DeepSeek-Flash decode regression in #532 (~2–3× slower, ~6s first-token).

This ports the pinning back, adapted to the refactored slab allocator (`slab_backed`/`slab_slot`, 3 gate/up/down buffers per entry): whole-buffer `mlock` at alloc, per-slot lock on fill / unlock under relief, the routed-memory budget cap for graceful degradation, cap-on-failure, and a relief routine that unpins the coldest ~10% of victims under pressure. +382/−1, builds warning-free.

Bench (M5 Max 128 GB, DeepSeek-V4-Flash IQ2XXS, `--simulate-used-memory 72` to force the streaming regime):

| build | gen t/s | first-token |
| --- | --- | --- |
| before (broken) | 0.3–2.5 | ~6.0 s |
| **after (this PR)** | **~7.0** | **~1.7 s** |
| baseline `80ebbc3` | ~6.1 | fast |

Recovers past baseline — the `427e281` kernel work pays off once experts stay resident.

Fixes #532

### DIFF
--- a/ds4_metal.m
+++ b/ds4_metal.m
@@ -622,6 +622,14 @@ static uint32_t g_stream_expert_cache_slab_count;
 static uint32_t g_stream_expert_cache_slab_total_slots;
 static uint32_t g_stream_expert_cache_free_slots[DS4_METAL_STREAM_EXPERT_CACHE_MAX_ENTRIES];
 static uint32_t g_stream_expert_cache_free_slot_count;
+static uint8_t g_stream_expert_cache_slab_slot_locked[DS4_METAL_STREAM_EXPERT_CACHE_MAX_ENTRIES];
+static uint32_t g_stream_expert_cache_mlock_budget_cap;
+static uint8_t g_stream_expert_cache_mlock_relief_applied;
+static uint64_t g_stream_expert_cache_mlock_bytes;
+static uint64_t g_stream_expert_cache_mlock_fail_bytes;
+static uint64_t g_stream_expert_cache_mlock_failures;
+static double g_stream_expert_cache_mlock_ms;
+static int g_stream_expert_cache_mlock_warned;
 static uint64_t g_stream_expert_cache_slab_slot_bytes;
 static uint64_t g_stream_expert_cache_cb_seq;
 static uint64_t g_stream_expert_cache_done_seq;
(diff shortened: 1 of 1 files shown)


thombles:
I can confirm this dramatically improves memory management on the new `ds4f-mxfp4` branch.

PR #582, opened 2026-07-20
server: emit GLM tool-call args with their real JSON types (fixes #569)

### LINKED ISSUE #569: server: GLM tool-call parser stringifies all argument values (numbers/arrays become strings), breaking schema-validating clients
Branch: glm5.2 (bd89932, also present at the current branch tip) Model: GLM 5.2 Component: ds4_server.c, OpenAI-compatible /v1/chat/completions tool calling

Summary
On the glm5.2 branch, the GLM tool-call parser emits every tool argument as a JSON string, regardless of its actual type. Numbers become "10", arrays become "[]", booleans become "true". Any client that validates tool-call arguments against the tool's JSON schema (Chatwise, and generally anything using zod/ajv-style validation) rejects the call.

Reproduction
curl -s http://localhost:18004/v1/chat/completions -H "Content-Type: application/json" -d '{
  "model": "glm-5.2-chat",
  "messages": [{"role":"user","content":"Search the web for ByteDance inference chips, return 10 results"}],
  "tools": [{
    "type": "function",
    "function": {
      "name": "WebSearch",
      "description": "Search the web",
      "parameters": {
        "type": "object",
        "properties": {
          "query": {"type": "string"},
          "max_results": {"type": "number"},
          "exclude_domains": {"type": "array", "items": {"type": "string"}}
        },
        "required": ["query"]
      }
    }
  }]
}'
Actual (note max_results is a string):

"function": {
  "name": "WebSearch",
  "arguments": "{\"query\":\"ByteDance inference chips\",\"max_results\":\"10\"}"
}
Expected:

"arguments": "{\"query\":\"ByteDance inference chips\",\"max_results\":10}"
Client-side symptom (Chatwise, zod validation):

Invalid input for tool WebSearch: Type validation failed:
Value: {"query":"...","exclude_domains":"[]","max_results":"10"}
[{ "expected": "number", "code": "invalid_type", "path": ["max_results"] }]
Root cause
In parse_glm_generated_message_ex (ds4_server.c:5049), every <arg_value> is added with is_string hardcoded to "true":

tool_call_json_args_add(&args, key, value, "true");
GLM's wire format (<arg_key>name</arg_key><arg_value>10</arg_value>) carries no type information — unlike DeepSeek's DSML, which has the string="true|false" attribute that the DSML path already honors. So the GLM path JSON-escapes everything into a string.

Possible fixes
Heuristic (self-contained): try to parse the raw <arg_value> text as a JSON scalar/array/object (number, true/false/null, [...], {...}); if it parses, emit it raw (the existing json_minify_raw_value path, i.e. is_string = "false"), otherwise emit a string. This mirrors what most GLM tool parsers do.
Schema-aware (precise): look up the parameter's declared type in the request's tools[].function.parameters and coerce accordingly (what vLLM's GLM tool parser does). More work, since the server currently doesn't consult client tool schemas at parse time.
Happy to test a patch — I can reproduce this reliably.

gilbert-barajas:
`parse_glm_generated_message_ex` hardcoded `is_string="true"` on every GLM tool-call argument, so numbers/arrays/booleans/null were JSON-escaped into strings (`"10"`, `"[]"`, `"true"`) — which schema-validating clients (zod/ajv, Chatwise) reject. GLM's `<arg_value>` wire format carries no type attribute, unlike DeepSeek's DSML `string="true|false"` that the DSML path already honors via `json_minify_raw_value`.

Adds `glm_arg_value_is_raw_json()` — true only when the whole value is one complete, valid, **non-string** JSON value (number / `true`/`false`/`null` / `[...]` / `{...}`), reusing the `json_skip_value` primitive `json_minify_raw_value` already relies on. Plain text and JSON string literals stay quoted, so `is_string` is `"false"` for raw JSON values and `"true"` otherwise.

Verified with a new parser test (`test_parse_glm_tool_call_arg_types`): the #569 curl's args now emit `max_results: 10`, `exclude_domains: []`, `safe_search: true`, `cursor: null`, `opts: {…}` with correct types, while `query` stays a string. Also fixes an existing test that had encoded the bug (`"timeout": "10"` → `"timeout": 10`). +70/−2.

Fixes #569

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -5690,6 +5690,23 @@ static void trim_const_span(const char **start, const char **end) {
     while (*end > *start && isspace((unsigned char)(*end)[-1])) (*end)--;
 }
 
+/* GLM's <arg_value> wire format carries no type information, so a value such
+ * as 10, true, null, [], or {...} is indistinguishable from plain text on the
+ * wire. Heuristically detect a complete non-string JSON value and emit it raw,
+ * mirroring the DSML string="false" path through json_minify_raw_value; plain
+ * text and JSON string literals stay quoted JSON strings. Returns true only
+ * when the entire value is one valid non-string JSON value, so the raw branch
+ * never emits an unquoted fragment (issue #569). */
+static bool glm_arg_value_is_raw_json(const char *s) {
+    if (!s) return false;
+    const char *p = s;
+    json_ws(&p);
+    if (*p == '"') return false;
+    if (!json_skip_value(&p)) return false;
+    json_ws(&p);
+    return *p == '\0';
+}
+
 static bool parse_glm_generated_message_ex(const char *text,
                                            bool require_thinking_closed,
                                            char **content_out,
@@ -5795,7 +5812,8 @@ static bool parse_glm_generated_message_ex(const char *text,
             }
             char *raw_value = xstrndup(p, (size_t)(value_end - p));
             char *value = dsml_unescape_text(raw_value);
-            tool_call_json_args_add(&args, key, value, "true");
+            const char *is_string = glm_arg_value_is_raw_json(value) ? "false" : "true";
+            tool_call_json_args_add(&args, key, value, is_string);
             free(key);
             free(raw_value);
             free(value);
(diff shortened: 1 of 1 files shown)


garnetlyx:
Verified end-to-end on M3 Ultra Mac Studio (Metal backend, `GLM-5.2-UD-Q2_K_RoutedQ2K.gguf`): applied the helper + call-site + test hunks to my fork, rebuilt, and ran both unit tests and a live OpenAI-compatible request. `test_parse_glm_tool_call_arg_types` passes, and a real `set_todos` call now emits `"todos": [{...}]` as a native array instead of a stringified value. Plain-text string args remain correctly quoted. Thanks for the clean fix.

PR #584, opened 2026-07-21
speed-bench: add Apple M5 Max results

altdotboy:
Adds a `ds4-bench` sweep for the **Apple M5 Max** (128 GiB) to `speed-bench/`.

Run with the canonical command from `speed-bench/README.md`:

```
./ds4-bench \
  -m ds4flash.gguf \
  --prompt-file speed-bench/promessi_sposi.txt \
  --ctx-start 2048 --ctx-max 65536 --step-incr 2048 --gen-tokens 128
```

- Model: DeepSeek-V4-Flash IQ2XXS (0731), Metal backend, full residency (~80.8 GiB).
- Files: `speed-bench/m5_max.csv` (full 32-frontier sweep) and `speed-bench/m5_max_ts.svg`.

Representative points:

| Context | Prefill t/s | Decode t/s (steady) |
|--:|--:|--:|
| 2,048 | 662 | ~40 |
| 32,768 | 534 | 34.9 |
| 65,536 | 484 | 30.7 |

**Toolchain note for anyone re-running this on M5/A19+ hardware:** the Metal 4
Tensor API fast-prefill path (`ds4_metal.m`) is gated behind a *compile-time*
SDK check (`__MAC_OS_X_VERSION_MAX_ALLOWED >= 260000`), not a runtime OS
check. Building with only the Command Line Tools (SDK 15.5) silently compiles
that path out even on macOS 26 hardware that fully supports it — prefill
throughput was roughly halved in that configuration, decode was unaffected.
Point `xcode-select` at a full Xcode install with the macOS 26 SDK
(`sudo xcode-select -s /Applications/Xcode.app/Contents/Developer`) and
rebuild to get the real numbers; that's what these figures reflect
(`tensor_matmul=on` in the startup banner).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #587, opened 2026-07-21
speed-bench: add M5 Max (Q2+Q4) results

thomasjohnflaherty:
Benchmark results for Apple M5 Max (128 GB unified memory), Metal backend, mixed Q2+Q4 DeepSeek V4 Flash GGUF, at commit efdadd4.

Canonical sweep from the speed-bench README: ctx 2048..65536, step 2048, 128 greedy decode tokens per frontier. Headlines: 485 t/s prefill / 39 t/s gen @2k, 238 t/s prefill / 21.5 t/s gen @65k.

CSV is in the current ds4-bench output format (includes the gen_first_ms / steady-state columns); plot_speed.py renders it as-is.


thomasjohnflaherty:
Heads-up that this overlaps with #584, which also adds M5 Max results at the same two paths (`speed-bench/m5_max.csv` and `m5_max_ts.svg`), so the two would collide on identical filenames.

They're not redundant, though: #584 is pure IQ2XXS full-residency, this PR is a mixed Q2+Q4 GGUF. The `kvcache_bytes` column is identical across both, so KV precision and the ctx 2048→65536 sweep are apples-to-apples — the divergence is purely the weight quant. The two curves cross: this run has higher prefill throughout (485 vs 340 t/s @2k), while #584's decode pulls ahead past ~32k. Having both quant points on the same machine class seems useful.

I've renamed mine to `m5_max_q2q4.csv` / `m5_max_q2q4_ts.svg` so both merge without collision; #584 can keep `m5_max.*`. Happy to fold into a single combined file with a quant column instead if you'd prefer.

Methodology note for comparability: these are steady-state numbers (the `gen_steady_tps` column) with a warm first token (`gen_first_ms` ~28 ms @2k), all greedy, `--gen-tokens 128`, at `efdadd4` (current upstream main).

OPS-NeoRetro (changes requested):
@thomasjohnflaherty, I think this is kind of stale but you should rebase and do reruns for more accurate results. Don't forget to kill and shut down all visible background apps and processes. This is already a month old.

PR #589, opened 2026-07-21
Support SSD streaming and MTP drafter when using two GPUs and caching of experts in VRAM

iSevenDays:
Tested on two Nvidia 4090d 48G. My command was
```
/root/antirez/ds4/ds4-server --cuda --ssd-streaming --ssd-streaming-cache-experts 40GB --gpu-vram 45,45 --ctx 262144 -m /root/antirez/ds4/gguf/DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix.gguf --mtp /root/antirez/ds4/gguf/DeepSeek-V4-Flash-DSpark-support.gguf --dspark --host 0.0.0.0 --port 8002
```

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -15881,28 +15881,51 @@ static bool metal_graph_configure_dspark_capture(
         return false;
     }
 
+    /* DSpark support weights, drafter kernels, and target-hidden capture all
+     * run on the configured executor tier (see ds4_session_prepare_dspark_draft,
+     * ds4.c:59635). The legacy ds4_gpu_tensor_alloc(bytes) always allocates on
+     * tier 0 (ds4_cuda.cu:2592), so under multi-tier SSD streaming — where the
+     * DSpark executor is tier 1 — every capture/drafter kernel that touched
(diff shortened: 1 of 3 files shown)


zom-2018:
One link-time issue for ROCm builds spotted while testing this branch: `ds4_gpu_debug_probe` is declared in `ds4_gpu.h:196` under `#if defined(DS4_ROCM_BUILD) || ...`, and the call sites are in shared `ds4.c` (e.g. `metal_graph_set_active_tier_batch`), but the only definition is in `ds4_cuda.cu:391` — the header comment says "Defined in ds4_cuda.cu / ds4_rocm.cu", yet `ds4_rocm.cu`/`rocm/*.cuh` contain no definition. Since the `strix-halo` target links `ds4.o` without `ds4_cuda.o`, ROCm builds should fail with an undefined reference.

Suggested fix: a no-op `ds4_gpu_debug_probe` definition for ROCm (in `ds4_rocm_compat.cu` or via the unavailable-stub file), or guard the declaration/call sites accordingly.

PR #590, opened 2026-07-21
DSpark: replay-free partial accepts + encoder-batched captures (+40% greedy, byte-lossless)

gilbert-barajas:
Two coupled improvements to the DSpark speculative-decode path (`fc9efd1`), Metal, no backend regressions.

## What

1. **Replay-free partial accepts.** On a partial accept of *k* drafts, the accepted prefix's KV is adopted from the verify pass via per-depth frontier checkpoints, instead of rolling back and re-decoding each accepted token. Re-decode on a partial accept cost ~24 ms/token — *more* than a plain decode; it's now ~0 (`replay=0.000`).
2. **Encoder-batched captures.** The per-depth checkpoint captures are encoded as compute-kernel copies on the already-open batch encoder (`ds4_gpu_tensor_copy_f32_inline`) rather than separate blit encoders, which were tearing down and rebuilding the encoder ~350–500× per verify. Bounds-checked blit fallback retained.

## Gains

Metal, DeepSeek-V4-Flash Q2_K, full residency, server decode-phase, cooled:

- **Greedy single-stream: 40.2 → 56.4 t/s (+40%)** on a high-acceptance prompt (≈51 on diverse prompts). `replay=0.000, accept_rate 94.5%` confirms partial accepts commit through the checkpoint path.

## Testing (per CONTRIBUTING.md)

Machine: Apple M5 Max, 128 GB, Metal. `make clean && make`.

- **Build:** warning-neutral — this branch emits the *identical* 27 warnings as the unmodified base (`comm` diff = 0 new); all are the macOS-27 `didModifyRange:` SDK deprecation, not from this change. `make cpu` builds clean.
- **Correctness:**
  - `make dspark-verify-depth` → **`worst_argmax_gap = 0.000`** — every committed greedy token (partial accepts included) is the exact plain-decode argmax.
  - `./ds4_test --metal-kernels` → OK; compressor/router exactness all `max_ulp = 0/0/0`.
- **No base-path regression (code-level):** the diff touches only DSpark-gated / checkpoint / capture code (and comments) — the non-speculative decode path is unmodified, so `--dspark`-off decode is byte-for-byte the same code. Base decode measured unchanged (server baseline 40.18 t/s; `ds4-bench` base 485 prefill / 39 gen @ ctx 2048).
- Greedy output is byte-identical to baseline where no near-tie occurs. The single Q2_K near-tie float flip is **pre-existing** — it fires on the unmodified base binary too (a batched-verify-vs-sequential-decode quantization tie-break), neither introduced nor fixed here.

`--logprob-vectors` was skipped — the official continuation `.vec` isn't in the repo; glad to run it against the official vectors. A clean `ds4-bench` before/after CSV is best captured on a thermally-settled box (this machine was saturated during testing).

## Notes for the reviewer

- **`net_saved` in the stats line reads understated** with the scheduler disabled: the replay-free path skips the re-decode whose timing fed `saved_ms`, so `saved_ms` no longer accrues in that mode and `net_saved` shows a misleading negative. **Wall-clock throughput is the ground truth (+40%).** Glad to credit `saved_ms` from the verify-pass timing in a follow-up.
- **No backend regression.** Metal change is additive (`_inline` copy + fallback). `ds4_cuda.cu` gets a 9-line `extern "C"` alias to the existing device copy. CPU / GLM / distributed / tensor-parallel sessions fall back to a plain one-token decode. C99 throughout the core.

Diff: **+363 / −10 across 4 files** (`ds4.c`, `ds4_metal.m`, `ds4_gpu.h`, `ds4_cuda.cu`).

---

A follow-up PR adds opt-in `DS4_DSPARK_SAMPLE=1` sampling-mode speculation at temperature > 0 via the rejection rule — distributionally lossless — and will be sent after this one merges.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -19515,10 +19515,13 @@ static bool metal_graph_capture_prefix_slot_attn_state(ds4_gpu_graph *g,
         !g->spec_prefix_attn_state_kv[il]) return true;
     const uint64_t bytes = ds4_gpu_tensor_bytes(g->layer_attn_state_kv[il]);
     g->spec_prefix_n_comp[slot][il] = g->layer_n_comp[il];
-    return ds4_gpu_tensor_copy(g->spec_prefix_attn_state_kv[il],
+    /* Compute-kernel copies: a blit here would end and reopen the batch
+     * compute encoder around every capture, and these run interleaved with
+     * the per-row compressor dispatches. */
+    return ds4_gpu_tensor_copy_f32_inline(g->spec_prefix_attn_state_kv[il],
                                  (uint64_t)slot * bytes,
                                  g->layer_attn_state_kv[il], 0, bytes) != 0 &&
-           ds4_gpu_tensor_copy(g->spec_prefix_attn_state_score[il],
+           ds4_gpu_tensor_copy_f32_inline(g->spec_prefix_attn_state_score[il],
                                  (uint64_t)slot * bytes,
                                  g->layer_attn_state_score[il], 0, bytes) != 0;
 }
@@ -19530,10 +19533,10 @@ static bool metal_graph_capture_prefix_slot_index_state(ds4_gpu_graph *g,
         !g->spec_prefix_index_state_kv[il]) return true;
     const uint64_t bytes = ds4_gpu_tensor_bytes(g->layer_index_state_kv[il]);
     g->spec_prefix_n_index_comp[slot][il] = g->layer_n_index_comp[il];
-    return ds4_gpu_tensor_copy(g->spec_prefix_index_state_kv[il],
+    return ds4_gpu_tensor_copy_f32_inline(g->spec_prefix_index_state_kv[il],
                                  (uint64_t)slot * bytes,
                                  g->layer_index_state_kv[il], 0, bytes) != 0 &&
-           ds4_gpu_tensor_copy(g->spec_prefix_index_state_score[il],
+           ds4_gpu_tensor_copy_f32_inline(g->spec_prefix_index_state_score[il],
                                  (uint64_t)slot * bytes,
                                  g->layer_index_state_score[il], 0, bytes) != 0;
 }
--- a/ds4.c
+++ b/ds4.c
@@ -14896,11 +14896,27 @@ typedef struct {
     ds4_gpu_tensor *spec_prefix1_attn_state_score[DS4_MAX_LAYER];
     ds4_gpu_tensor *spec_prefix1_index_state_kv[DS4_MAX_LAYER];
     ds4_gpu_tensor *spec_prefix1_index_state_score[DS4_MAX_LAYER];
+    /* DSpark per-depth prefix checkpoints.  The block verifier steps the
+     * compressor token by token, so the mutable frontier after every batch
+     * row can be captured in-stream for a few small blits.  Slot t holds the
+     * frontier after verify row t (the state after committing t+1 drafts); a
+     * partial accept of k drafts then commits by copying slot k-1 back
+     * instead of restoring the pre-block snapshot and re-decoding the
+     * accepted prefix token by token.  Each tensor packs all slots
+     * contiguously (slot-major), generalizing the prefix1 buffers. */
+    ds4_gpu_tensor *spec_prefix_attn_state_kv[DS4_MAX_LAYER];
+    ds4_gpu_tensor *spec_prefix_attn_state_score[DS4_MAX_LAYER];
+    ds4_gpu_tensor *spec_prefix_index_state_kv[DS4_MAX_LAYER];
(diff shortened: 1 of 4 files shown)


thomasjohnflaherty:
Independent datapoint from another M5 Max (128 GB, Metal, mixed Q2+Q4 `ds4flash.gguf`), on current main (`efdadd4`) — reproduces both behaviors this PR describes:

- **Partial-accept replay cost:** best-case greedy code-gen A/B came out a net loss — 38.2 t/s with `--dspark` vs 40.0 baseline, despite an 85% accept rate on proposed drafts (`net_saved=-149ms`, 60 proposals/150 cycles). Consistent with the ~24 ms/token replay overhead described here.
- **Pre-existing greedy near-tie flip:** `make dspark-acceptance` fails on the redis fixture with a greedy output divergence (baseline: "…known for its speed and support for various data types" / dspark: "…known for its speed and versatility in handling key-value") — matching the batched-verify-vs-sequential-decode tie-break you note as pre-existing.

Baseline sanity: base numbers here match yours (485 t/s prefill / ~39-40 t/s gen @2k — full sweep in #587). Happy to re-run the acceptance fixture and the A/B on this branch if useful.

gilbert-barajas:
Appreciate the independent M5 Max datapoint — that ~24 ms/token replay match is exactly the confirmation this needed. Same thing I hit before writing the fix: baseline DSpark net-negative despite a high accept rate.

And you've got the tie-break right — that greedy flip in `dspark-acceptance` is pre-existing (batched-verify vs sequential-decode), not introduced here. #590 just calls it out so it doesn't get pinned on the replay-free path.

Please do run the A/B on the branch — that's the number that matters. #590's whole job is killing that replay cost: partial accepts no longer replay, so your `net_saved=-149ms` should flip positive. On my M5 Max the same greedy code-gen A/B that lost on main comes out **~+40% on this branch, byte-lossless** — output is identical to baseline, so you can diff the two runs and confirm only the speed changed.

If it's easy: `make dspark-acceptance` on the branch, then the greedy `--dspark` vs baseline A/B on the same prompt. Curious whether you land near my +40% or somewhere else — either way it's a great datapoint. Thanks for digging in.

thomasjohnflaherty:
Ran the branch (`567e798`) on the same M5 Max / mixed Q2+Q4 setup. Results confirm the fix, with one tuning finding that may interest you:

- **Replay fix verified:** `replay=0.000` throughout, and the same code-gen A/B that lost on main (38.2 vs 40.0) now runs at parity on default settings — `net_saved` flipped from −149 ms to +9.6 ms.
- **Why parity and not +40% at defaults:** on this quant the drafter's confidence rarely clears the default 0.9 gate — `no_draft=120` of 147 cycles, so there was almost nothing to accelerate. Quant noise on the mixed Q2+Q4 seems to depress draft confidence relative to your Q2_K run.
- **`--dspark-confidence 0.5` unlocks it: 50.69 t/s vs 39.86 baseline (+27%), byte-identical output.** `accept_rate=80.65%`, `avg_accept=2.885`, `net_saved=+1121ms`, 40/52 cycles proposing. Without your replay fix this proposal volume would have been a heavy loss, so the win is directly attributable to this branch.
- `make dspark-acceptance`: hello fixture penalty shrank −199 ms → −19 ms (32-token prompts still can't amortize); the redis near-tie flip persists exactly as you documented — pre-existing, not this PR.

Takeaway from this datapoint: on mixed quants the default confidence gate may leave most of the branch's benefit on the table — possibly worth a note in the DSpark README section, or a quant-aware default down the line.

gilbert-barajas:
> Your confidence-gate finding was the right call — main has since moved the same direction:
> `a51e6ec` (07-26) retunes the default to **0.6 on Metal / 0.7 on CUDA and ROCm** for greedy and
> opportunistic sampling (0.8 for exact sampling), and `--help` + the README now document it. Your
> +27 % at 0.5 on the mixed Q2+Q4 quant is a better datapoint than mine for the "quant noise
> depresses draft confidence" mechanism — would you mind if it gets quoted in the README's DSpark
> section when the note lands? Separately: the replay-free commit from this thread is now in main as
> `74e8f11`; what survives of this PR is a small follow-up (encoding the in-verify frontier captures
> as compute-kernel copies instead of blit round-trips) which I'll re-measure on current main before
> offering. Either way, thanks — your replay-cost and gate measurements are the two numbers that made
> this real.

**thomasjohnflaherty (07-22, first comment)** — reproduced both behaviors on their M5 Max (replay cost,
pre-existing greedy near-tie flip). Drafted reply:

> (covered by the reply above — the near-tie flip you confirmed is now tracked upstream separately;
> no standalone reply needed unless m5Max prefers closing the loop explicitly.)

**gilbert-barajas (07-22, our own reply)** — no action; it is ours.

PR #602, opened 2026-07-24
rocm: AMD Instinct (CDNA/gfx90a) support with multi-GPU device placement

ewindisch:
Adds AMD Instinct (CDNA2, gfx90a — tested on dual MI210) support to the existing ROCm backend, which previously targeted RDNA3/Strix Halo only, plus native multi-GPU device placement and a set of wave64-specific performance fixes. Squashed history below is preserved as a series of well-scoped commits (each originally developed/tested as its own branch); happy to split into separate PRs if preferred, though GitHub doesn't support true cross-fork stacking so this landed as one PR — see commit-by-commit history for incremental review.

## What's included

1. **CDNA/gfx90a build support** — `make mi200`/`make mi300` targets, `DS4_ROCM_HAS_WMMA_W32`/`DS4_ROCM_HAS_MFMA_W64` arch macros, link fixes for this toolchain.
2. **CDNA MFMA Q8_0 batched matmul kernel** — the RDNA-only WMMA kernel didn't compile on gfx90a at all; adds the MFMA counterpart. Validated against a CPU f16 reference (`make test-rocm-matmul-q8`).
3. **Wave64 shuffle-mask/width fix** — `__shfl_*_sync` masks built from `threadIdx.x` were wrong on wave64 (right by coincidence on wave32); fixes the actual `HSA_STATUS_ERROR_EXCEPTION` crash on real hardware.
4. **Wave64 indexer/MoE hotlist port** — ports the indexer WMMA kernel's per-wave tile assignment to wave64 (was silently leaving half the output tiles unwritten on CDNA); gates wave32-shaped hotlist MoE kernels off on wave64 in favor of the already-correct portable kernels.
5. **Batch graph-prefill correctness fix** — root cause of a long-standing, hard-to-characterize multi-GPU/batch-prefill correctness bug: ROCm's multi-row fallback in the batched HC-normalization path only normalized row 0, leaving other prompt rows zero. Deterministic once isolated from sampling-temperature noise (which had made it look like GPU-scheduling nondeterminism across many earlier debugging passes). Fixed by porting the row-aware normalization the CUDA backend already had. Verified with 20+ independent multi-GPU runs, byte-identical across runs, coherent prompt-dependent generation.
6. **Native multi-GPU device placement** — in-process pipeline/layer placement across GPU tiers for ROCm (mirrors the CUDA backend's existing multi-tier scaffolding), including cross-device tensor copy with peer-access validation and a pinned-host-bounce fallback for GPUs without P2P/XGMI.
7. **Per-tier q8 f16 dequant-cache preload fix** — a preload-sweep "give up after first denial" latch was a single global instead of per-GPU-tier, so a budget denial on one tier permanently stopped the preload sweep from ever trying a less-constrained tier.
8. **Wave64-native kernel rewrites** — two of the hottest decode-path kernels (`matmul_q8_0_pair_f32_...`, `moe_gate_up_mid_decode_lut_...`) were tiled for 32-lane wavefronts (inherited from the RDNA3-only codebase) and, while correct on wave64 via commit 3's fix, left roughly half of CDNA's per-wavefront throughput unused. Adds true 64-lane-native counterparts, dispatched via a runtime wave-size check, with the original kernels left completely untouched (zero behavior change on RDNA3/Strix Halo). 46% and 24.5% reductions in average per-call kernel time respectively, measured via rocprofv3 kernel-level profiling on a 100-token decode run. Both validated against a CPU double-precision reference with non-block-aligned dimensions to stress tail handling.

## Testing

- `make test-rocm-matmul-q8`, `make test-rocm-matmul-pair-q8`, `make test-rocm-moe-gate-up-mid-hwarp32` all pass on gfx90a.
- 20+ independent multi-GPU generation runs, byte-identical outputs across runs, no garbling.
- Real end-to-end generation verified on a dual-MI210 box running an 80GB IQ2-quantized MoE model with `--gpu-devices 0,1 --gpu-vram auto`.
- All changes are additive/gated (new arch macros, new kernels behind runtime wave-size checks, new code paths behind `g_n_gpus`/device-count checks) — RDNA3/Strix Halo and single-GPU code paths are unmodified.

## Known gap, not addressed by this PR

Quality mode + SSD streaming hits a pre-existing "missing compact selected experts" decode error under ROCm. This reproduces independently of everything in this PR and is being tracked separately.


janbridley:
I am able to confirm this PR works across both GCDs of a MI250X. However, performance is quite poor. @ewindisch, do you have benchmark numbers for your 2xMI210?

```

> HIP_VISIBLE_DEVICES=0,1 ~/ds4/ds4-bench --rocm \
    -m gguf/DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix.gguf \
    --gpu-devices 0,1 --gpu-vram 45,61 \ # More even layer split improves prefill
    --prompt-file ~/ds4/speed-bench/promessi_sposi.txt \
    --ctx-start 2048 --ctx-max 262144 --step-mul 4.0 \
    -n 256 \
   
| ctx_tokens | prefill_tokens | prefill_tps | gen_tokens | gen_tps | gen_first_ms | gen_steady_tokens | gen_steady_tps | kvcache_bytes |
| ---------- | -------------- | ----------- | ---------- | ------- | ------------ | ----------------- | -------------- | ------------- |
| 2048       | 2048           | 112.83      | 256        | 13.45   | 70.816       | 255               | 13.46          | 52184460      |
| 8192       | 6144           | 148.51      | 256        | 12.94   | 73.172       | 255               | 12.95          | 136750476     |
| 32768      | 24576          | 142.60      | 256        | 11.99   | 77.927       | 255               | 12.00          | 475014540     |
| 131072     | 98304          | 118.40      | 256        | 10.00   | 94.482       | 255               | 10.01          | 0             |
```

ewindisch:
> I am able to confirm this PR works across both GCDs of a MI250X. However, performance is quite poor. @ewindisch, do you have benchmark numbers for your 2xMI210?

Unfortunately I have issues with my Mi210 GPUs that cause throttling unrelated to the code presented here, so they are not a good benchmark. I get approximately 4 tok/sec across two PCIe Mi210 cards, with transfers over PCIe Gen4 8x, with my cards throttled to 75W.

I agree that performance can/should be optimized. This patchset exists to implement functionality, not to optimize the kernels.

PR #605, opened 2026-07-25
CUDA: resident expert cache and faster selected-expert uploads for --ssd-streaming decode

iCreil:
While testing `--ssd-streaming` decode on CUDA (RTX PRO 6000, 96 GB, the
q4 Flash GGUF served from tmpfs) I measured a flat ~2.3 tok/s no matter
how warm the page cache is: every decode step re-uploads all selected
experts of every layer from the host map. The backend rewrite left the
CUDA resident expert cache API as stubs (`configured_count()` returns 0,
the seed hooks are no-ops), so only the per-layer staging buffers exist.
This PR restores the resident cache and then chips away at what the
timings said was left. Warm single-stream decode goes from ~2.3 to
29-32 tok/s; the four commits are independent steps and each one is
A/B-able on the same binary via env switches. Happy to rework any part.

## 1. Restore the resident expert cache (`~2.3 -> 16.7 tok/s`)

Brings back a VRAM LRU cache for streamed experts in front of the
staging buffers: hits become device-to-device copies, misses load the
expert into the cache once. Compared to the pre-rewrite cache this
version also:

- keeps one cache per (gate,down) byte-size class (4 classes):
  mixed-quant GGUFs interleave layers with different expert byte sizes,
  and a single cache keyed on the last-seen size is released and
  rebuilt on every size transition, never accumulating hits;
- lets prefill batch loads read the cache (hits become d2d copies) and
  append into free capacity, but never evict: a long prompt cannot
  cycle out the warm decode working set. Decode loads evict normally;
- allocates the staging buffers with the cache still in place and only
  sacrifices it (once, with a retry) when VRAM is genuinely short;
- implements the existing prefill-seed/hotlist hooks on top of the same
  cache, and only reports a budget to the shared working-set estimates
  when the operator opted in (budget flag or env), so default runs keep
  the current planner behavior.

Budget comes from the existing `--ssd-streaming-cache-experts` /
`DS4_CUDA_STREAMING_EXPERT_CACHE_N`, live-capped against free VRAM with
a reserve (`DS4_CUDA_STREAMING_EXPERT_CACHE_RESERVE_GB`, default 16).
`DS4_CUDA_STREAMING_EXPERT_CACHE_N=0` disables the cache (used for the
A/B numbers).

## 2. Skip the per-load VRAM query once the cache has settled (`16.7 -> 23.5`)

Host-side phase timing showed ~90% of the steady-state selected load
(~920us of ~1010us) inside cache prepare: when the VRAM-capped capacity
lands below the requested budget (e.g. 7000 requested, 5892 fitting),
the `capacity >= target` fast path can never hit and every load of
every layer falls through to `cudaMemGetInfo` (~0.9ms each — ~55ms of
the ~60ms decode step at 61 layers). Remember the request each class
settled at and return immediately while dims and budget are unchanged;
the mark is cleared on release and re-evaluated on any budget change.

## 3. Pipeline the streamed expert upload path (`23.5 -> ~24.5`)

A cache miss paid ~2.3ms for a ~13.5 MiB expert: each of the three
tensor copies pread()s the whole tensor into pinned staging as a single
64 MiB-chunk pass and then fully synchronizes the upload stream, so
reads and uploads never overlap. Split the copy into 2 MiB chunks
(`DS4_CUDA_STREAM_COPY_CHUNK_MB`) through the existing 4-buffer ring
and issue the three tensors of an expert as one async group with a
single flush.

## 4. Plain pageable uploads when there is no O_DIRECT path (`~24.5 -> 29-32`)

With the model map backed by page cache/tmpfs, uploading straight from
the (pageable) map runs ~3x faster than the staged pread bounce on this
box (~2.3 -> ~0.8ms per expert). Default to the plain copy only when no
O_DIRECT read path is active — real direct-I/O SSD streaming keeps the
staged path and its page-cache bypass. `DS4_CUDA_STREAM_MISS_PLAIN_COPY=1/0`
forces either way.

## Numbers

RTX PRO 6000 (96 GB), DeepSeek V4 Flash q4 GGUF in tmpfs,
`--ssd-streaming --ssd-streaming-cache-experts 7000 --ctx 32768
--prefill-chunk 1024`, 200 generated tokens per request, single stream,
steady state (2-3 requests of warm-up):

| configuration | tok/s |
|---|---|
| main (no resident cache) | ~2.3 flat |
| + resident cache | 16.7 |
| + settled budget | 23.5 |
| + pipelined uploads | ~24.5 |
| + plain pageable uploads | **29-32** |

Cache hit rate at steady state is ~96%; what remains of the step is
roughly one third miss uploads, the rest compute and the per-layer
router readback. Greedy output is unchanged in every A/B above
(cache on vs off, staged vs plain uploads — the copies are byte-exact).

With `--batched-session N` the resident cache is shared by the
sessions, so the per-session pipelining scales aggregates too: on top
of the #604 branch I measured 30 / 34 / 36.5 tok/s aggregate at 1/2/4
concurrent clients with greedy streams identical to the solo run.

## Notes and limitations

- CUDA only: Metal keeps its own (already resident-capable) cache
  paths untouched, and the ROCm layer-cache path is untouched too.
- The cache lives on the single-GPU streaming placement (same
  constraint as the selected staging buffers).
- Exact configs and per-step measurements are in the commit messages;
  I can share the bench harness if useful. If you had a different shape
  in mind for streaming caching, happy to adapt or split this.

### DIFF
--- a/ds4_cuda.cu
+++ b/ds4_cuda.cu
@@ -472,6 +472,11 @@ static void *g_stream_selected_stage[4];
 static cudaEvent_t g_stream_selected_stage_event[4];
 static uint64_t g_stream_selected_stage_bytes;
 static cudaStream_t g_stream_selected_upload_stream;
+/* Staging-buffer cursor shared across grouped async copies: buffer reuse is
+ * fenced by the per-buffer events, so back-to-back async copies (e.g. the
+ * three tensors of one streamed expert) pipeline the host read of one
+ * tensor with the device upload of the previous one. */
+static uint64_t g_stream_selected_stage_cursor;
 
 static int cuda_ok(cudaError_t err, const char *what);
 static const char *cuda_model_range_ptr_from_fd(
@@ -1719,22 +1724,45 @@ static int cuda_stream_selected_stage_pool_alloc(uint64_t bytes) {
         }
     }
     g_stream_selected_stage_bytes = bytes;
+    g_stream_selected_stage_cursor = 0;
     return 1;
 }
 
-static int cuda_model_copy_to_device_streamed(
+static int cuda_model_copy_to_device_streamed_flush(const char *what) {
+    if (!g_stream_selected_upload_stream) return 1;
+    const cudaError_t err =
+        cudaStreamSynchronize(g_stream_selected_upload_stream);
+    if (err != cudaSuccess) {
+        fprintf(stderr,
+                "ds4: CUDA streaming selected upload sync failed for %s: %s\n",
+                what ? what : "expert", cudaGetErrorString(err));
+        (void)cudaGetLastError();
+        return 0;
+    }
+    return 1;
+}
+
+static int cuda_model_copy_to_device_streamed_ex(
         char *dst,
         const void *model_map,
         uint64_t model_size,
         uint64_t offset,
         uint64_t bytes,
-        const char *what) {
+        const char *what,
+        int flush) {
     if (!dst || !model_map || offset > model_size ||
         bytes > model_size - offset) {
         return 0;
     }
     if (bytes == 0) return 1;
-    if (g_model_fd < 0 ||
+    /* DS4_CUDA_STREAM_MISS_PLAIN_COPY=1: upload straight from the (pageable)
+     * model map and let the driver stage internally — measurement switch for
+     * page-cache/tmpfs maps where the pread bounce is the bottleneck. */
+    static int plain_copy = -1;
+    if (plain_copy < 0) {
+        plain_copy = getenv("DS4_CUDA_STREAM_MISS_PLAIN_COPY") != NULL;
+    }
+    if (g_model_fd < 0 || plain_copy ||
         (g_model_fd_host_base != NULL && model_map != g_model_fd_host_base)) {
         return cuda_ok(cudaMemcpy(dst,
                                   (const char *)model_map + offset,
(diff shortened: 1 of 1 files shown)


nexus-cw:
We independently ported the transport half of this PR (the streamed-upload pipelining and plain-copy default — not the resident-cache redesign, which duplicates functionality we already have on our fork) onto an NVIDIA GB10 (unified CPU/GPU memory over NVLink-C2C) + MXFP4 expert-LRU stack, and measured a consistent 11-25% regression from the pipelining change specifically (isolated via an env toggle; the plain-copy default alone was neutral; byte-identical outputs throughout). Our read: the pipelining's benefit comes from hiding host-read latency behind PCIe DMA, which doesn't exist as a mechanism on memory-unified hardware, so the ring-cursor/deferred-sync bookkeeping becomes net overhead. Between this PR's numbers on discrete PCIe cards and ours on unified memory, the right transport appears to be arch-dependent rather than one-size — worth gating the pipelined path on detected-discrete vs unified at init, especially since much of the consumer local-AI hardware trend (Mac minis, Strix Halo, DGX Spark class) is unified. Full numbers in our fork's research/gb10/MEASUREMENTS.md.

iCreil:
Thanks for porting it and for the measurements — unified-memory hardware is a datapoint we couldn't produce (everything in this PR was measured on a discrete PCIe card), so this is really useful.

Your read matches ours: the pipelining exists purely to overlap host reads with PCIe DMA, so where that transfer doesn't exist the ring-cursor/deferred-sync bookkeeping is pure overhead. One note on how this PR is wired, in case it changes the picture on your side: the staged/pipelined path only engages on the direct-I/O read path (`g_model_direct_fd >= 0`); when the model map is host-resident (mmap/tmpfs) the default is already the plain single memcpy (`DS4_CUDA_STREAM_MISS_PLAIN_COPY` auto-defaults to 1 in that case). On a unified-memory system the model would typically be host-resident and land on the plain path out of the box — but if your port reaches the staged path anyway, gating it at init on detected unified memory (e.g. `cudaDeviceProp::integrated` or `cudaDevAttrPageableMemoryAccessUsesHostPageTables`) sounds reasonable and cheap. Happy to add that gate to this PR if the maintainer prefers it built-in, or as a follow-up.

Issue #500, opened 2026-07-05
Many prefills while using Claude Code extension


liqili:
Hi @antirez, 
Thank you so much for your great work. I have run ds4 Q2 model on my Mac Studio 128GB. The only issue I came across is the prefills which make the prompt very slow. This happens when I try to ask Claude to summarize my code base and give plans. Any idea how I can improve the performance?

./ds4-server \
  --model ds4flash.gguf \
  --host 0.0.0.0 \
  --port 8080 \
  --ctx 65536 \
  --kv-disk-dir ./cache/kv \
  --kv-disk-space-mb 8192
ds4: Metal device Apple M4 Max, 128.00 GiB RAM
ds4: Metal 4 tensor API disabled for pre-M5/pre-A19 devices
ds4: drift-patch flags hc_stable=on norm_unify=on kv_raw_f32=off rope_exp2_log2=off math_safe=off tensor_matmul=off
ds4: requesting Metal residency (may take tens of seconds)... done
ds4: warming Metal model views... done
ds4: Metal model views created in 1.617 ms, residency requested in 31074.930 ms, warmup 4.012 ms (mapped 93065.67 MiB from offset 5.09 MiB)
ds4: Metal mapped mmaped model as 2 overlapping shared buffers
ds4: metal backend initialized for graph diagnostics
0705 16:36:50 ds4-server: context buffers 1393.67 MiB (ctx=65536, backend=metal, prefill_chunk=4096, raw_kv_rows=4352, compressed_kv_rows=16386)
0705 16:36:50 ds4-server: KV disk cache ./cache/kv (budget=8192 MiB, cross-quant=accept, min=512, cold_max=30000, continued=10000, trim=32, align=2048, hit_half_life=21600s)
0705 16:36:50 ds4-server: listening on http://0.0.0.0:8080
0705 16:42:19 ds4-server: chat ctx=0..24527:24527 TOOLS prompt start
0705 16:42:19 ds4-server: chat ctx=0..24527:24527 TOOLS prefill chunk 0/24527 (0.0%) chunk=0.00 t/s avg=0.00 t/s 0.001s
0705 16:42:32 ds4-server: chat ctx=0..24527:24527 TOOLS prefill chunk 4096/24527 (16.7%) chunk=316.37 t/s avg=316.35 t/s 12.947s
[...]

PR #611, opened 2026-07-26
Server: chat/completions live tool continuation via tool_call_id

### LINKED ISSUE #609: OpenAI /v1/chat/completions tool rounds never hit live continuation (match=tool-output-ids); live slot discarded despite full-prompt common
On the Anthropic endpoint, tool rounds continue the live KV slot cheaply (`anthropic live continuation match=tool-output-ids`, delta-only prefill — see #500 for a working example at 85K context with 400–900 token deltas). On the OpenAI chat/completions endpoint, the equivalent never happens: every tool round logs a live miss where `common` equals the entire previous prompt, the live slot is discarded, and prefill restarts from the oldest disk checkpoint.

Example (Open WebUI, native function calling, OpenAI protocol — four consecutive tool rounds in one conversation):

```
live kv cache miss live=6034 prompt=6128 common=5817 reason=token-mismatch
kv cache stored tokens=6034 trimmed=0 reason=evict key=token-text size=102.08 MiB save=16.6 ms
chat ctx=5779..6128:349 TOOLS prompt done 1.857s
live kv cache miss live=6213 prompt=8238 common=6128 reason=token-mismatch
kv cache stored tokens=6213 trimmed=0 reason=evict key=token-text size=104.43 MiB save=16.5 ms
chat ctx=5779..8238:2459 TOOLS prompt done 9.231s
live kv cache miss live=8633 prompt=14903 common=8238 reason=token-mismatch
kv cache stored tokens=8633 trimmed=0 reason=evict key=token-text size=136.21 MiB save=15.1 ms
chat ctx=5779..14903:9124 TOOLS prompt done 31.253s
live kv cache miss live=15139 prompt=20460 common=14903 reason=token-mismatch
kv cache stored tokens=15139 trimmed=0 reason=evict key=token-text size=221.61 MiB save=25.7 ms
```

Note the pattern: `common` always reaches exactly the previous prompt's end (6128, 8238, 14903), because the only divergence is the re-serialised assistant tool-call turn — the client strips thinking to an empty `<think></think>` and re-encodes the tool call in OpenAI JSON format. Instead of rolling back to `common` and prefilling the delta, the live slot is discarded and prefill restarts from the system-prompt disk checkpoint (5779), so each round pays (context − 5779) tokens. At ~90K context this compounds to ~6-minute tool rounds.

The evicted snapshots store the divergent generated tail untrimmed (`trimmed=0` — this persists with `--kv-cache-boundary-trim-tokens 1024`, which appears to only affect cold-path saves), so they accumulate with `hits=0` and evict each other; only the system-prompt checkpoint ever hits.

Exact DSML tool replay is active and reports healthy: tool-call lines show `raw_tool_text=1`, and `--trace` shows `tool_replay: mem=N disk=0 canonical=0 missing_ids=0` with correct `call_*` IDs present in the replayed request.

A rollback-to-common on the live slot, or an OpenAI-path equivalent of the `tool-output-ids` continuation matching on `tool_call_id`, would reduce these rounds to delta-only prefill like the Anthropic path.

**Environment**

- ds4: main @ 0a7ad77
- macOS 26.5.2, Apple M5 Max, 128 GB
- Model: `DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2.gguf` (fully resident, no SSD streaming)
- Launch: `./ds4-server -m <model> --ctx 100000 --kv-disk-dir ~/.cache/ds4-kv --kv-disk-space-mb 8192 --kv-cache-boundary-trim-tokens 1024 --host 127.0.0.1 --port 8300`
- Client: Open WebUI 0.10.2, Native function calling, streaming, `stream_options.include_usage`

Happy to test a branch — this setup reproduces the pattern on every multi-round tool conversation.

starforge-labs:
Fixes #609 - On the Anthropic endpoint, tool rounds continue the live KV slot cheaply
(`anthropic live continuation match=tool-output-ids`). On the OpenAI
chat/completions endpoint the equivalent never fires: clients replay the
assistant tool-call turn re-serialised (thinking stripped to an empty
block, tool call re-encoded from JSON), so exact token/text prefix
matching fails at the previous generation's tail even though `common`
reaches the entire previous prompt. The live slot is discarded and
prefill restarts from an older disk checkpoint — O(context) per tool
round. Before (Open WebUI, native function calling):

```
live kv cache miss live=15139 prompt=20460 common=14903 reason=token-mismatch
kv cache stored tokens=15139 trimmed=0 reason=evict key=token-text size=221.61 MiB
chat ctx=5779..14903:9124 TOOLS prompt done 31.253s
```

The trailing `role:"tool"` message's `tool_call_id` is the same precise
continuation handle as Anthropic's `tool_use_id`, so this PR reuses that
fast path for API_OPENAI chat: collect trailing tool ids at parse time
(`chat_prepare_live_continuation`, mirroring the Anthropic prepare), and
when the ids and live token frontier match, keep the sampled KV and
append only EOS + tool results + next assistant prefix.

After, same client, same conversation shape, ~24k context:

```
chat live continuation match=tool-output-ids ids=1 cached=19597 prompt=22207
chat ctx=19597..22207:2610 TOOLS prompt done 11.234s
chat live continuation match=tool-output-ids ids=1 cached=23507 prompt=23552
chat ctx=23507..23552:45 TOOLS prompt done 0.798s
chat live continuation match=tool-output-ids ids=1 cached=24151 prompt=24164
chat ctx=24151..24164:13 TOOLS prompt done 0.230s
```

Tool rounds drop from 45-90s to 0.2-11s, and the per-round ~300 MiB
evict/store KV disk cycle disappears. A genuine mid-conversation prompt
change still evicts and re-prefills correctly (verified by editing
history mid-chat).

Design notes:

- State is stored in the existing `anthropic_live_*` request/slot
  fields; the matcher, remember, and clear sites are protocol-gated. A
  rename to something protocol-neutral would touch many more lines —
  happy to do that if preferred.
- The frontier is only remembered for chat when
  `should_canonicalize_tool_checkpoint()` is false, since
  canonicalization rewrites the live session and would strand the
  remembered frontier.
- Every gate failure returns 0 and falls through to the existing
  matching chain, so behaviour is unchanged whenever the fast path
  doesn't apply.
- Adds `test_chat_live_tail_renders_tool_results_only` covering the
  role:"tool" tail rendering and the protocol gate; full embedded suite
  passes.

## Testing

- `make test`: full suite passes — all tracks including `--server`,
  `--tool-call-quality`, `--long-context`, and `--logprob-vectors` —
  on Apple M5 Max 128GB, Metal backend, macOS 26.5.2, model
  `DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2.gguf` (q2).
- `make cpu`: builds clean (verified on x86-64 Linux).
- Live verification: same model fully resident on Metal, Open WebUI 0.10.x
  native function calling — eleven consecutive `chat live continuation` hits
  across chained tool rounds at 6k–24k context (0.2–11s per round vs 45–90s
  before), zero evict/store KV disk writes between rounds, and correct
  evict + re-prefill on a genuine mid-conversation prompt edit.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -2964,6 +2964,51 @@ static void anthropic_prepare_live_continuation(request *r,
                                          &r->tool_orders, r->think_mode);
 }
 
+static bool chat_msg_is_tool_result_tail(const chat_msg *m) {
+    return m && (!strcmp(m->role, "tool") || !strcmp(m->role, "function")) &&
+           ((m->tool_call_id && m->tool_call_id[0]) ||
+            m->tool_call_ids_len > 0);
+}
+
+/* Prepare the OpenAI chat/completions live-tool fast path.
+ *
+ * Chat/completions has no server-side response object, but the tool_call_id
+ * replayed in trailing role:"tool" messages is still a precise continuation
+ * handle, exactly like Anthropic's tool_use_id.  Clients re-serialise the
+ * assistant tool-call turn (thinking stripped, tool call re-encoded as JSON),
+ * so exact token/text prefix matching fails at the previous generation's tail
+ * even though the visible history is a faithful replay.  When the trailing
+ * tool ids match the live sampled frontier, generate_job() can keep the
+ * sampled KV and append only EOS + tool results + next assistant prefix,
+ * instead of discarding the live slot and re-prefilling from an older
+ * checkpoint.  State is stored in the shared anthropic_live_* request fields;
+ * the matcher and remember/clear sites are protocol-gated. */
+static void chat_prepare_live_continuation(request *r,
+                                           const chat_msgs *msgs) {
+    if (!r || r->api != API_OPENAI || !msgs || msgs->len == 0) return;
+
+    int tail_end = msgs->len;
+    while (tail_end > 0 && role_is_system(msgs->v[tail_end - 1].role)) tail_end--;
+    int tail_start = tail_end;
+    while (tail_start > 0 &&
+           chat_msg_is_tool_result_tail(&msgs->v[tail_start - 1]))
+    {
+        tail_start--;
+    }
+    if (tail_start == tail_end) return;
+
+    stop_list_clear(&r->anthropic_live_call_ids);
+    for (int i = tail_start; i < msgs->len; i++) {
+        chat_msg_collect_tool_call_ids(&msgs->v[i], &r->anthropic_live_call_ids);
+    }
+    if (r->anthropic_live_call_ids.len == 0) return;
+
+    free(r->anthropic_live_suffix_text);
+    r->anthropic_live_suffix_text =
+        render_live_tool_tail_for_syntax(r->model_syntax, msgs, tail_start,
+                                         &r->tool_orders, r->think_mode);
+}
+
 /* The API parsers are intentionally selective JSON parsers: they keep only
  * fields that affect model semantics, rendering, streaming, or cache keys, and
  * skip extension fields.  The output is always a rendered DS4 chat/completion
@@ -3126,6 +3171,7 @@ static bool parse_chat_request(ds4_engine *e, server *s, const char *body, int d
         think_mode_from_enabled(thinking_enabled, reasoning_effort), ctx_size);
     kv_cache_restore_tool_memory_for_messages(s, &msgs);
     tool_memory_attach_to_messages(s, &msgs, &r->tool_replay);
+    chat_prepare_live_continuation(r, &msgs);
     const char *active_tool_schemas = r->has_tools ? tool_schemas : NULL;
     r->prompt_preserves_reasoning =
         chat_history_uses_tool_context(&msgs, active_tool_schemas);
@@ -9718,7 +9764,8 @@ static int anthropic_live_continuation_prompt(server *s, server_slot *slot,
                                               ds4_tokens *effective_prompt,
                                               int *matched_ids) {
     if (!s || !slot || !req || !effective_prompt) return 0;
-    if (req->api != API_ANTHROPIC || !req->anthropic_live_suffix_text) return 0;
+    if (req->api != API_ANTHROPIC && req->api != API_OPENAI) return 0;
+    if (!req->anthropic_live_suffix_text) return 0;
     if (req->anthropic_live_call_ids.len == 0) return 0;
     if (!anthropic_live_matches_request(s, slot,
                                         &req->anthropic_live_call_ids,
@@ -11147,7 +11194,8 @@ static void generate_job(server *s, server_slot *slot, job *j) {
                                                     &anthropic_live_match_ids);
         if (cached > 0) {
             anthropic_live_continuation = true;
-            cache_source = "anthropic-tool-output";
+            cache_source = j->req.api == API_ANTHROPIC ? "anthropic-tool-output"
+                                                       : "chat-tool-output";
             prompt_for_sync = &effective_prompt;
         }
     }
@@ -11283,7 +11331,8 @@ static void generate_job(server *s, server_slot *slot, job *j) {
                    prompt_tokens);
     } else if (anthropic_live_continuation) {
         server_log(DS4_LOG_PREFILL,
-                   "ds4-server: anthropic live continuation match=tool-output-ids ids=%d cached=%d prompt=%d",
+                   "ds4-server: %s live continuation match=tool-output-ids ids=%d cached=%d prompt=%d",
+                   j->req.api == API_ANTHROPIC ? "anthropic" : "chat",
                    anthropic_live_match_ids,
                    cached,
                    prompt_tokens);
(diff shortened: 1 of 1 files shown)


Snail3D:
Another affected client, confirming this fix matters beyond Open WebUI: **Hermes Agent** (OpenAI chat/completions + tools) against ds4-server (DeepSeek V4 Flash, Metal, `--batched-session 4`, `--kv-disk-dir`).

From a live session today (macOS, Apple Silicon), the same signature — `common` reaches the previous prompt's tail, slot discarded anyway:

```
live kv cache miss live=21704 prompt=32809 common=21366 reason=token-mismatch
live kv cache miss live=32790 prompt=37838 common=20542 reason=token-mismatch
```

Hermes additionally spawns a background review fork (a second concurrent agent loop replaying the same conversation with a divergent tail), so without live continuation every fork/main alternation mutually evicts the 4 KV slots. Client-observed cache rates bounce between 64–100% turn to turn, with 50–130s latencies at only ~40K context. At the extreme end we watched a 175K-context client cold-prefill from token 0 four times in 15 minutes (`ctx=0..175249:175249 prompt done 643.560s`).

Happy to build this PR branch and validate it against Hermes — the exact repro environment is running here (ds4-server on 127.0.0.1:8001, Hermes gateway driving it over Telegram). Just say the word.

Flor1an-B:
Independent verification, on the exact use case this would help me with (OpenCode, which talks to `ds4-server` over `/v1/chat/completions`, not the Anthropic endpoint).

Applied this diff to a fresh clone of `main`, built clean, ran a minimal 2-turn tool-call exchange (`read_file` tool, real production quant) against both a baseline build and this patch, same conversation on both:

**Baseline (no patch):**
```
live kv cache miss live=395 prompt=419 common=254 reason=token-mismatch
```

**With this patch:**
```
chat live continuation match=tool-output-ids ids=1 cached=395 prompt=420
```

Cache hit instead of a miss, exactly as described — only the small delta needs prefilling instead of restarting from `common`. Small example so the absolute numbers aren't dramatic, but the mechanism triggers correctly and matches the qualitative fix described here and in #609. This is the same class of problem reported in #500 (Claude Code / large-codebase tool rounds), for the subset of users going through the OpenAI-compatible endpoint rather than the native Anthropic one — worth linking the two threads together if it isn't already, since #500's later comments (config tuning) don't actually address this path.

blakesplay:
A beautiful fix and should be merged ASAP. Thank you!

starforge-labs:
Rebased onto latest main and verified clean build.

starforge-labs:
> Another affected client, confirming this fix matters beyond Open WebUI: **Hermes Agent** (OpenAI chat/completions + tools) against ds4-server (DeepSeek V4 Flash, Metal, `--batched-session 4`, `--kv-disk-dir`).
> 
> From a live session today (macOS, Apple Silicon), the same signature — `common` reaches the previous prompt's tail, slot discarded anyway:
> 
> ```
> live kv cache miss live=21704 prompt=32809 common=21366 reason=token-mismatch
> live kv cache miss live=32790 prompt=37838 common=20542 reason=token-mismatch
> ```
> 
> Hermes additionally spawns a background review fork (a second concurrent agent loop replaying the same conversation with a divergent tail), so without live continuation every fork/main alternation mutually evicts the 4 KV slots. Client-observed cache rates bounce between 64–100% turn to turn, with 50–130s latencies at only ~40K context. At the extreme end we watched a 175K-context client cold-prefill from token 0 four times in 15 minutes (`ctx=0..175249:175249 prompt done 643.560s`).
> 
> Happy to build this PR branch and validate it against Hermes — the exact repro environment is running here (ds4-server on 127.0.0.1:8001, Hermes gateway driving it over Telegram). Just say the word.

Appreciate you testing it — I think this explains part of your picture but probably not the whole thing, and it's worth separating the two failure modes so whoever picks this up next isn't chasing one fix for two bugs.

The fork/main slot contention (4 KV slots, mutually evicted every alternation) is a scheduling problem #611 was never built for — that's #668/#698 territory, not this patch.

But there's likely a second issue specific to Hermes' request shape that #611 also doesn't cover. I ran this exact patch against Hermes Agent myself a couple weeks back (separately from Open WebUI, which is what #611 was built and tested against) and found `chat live continuation` essentially never fires for Hermes' tool-call traffic — every round falls straight through to `reason=token-mismatch` → full disk evict/reload, same 350–500MB-per-round pattern you're describing. Never fully root-caused which part of Hermes' serialization breaks the tail match (message ordering, extra metadata between rounds — haven't isolated it), but it's a real gap in the matcher, independent of the slot-contention issue.

So: even with #668/#698 solving the multi-slot scheduling, I'd expect Hermes traffic to still cold-reload on every tool round underneath that, because the continuation matcher itself doesn't recognize its request shape.

If I get time this weekend I'll try to actually isolate what's different about Hermes' tool-call serialization and see if the matcher can be widened to cover it. Will report back either way — happy to compare notes if you get to it first.

For now, hoping @antirez gets a chance to merge this one — the Open WebUI case it was built for is confirmed fixed by two more people since, and the gaps above (Hermes, fork contention) both look like natural follow-ups rather than reasons to hold this back.

JordiPosthumus:
I have a current-main candidate for the same OpenAI tool-continuation problem covered here and in #811. I would prefer to contribute the useful hardening to existing work rather than open another competing PR.

The candidate checks original pre-call history before tool-memory attachment, visible assistant content and ordered calls, explicit reasoning when supplied, the full live token/image frontier, and unique slot ownership. It also tests the handoff where the client sends tool results before the producing worker publishes its binding. Omitted reasoning remains supported.

On M3 Ultra/Metal with DeepSeek-V4-Flash-Vision-Exp MXFP4, generated-tool continuations reused 1,109 tokens (exact replay), 1,112 (omitted reasoning), and 1,227 (with an image), with a 14-token suffix. Identity changes reject the new shortcut. A separate cancellation integration test restored/reused 1,272 tokens with zero suffix; that required request-local rollback, not just tool-call routing. This is focused validation, not a full-suite or forced real-model race claim.

The independently audited candidate is [92af008c](https://github.com/JordiPosthumus/ds4/commit/92af008c1465e199f82a39d07202202da244e21e). Would a focused hardening follow-up against this branch be useful? I can adapt the relevant pieces and tests without bundling the cancellation or vision-persistence work.

PR #647, opened 2026-08-01
cuda: implement real per-(layer,expert) LRU for --ssd-streaming-cache-experts

### LINKED ISSUE #639: --ssd-streaming-cache-experts is silently inert on CUDA — expert cache never populated, hit rate 0
On the CUDA backend, the `--ssd-streaming-cache-experts` budget has no effect: `ds4_gpu_set_streaming_expert_cache_budget()` / `_expert_bytes()` are no-op stubs (`(void)bytes;`) and `ds4_gpu_stream_expert_cache_configured_count()` hard-returns 0. `cuda_stream_selected_cache_begin_load()` unconditionally invalidates and re-fetches every selected expert's gate/up/down bytes from the mmap'd file on every token — the flag parses, prints its budget, and does nothing. The Metal backend has a real per-(layer,expert) LRU (`g_stream_expert_cache`, `..._peek` / `_install_loaded` / `_prune_global`), so this looks like the CUDA half never landed rather than a design choice.

Measured on a GB10 (DGX Spark class, sm_121), DeepSeek-V4-Flash, `--ssd-streaming`:

- decode tok/s is flat (~1.0) across cache budgets 8GB-100GB, cold vs popularity-preload identical
- instrumented fetch counters show hit rate exactly 0.000 and ~3GB/token read from file at every budget
- NVMe sustained at ~85% of its ceiling while GPU sits at 18% SM — fully disk-bound on re-fetch traffic

We ported the Metal LRU design to CUDA on our research fork (device-resident entries keyed (layer,expert), global LRU prune, budget semantics per the CLI contract): decode went 1.03 -> 2.96 tok/s at 100GB budget with 81% measured hit rate, greedy outputs byte-identical. Branch with the port + measurements: https://github.com/nexus-cw/ds4/tree/research/gb10 (LRU port in 416533c; counters in 1961967; numbers in research/gb10/MEASUREMENTS.md). Happy to shape it into a PR if the approach fits how you'd want the CUDA side to work.

nexus-cw:
Fixes #639.

## Problem

`--ssd-streaming-cache-experts` has never had any effect on the CUDA backend: `cuda_stream_selected_cache_begin_load()` unconditionally invalidates and re-fetches every selected expert's gate/up/down bytes from the mapped model file on every call, and `ds4_gpu_set_streaming_expert_cache_budget()`/`_expert_bytes()` were no-op stubs while `ds4_gpu_stream_expert_cache_configured_count()` hard-returned 0 -- at any budget, on any model. Metal's GLM streaming implementation, by contrast, has always had a real per-(layer,expert) LRU (`g_stream_expert_cache`) with genuine hit/miss tracking.

## Design

**Metal-LRU parity.** Ports Metal's design to CUDA: a persistent, device-resident `(layer,expert)`-keyed table (`cuda_stream_expert_cache_entry`, a fixed 80x384 array) is consulted by `cuda_stream_expert_cache_peek()` before `cuda_stream_selected_cache_begin_load()` falls back to its existing mapped-file fetch. A hit is served with a device-to-device `cudaMemcpy` into the existing per-call packed staging buffer that the downstream decode kernels already read from, so the kernel-facing interface is unchanged; a miss still pays the existing fetch and then additionally installs those bytes into the persistent cache (`cuda_stream_expert_cache_install()`) so the entry survives past this call. Eviction is global least-recently-used across the whole table.

**Budget semantics per CLI contract.** `N` in `--ssd-streaming-cache-experts N` (or the NGB form, already converted to a plain expert count upstream of the GPU backend by `ds4.c`, identically for Metal and CUDA) is the entry-count budget, clamped to the table's fixed bound -- no CUDA-specific CLI or budget-math changes were needed.

**Device-resident entries.** Unlike Metal's single-size-class slab allocator (which excludes mixed-precision "boosted" layers off its uniform size class from caching), CUDA's per-entry `cudaMalloc` has no such uniform-size requirement, so all routed layers -- including any boosted ones -- participate in one cache here.

**Pooled allocator.** `cuda_stream_expert_cache_install()`/`_clear_entry()` originally called `cudaMalloc`/`cudaFree` directly on every miss/eviction, which thrashes badly at a too-small budget (install, evict, install, evict, every call, with no offsetting hit-rate benefit) -- a real regression below the no-cache baseline. A size-keyed device-buffer pool (`cuda_stream_expert_pool_class_for/_alloc/_free/_release_all`) turns install/evict into pool pop/push, with real `cudaMalloc` only on genuine growth; every existing whole-cache-reset call site (model swap, streaming-mode toggle, budget change) still releases the pool back to the driver, so real teardown is unchanged.

**Env-gated counters.** Adds `DS4_CUDA_STREAM_STATS=1`: four host-side counters (fetch_calls, expert_fetches, cache hits/misses, bytes_from_file/cache) at the fetch decision point, printed via a new `ds4_gpu_print_cuda_stream_stats()` called from the CLI's single-shot generation path (`ds4_cli.c`). A no-op stub is added to `ds4_metal.m` for build symmetry -- Metal already has its own richer instrumentation via `--expert-profile`. This closes the observability gap that made the original no-op possible to ship unnoticed, and gives reviewers a way to directly observe hit rate on their own hardware.

## Commits

1. `cuda: add DS4_CUDA_STREAM_STATS=1 diagnostic counters for the SSD streaming expert fetch path` -- observability scaffolding, structurally-zero hits/bytes_from_cache at this point (no cache exists yet).
2. `cuda: implement a real per-(layer,expert) LRU expert cache for --ssd-streaming-cache-experts` -- the fix itself.
3. `cuda: pool the expert LRU's device buffers to avoid cudaMalloc/cudaFree thrash at small budgets` -- correctness-adjacent quality fix so a too-small budget doesn't regress below the no-cache baseline.

## Measurements (from the linked issue, IQ2_XXS DeepSeek V4 Flash artifact, single GB10-class GPU)

Short-bench decode throughput at a 100 GB cache budget: **1.03 -> 2.96 t/s** (measured hit rate ~81%). A long multi-prompt session converges to a **~96-98%** hit rate as the working set of routed experts stabilizes. In this PR's own testing (20 GB budget, single process, an 8-question multi-topic prompt over ~220 generated tokens): hit_rate=0.515, decode **1.62 -> 4.41 t/s**, cache saturating at its configured 2522-entry budget.

## Test evidence

- **Correctness/determinism**: greedy-decode output is byte-identical with the cache enabled (20 GB budget) versus a clean pre-cache build (this repo's upstream/main, where no persistent cache exists at all) -- verified on both a single-question prompt ("What is the capital of France?" -> "The capital of France is Paris.") and the 8-question multi-topic prompt above (all 8 answers identical, byte-for-byte, across both binaries).
- **Hit-rate observability**: `DS4_CUDA_STREAM_STATS=1` on the multi-topic session reports `hits=8256 misses=7788 hit_rate=0.515 ... budget=2522 entries=2522` -- a real, populated cache, not the previous structural zero.
- **No regression in default (non-streaming) mode**: identical output and comparable throughput (16.23 vs 16.30 t/s) with `--ssd-streaming` omitted entirely, on branch versus upstream/main.
- **Build**: `make clean && make cuda-spark` -- clean, no warnings.
- **`./ds4_test`**: this branch shows the same pre-existing-flaky section set as a clean `upstream/main` run in this environment (`tool-call-quality`, `logprob-vectors`, `metal-kernels`, and `metal-tensor-equivalence`/`think-tool-recovery`, which flip between runs on both trees) -- no new failing sections introduced by this change.

### DIFF
--- a/ds4_cli.c
+++ b/ds4_cli.c
@@ -664,7 +664,9 @@ static int run_sampled_generation(ds4_engine *engine, const cli_config *cfg, con
             "ds4: prefill: %.2f t/s, generation: %.2f t/s\n",
             prefill_s > 0.0 ? (double)prompt->len / prefill_s : 0.0,
             decode_s > 0.0 ? (double)generated / decode_s : 0.0);
+#if !defined(DS4_NO_GPU) && !defined(__APPLE__)
     ds4_gpu_print_cuda_stream_stats();
+#endif
 
     ds4_session_free(session);
     return 0;
--- a/ds4_cli.c
+++ b/ds4_cli.c
@@ -1,5 +1,6 @@
 #include "ds4.h"
 #include "ds4_distributed.h"
+#include "ds4_gpu.h"
 #include "ds4_gpu_args.h"
 #include "ds4_tp.h"
 #include "ds4_help.h"
@@ -663,6 +664,7 @@ static int run_sampled_generation(ds4_engine *engine, const cli_config *cfg, con
             "ds4: prefill: %.2f t/s, generation: %.2f t/s\n",
             prefill_s > 0.0 ? (double)prompt->len / prefill_s : 0.0,
             decode_s > 0.0 ? (double)generated / decode_s : 0.0);
+    ds4_gpu_print_cuda_stream_stats();
 
     ds4_session_free(session);
     return 0;
--- a/ds4_cuda.cu
+++ b/ds4_cuda.cu
@@ -153,18 +153,19 @@ static cuda_stream_selected_cache g_stream_selected_cache;
  * DS4_CUDA_STREAM_STATS=1 diagnostic counters. Closes the CUDA-side
  * instrumentation gap versus Metal's --expert-profile / g_stream_expert_cache_hits
  * family: CUDA's SSD-streaming selected-expert path
- * (cuda_stream_selected_cache_begin_load, below) has never had per-generation
- * fetch/hit/miss/byte counters. These are cheap host-side atomics incremented
- * at the exact fetch decision points traced for the P3a expert-cache-sweep
- * diagnosis: every call unconditionally invalidates any prior staged content
- * (see cuda_stream_selected_cache_invalidate() immediately below) and then
- * copies every selected expert's gate/up/down bytes fresh from the mapped
- * model file view via cuda_model_copy_to_device_streamed() -- there is no
- * lookup against a persistent, expert-id-keyed cache anywhere in this path
- * (unlike ds4_metal.m's g_stream_expert_cache, a real per-(layer,expert) LRU).
- * So "hits"/"bytes_from_cache" are wired here for symmetry with the Metal
- * counters and to make that absence an empirically-reportable zero rather
- * than an assertion -- not because a hit path exists today.
+ * (cuda_stream_selected_cache_begin_load, below) previously had no
+ * fetch/hit/miss/byte counters, and (before the persistent
+ * cuda_stream_expert_cache_* LRU further below was added) no real cache to
+ * count hits against either -- every begin_load() call unconditionally
+ * invalidated the same-call packed staging buffer
+ * (g_stream_selected_cache, via cuda_stream_selected_cache_invalidate())
+ * and re-fetched every selected expert's gate/up/down bytes fresh from the
+ * mapped model file view. That gap is now closed: cuda_stream_expert_cache_peek()
+ * is consulted before every fetch, so hits/bytes_from_cache are real
+ * measurements of the persistent per-(layer,expert) LRU below, not
+ * structurally-zero placeholders. These remain cheap host-side counters
+ * (not std::atomic) because this fetch path only ever runs on the one
+ * CUDA-dispatch thread.
  */
 static uint64_t g_cuda_stream_stats_fetch_calls;     /* begin_load invocations */
 static uint64_t g_cuda_stream_stats_expert_fetches;  /* distinct (layer,expert) loads */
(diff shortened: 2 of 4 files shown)


nexus-cw:
Noting overlap with #605 (iCreil), which attacks the same missing-CUDA-expert-cache problem and was filed first — apologies for not spotting it before opening this. The approaches look complementary rather than competing: #605 adds a resident cache plus upload pipelining and pageable-copy optimizations (impressive 29-32 tok/s on an RTX PRO 6000), while this PR ports the Metal per-(layer,expert) LRU semantics — budget accounting per the CLI contract, frequency+recency eviction, pooled buffers, and env-gated hit/miss counters. Happy to rebase this on top of #605 if that lands first, or to fold the LRU/budget/counters pieces into #605's structure if iCreil prefers — whichever shape is most useful. Measurements in both PRs agree on the headline: the expert cache is worth 3-14x on streamed decode.

nexus-cw:
Pushed one more commit (ab98473) with a fix for an accounting bug we found while hardening this cache in production.

**The bug:** the LRU budget was enforced only as an entry *count* derived from the dominant slab size class, while CUDA caches every routed layer at its actual per-expert size — layers off the slab class cache entries larger than `expert_bytes`, so `N entries * slab_bytes` under-states real device consumption. On top of that, the buffer pool's free lists park real `cudaMalloc`'d memory that no accounting saw. Running a GB10 node with a nominal 100GB cache budget, we observed ~121GB of actual device memory consumed by the cache, which pushed the box into GPU-availability thrash.

**The fix:**
- valid entries are charged at their actual gate+up+down byte sizes;
- pool free-list bytes are counted as parked;
- installs make room *before* allocating — trimming parked pool buffers first (data-lossless), then evicting LRU — so counted + parked never exceeds `budget_experts * expert_bytes`, the exact dynamic-cache byte figure ds4.c plans and logs at startup. The entry-count cap remains as a secondary bound.

One subtlety: parked bytes reusable by the incoming request are treated as slack rather than new demand. A first cut that double-counted them forced a trim-then-malloc cycle on every miss at cap and cost ~30% decode throughput — exactly the alloc/free churn the pool exists to prevent.

`DS4_CUDA_STREAM_STATS=1` now also self-reports the cache's true footprint (budget / counted / pool-parked / device total / persistent pinned-host staging / entries) every 4096 fetch calls, and the end-of-generation stats call in ds4_cli.c is now compiled only into CUDA builds, which fixes `make cpu` linking on this branch.

The fix is production-verified on the same GB10 node at a 70GB budget: actual consumption stays at the nominal figure with no measurable throughput cost versus the pre-fix pooled path.

PR #652, opened 2026-08-02
moe: stage the IQ2 dictionary on every path in the gate/up tile kernels

tylerwagler:
## What

`moe_gate_up_mid_expert_tile8_row32_kernel` and `_tile8_rowspan_kernel` staged `s_iq2_grid`/`s_iq2_signs` **inside** their `xq_blocks <= 16` activation-staging branch, but `dev_dot_iq2_xxs_q8_K_block8_deq_lut` reads that dictionary on **every** path. An `xq_blocks > 16` layer decoded its IQ2 weights against uninitialised shared memory.

## Reachability

**Latent, not live.** v5mx has `n_embd 4096` → `xq_blocks` is exactly 16, so the guard holds by one block. It is one dim away from silent garbage (`n_embd 7168` → `xq_blocks 28`). The two IQ2 *down* kernels already stage the dictionary unconditionally — this pair was the outlier, not the rule. Q2_K kernels don't use the LUT and are unaffected.

## Cost (ptxas -v, sm_120f, measured on the GB10)

| kernel | regs before → after | smem |
|---|---|---|
| `tile8_row32` | 96 → 121 | 39,552 (unchanged) |
| `tile8_rowspan` | 120 → 119 | 39,552 (unchanged) |

`row32` grows 25 registers because the dictionary loops now always execute and their addressing stays live across the barrier. **Occupancy is unchanged** — at 256 threads/block, 121 regs still holds 2 blocks/SM (`65,536/(121*256) = 2.12`), and `2*39,552 < 102,400`. That keeps it clear of the occupancy cliff that made the NT=16 token-tile widen a regression in 21e992a.

## Numerics

The shipped path is untouched: at `xq_blocks <= 16` the same dictionary is staged into the same arrays in the same order before the same barrier, so the reachable path is bit-identical.

## Validation

- Compile-verified for `sm_120f` with CUDA 13.0 on the GB10.
- ptxas resource comparison vs baseline, above.
- **Not measured end-to-end.** Worth a `prefill-bitexact` gate run and a throughput triplet before merging, given the register change on a hot kernel.

https://claude.ai/code/session_01DDbxA7geWoR5gvgCXg7E5W


tylerwagler:
Opened against the wrong repository in error — this belongs on a downstream fork. Closing immediately; apologies for the noise.

PR #623, opened 2026-07-28
ROCm: restore DeepSeek V4 Flash decode performance and Q4 SSD streaming

kyuz0:
## Summary

This restores the ROCm DeepSeek V4 Flash paths that regressed when the backend
moved Q8 decode activations to FP32:

- restores prequantized Q8 decode for regular, paired, HC-expand, and
  low-rank attention-output projections;
- fixes Q4 SSD streaming by staging the compact selected-expert table before
  the pointer-backed routed-MoE launch;
- keeps both changes out of GLM. The Q8 selector checks the model family, and
  the SSD fix explicitly excludes `DS4_MODEL_FAMILY_GLM_DSA`;
- keeps `--quality` on the existing FP32 activation path;
- provides `DS4_ROCM_DSV4_PREQUANT_DECODE=0` as a diagnostic rollback.

## Tested models

- Single-node resident inference:
  `DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix.gguf`
- Two-node distributed inference and single-node SSD streaming:
  `DeepSeek-V4-Flash-Q4KExperts-F16HC-F16Compressor-F16Indexer-Q8Attn-Q8Shared-Q8Out-chat-v2-imatrix.gguf`

## Performance

AMD Radeon 8060S / gfx1151, ROCm 7.2.4.

Resident IQ2, 64-token decode:

| Context | FP32 activation path | Restored Q8 path |
| ---: | ---: | ---: |
| 512 | 9.19 t/s | 16.12 t/s |
| 1024 | 9.13 t/s | 15.89 t/s |
| 2048 | 9.01 t/s | 15.51 t/s |
| 4096 | 8.76 t/s | 14.82 t/s |

Prefill stayed effectively unchanged. Decode improved by 70-76%.

Two-node Q4 decode, layers `0:21 / 22:output`, compared with the current
upstream base (`antirez/main` at `54b36ed`):

| Context | Upstream | This PR | Improvement |
| ---: | ---: | ---: | ---: |
| 512 | 8.32 t/s | 13.94 t/s | +67.5% |
| 1024 | 8.25 t/s | 13.73 t/s | +66.4% |
| 2048 | 8.17 t/s | 13.48 t/s | +65.0% |
| 4096 | 7.97 t/s | 12.93 t/s | +62.2% |

Profiling localized the regression to Q8 attention and shared-expert
projections. Routed Q4 MoE performance was unchanged.

[...]

PR #656, opened 2026-08-02
Add ROCm MXFP4 routed expert kernels

jyatesdotdev:
## What this does

DeepSeek models quantized with MXFP4 routed experts cannot run on the ROCm backend today: the base branch fails on the first feed-forward batch (`gpu layer 0 ffn batch encode failed`), and the CPU backend rejects the expert tensor type. This pull request adds the missing kernels and then optimizes them for gfx1151 (Strix Halo):

- Resident MXFP4 × Q8_K kernels for the routed-expert gate/up and down projections, in both decode and prefill.
- The decode kernels read weights coalesced across the 32 lanes of a wavefront, unpack MXFP4 nibbles with register permutations, and keep the gate/up and down reductions entirely in registers.
- Batches of two to four tokens run through the direct decode kernels instead of launching partially filled prefill tiles.
- The prefill tile kernel vectorizes its activation loads, fuses the gate and up projections per token and row, keeps tile state in registers, and sizes shared-memory staging to the actual activation width, with an explicit opt-in for launches that need more than 48 KiB.
- Tiles that read the same expert weights are scheduled near each other in time so the weights are served from the L2 cache. On the production tensor shape this took the gate/up kernel from 247.7 ms to 222 ms per call.
- Three optional variants can be enabled with environment variables. Each produces byte-identical output to the default path, because the accumulation order is unchanged by construction; this is verified by comparing output buffers byte for byte, not by an error tolerance:
  - `DS4_ROCM_ENABLE_MXFP4_TILE4` — four-token tiles, about 29 KiB of shared memory, four workgroups per compute unit
  - `DS4_ROCM_ENABLE_MXFP4_ROW64` — 64-row tiles with 512 threads
  - `DS4_ROCM_MXFP4_DOWN_RGROUP=1..8` — the down-projection kernel amortizes activation staging across row groups
- Two build fixes this hardware needs: Strix host objects are compiled position-independent (without it the branch base does not link on the TheRock toolchain), and the quality scorer is compiled and linked in two steps, because hipcc inserts `-x c` ahead of the object list and otherwise tries to parse relocatable objects as C source.
- A new `make test-rocm` target runs the core test suite on hosts that only have ROCm.

## Performance

Same two-machine setup, same prompt of about 7,500 tokens, identical settings, both formats on this branch (current main plus these commits); the only change is the expert format of the model file. The MXFP4 run has the optional variants enabled (`DS4_ROCM_ENABLE_MXFP4_TILE4=1`, `DS4_ROCM_MXFP4_DOWN_RGROUP=4`); they are inert for Q4_K.

| expert format | prefill | decode |
|---|---:|---:|
| MXFP4 (this pull request) | 173.6 tokens/s | 13.5 tokens/s |
| Q4_K (reference) | 89.5 tokens/s | 14.0 tokens/s |

Prefill is 94% faster at decode parity. Decode is close to identical between the formats because single-token decode is dominated by dense attention and matrix products, not by the expert kernels.

## Correctness

- `make test-mxfp4-rocm` passes with zero failures for the `mid` and `out` stages at 1, 2, 3, 4, 5, 32, 128, and 512 tokens against a scalar CPU reference, and again at the production tensor shape (`MODEL_DIM=7168`, `N_EXPERT=8`).
- Each optional variant is additionally compared byte for byte against the default kernel output.
- End to end, on the released MXFP4 model file (79 GB, which fits on this hardware only when split across two machines): coherent greedy output at the rates in the table above.

## Scope

The branch is based on current `main` (`84cc882`) and contains only these commits; there are no merge commits. It originally targeted the `ds4f-mxfp4` feature branch; that branch has since been fully merged into main, so the pull request now targets main directly, rebased and re-verified.

Related performance work that is not specific to MXFP4 is left to separate pull requests so this one stays reviewable: batched multi-row decode kernels (#821), distributed transport work (#820), HIP graph capture (#819), server-side prefill reuse (#818), and the QKV normalization fusion (#730).

Three decode-side variants — register prefetch, cooperative shared-memory staging, and multiple rows per warp — were implemented, measured, and excluded because they were neutral or slower than the baseline (15.11 tokens/s baseline against 14.94, 11.18, and 14.43).

## Known limitation: SSD streaming (pre-existing)

Streaming Flash-family models from SSD fails on the ROCm backend at the branch base for every routed-expert quantization, with or without these commits. Main already carries the fix (#623); this branch does not contain it. One follow-up is worth noting for when it does: with MXFP4, the invalid cache state currently produces wrong output instead of failing with an error, and deserves a guard.

The full build, test, and measurement log is in the comments below.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -339,6 +339,12 @@ tests/test_mxfp4_rocm.o: tests/test_mxfp4_rocm.c ds4_gpu.h
 tests/test_mxfp4_rocm: tests/test_mxfp4_rocm.o ds4_rocm.o
 	$(HIPCC) $(ROCM_CFLAGS) -o $@ $^ $(ROCM_LDLIBS)
 
+tests/bench_mxfp4_rocm.o: tests/bench_mxfp4_rocm.c ds4_gpu.h
+	$(CC) $(filter-out -ffast-math,$(CFLAGS)) $(ROCM_HOST_CFLAGS) -DDS4_ROCM_BUILD -I. -c -o $@ $<
+
+tests/bench_mxfp4_rocm: tests/bench_mxfp4_rocm.o ds4_rocm.o
+	$(HIPCC) $(ROCM_CFLAGS) -o $@ $^ $(ROCM_LDLIBS)
+
 test-mxfp4-rocm: tests/test_mxfp4_rocm
 	./tests/test_mxfp4_rocm
 
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,7 @@ CPU_CORE_OBJS = ds4_cpu.o ds4_distributed.o ds4_tp.o ds4_ssd.o ds4_layer_pack.o
 CUDA_LDLIBS ?= -lm -Xcompiler -pthread -L$(CUDA_HOME)/targets/sbsa-linux/lib -L$(CUDA_HOME)/lib64 -lcudart -lcublas
 HIPCC ?= $(shell command -v hipcc 2>/dev/null || echo /opt/rocm/bin/hipcc)
 ROCM_ARCH ?= gfx1151
+ROCM_HOST_CFLAGS ?= -fPIC
 ROCM_CFLAGS ?= -O3 -ffast-math -g -fno-finite-math-only -pthread -D__HIP_PLATFORM_AMD__ -Wno-unused-command-line-argument --offload-arch=$(ROCM_ARCH)
 ROCM_LDLIBS ?= -lm -pthread -lhipblas -lhipblaslt
 DS4_LINK ?= $(NVCC) $(NVCCFLAGS)
@@ -62,7 +63,7 @@ DS4_LINK_LIBS ?= $(CUDA_LDLIBS)
 METAL_LDLIBS := $(LDLIBS)
 endif
 
-.PHONY: all help clean test test-metal-session-batch test-mxfp4-cuda test-cuda-session-batch test-cuda-mixed-batch dspark-acceptance dspark-verify-depth mtp-verify-depth cpu cuda cuda-spark cuda-generic cuda-regression strix-halo rocm
+.PHONY: all help clean test test-metal-session-batch test-mxfp4-cuda test-mxfp4-rocm test-cuda-session-batch test-cuda-mixed-batch dspark-acceptance dspark-verify-depth mtp-verify-depth cpu cuda cuda-spark cuda-generic cuda-regression strix-halo rocm
 
 ifeq ($(UNAME_S),Darwin)
 .PHONY: metal-decode-schedule-bench metal-prefill-variant-bench check-mxfp4-half-lut
@@ -156,6 +157,7 @@ help:
 	@echo "  make cuda CUDA_ARCH=sm_N Build CUDA with an explicit nvcc -arch value"
 	@echo "  make strix-halo          Build ROCm for Strix Halo / gfx1151"
 	@echo "  make rocm                Alias for make strix-halo"
+	@echo "  make test-mxfp4-rocm     Build and run the synthetic ROCm MXFP4 MoE test"
 	@echo "  make cpu                 Build CPU-only ./ds4, ./ds4-server, ./ds4-bench, ./ds4-eval, and ./ds4-agent"
 	@echo "  make test                Build and run tests"
 	@echo "  make dspark-verify-depth Run DSpark speculative verification smoke if support GGUF is present"
@@ -179,7 +181,7 @@ cuda:
 strix-halo:
 	$(MAKE) -B ds4 ds4-server ds4-bench ds4-eval ds4-agent \
 		CORE_OBJS="ds4.o ds4_distributed.o ds4_tp.o ds4_ssd.o ds4_rocm.o ds4_rocm_compat.o ds4_rocm_unavailable.o ds4_layer_pack.o" \
-		CFLAGS="$(CFLAGS) -DDS4_ROCM_BUILD" \
+		CFLAGS="$(CFLAGS) $(ROCM_HOST_CFLAGS) -DDS4_ROCM_BUILD" \
 		DS4_LINK="$(HIPCC) $(ROCM_CFLAGS)" \
 		DS4_LINK_LIBS="$(ROCM_LDLIBS)"
 
@@ -330,6 +332,15 @@ cuda/mmq/ds4_repack.o: cuda/mmq/ds4_repack.cu cuda/mmq/ds4_repack.h
 
 ds4_rocm.o: ds4_rocm.cu ds4_gpu.h ds4_iq2_tables_cuda.inc $(ROCM_SRCS)
 	$(HIPCC) $(ROCM_CFLAGS) -c -o $@ ds4_rocm.cu
+
+tests/test_mxfp4_rocm.o: tests/test_mxfp4_rocm.c ds4_gpu.h
+	$(CC) $(filter-out -ffast-math,$(CFLAGS)) $(ROCM_HOST_CFLAGS) -DDS4_ROCM_BUILD -I. -c -o $@ $<
+
+tests/test_mxfp4_rocm: tests/test_mxfp4_rocm.o ds4_rocm.o
+	$(HIPCC) $(ROCM_CFLAGS) -o $@ $^ $(ROCM_LDLIBS)
+
+test-mxfp4-rocm: tests/test_mxfp4_rocm
+	./tests/test_mxfp4_rocm
 
 ds4_rocm_compat.o: ds4_rocm_compat.cu ds4_gpu.h ds4_gpu_mgpu.h ds4_gpu_args.h
 	$(HIPCC) $(ROCM_CFLAGS) -c -o $@ ds4_rocm_compat.cu
@@ -488,4 +499,4 @@ mxfp4-dot-test: tests/test_mxfp4_dot.c
 	./tests/test_mxfp4_dot
 
 clean:
-	rm -f ds4 ds4-server ds4-bench ds4-eval ds4-agent ds4_cpu ds4_native ds4_server_test ds4_test ds4_agent_test gguf-tools/quality-testing/score_official gguf-tools/quality-testing/score_official.o speed-bench/metal_decode_schedule_bench speed-bench/metal_prefill_variant_bench speed-bench/*.o tests/test_q4k_dot tests/test_mxfp4_dot tests/test_mxfp4_metal tests/test_mxfp4_cuda tests/test_metal_session_batch tests/test_gpu_xdev tests/test_gpu_model_cache tests/test_gpu_lookup_cache_strict tests/test_engine_mgpu_refusal tests/test_engine_mgpu_runtime tests/test_engine_correctness tests/test_sampling tests/test_cuda_session_batch tests/test_cuda_mixed_batch tests/*.o *.o tests/cuda_long_context_smoke tests/cuda_long_context_smoke.o
+	rm -f ds4 ds4-server ds4-bench ds4-eval ds4-agent ds4_cpu ds4_native ds4_server_test ds4_test ds4_agent_test gguf-tools/quality-testing/score_official gguf-tools/quality-testing/score_official.o speed-bench/metal_decode_schedule_bench speed-bench/metal_prefill_variant_bench speed-bench/*.o tests/test_q4k_dot tests/test_mxfp4_dot tests/test_mxfp4_metal tests/test_mxfp4_rocm tests/test_mxfp4_cuda tests/test_metal_session_batch tests/test_gpu_xdev tests/test_gpu_model_cache tests/test_gpu_lookup_cache_strict tests/test_engine_mgpu_refusal tests/test_engine_mgpu_runtime tests/test_engine_correctness tests/test_sampling tests/test_cuda_session_batch tests/test_cuda_mixed_batch tests/*.o *.o tests/cuda_long_context_smoke tests/cuda_long_context_smoke.o
--- a/Makefile
+++ b/Makefile
@@ -465,10 +465,18 @@ test-cuda-mixed-batch: tests/test_cuda_mixed_batch
 endif
 
 ds4_test: ds4_test.o ds4_help.o ds4_kvstore.o rax.o $(CORE_OBJS)
+ifeq ($(UNAME_S),Darwin)
+	$(CC) $(CFLAGS) -o $@ ds4_test.o ds4_help.o ds4_kvstore.o rax.o $(CORE_OBJS) $(METAL_LDLIBS)
+else
 	$(DS4_LINK) -o $@ ds4_test.o ds4_help.o ds4_kvstore.o rax.o $(CORE_OBJS) $(DS4_LINK_LIBS)
+endif
 
 ds4_agent_test: ds4_agent_test.o ds4_help.o ds4_web.o ds4_kvstore.o linenoise.o $(CORE_OBJS)
+ifeq ($(UNAME_S),Darwin)
+	$(CC) $(CFLAGS) -o $@ ds4_agent_test.o ds4_help.o ds4_web.o ds4_kvstore.o linenoise.o $(CORE_OBJS) $(METAL_LDLIBS)
+else
 	$(DS4_LINK) -o $@ ds4_agent_test.o ds4_help.o ds4_web.o ds4_kvstore.o linenoise.o $(CORE_OBJS) $(DS4_LINK_LIBS)
+endif
 
 test: ds4_test ds4_agent_test ds4-eval q4k-dot-test mxfp4-dot-test \
 	tests/test_layer_pack tests/test_engine_mgpu_placement tests/test_gpu_args \
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ DS4_LINK_LIBS ?= $(CUDA_LDLIBS)
 METAL_LDLIBS := $(LDLIBS)
 endif
 
-.PHONY: all help clean test test-metal-session-batch test-mxfp4-cuda test-mxfp4-rocm test-cuda-session-batch test-cuda-mixed-batch dspark-acceptance dspark-verify-depth mtp-verify-depth cpu cuda cuda-spark cuda-generic cuda-regression strix-halo rocm
+.PHONY: all help clean test test-rocm test-metal-session-batch test-mxfp4-cuda test-mxfp4-rocm test-cuda-session-batch test-cuda-mixed-batch dspark-acceptance dspark-verify-depth mtp-verify-depth cpu cuda cuda-spark cuda-generic cuda-regression strix-halo rocm
 
 ifeq ($(UNAME_S),Darwin)
 .PHONY: metal-decode-schedule-bench metal-prefill-variant-bench check-mxfp4-half-lut
@@ -158,6 +158,7 @@ help:
 	@echo "  make strix-halo          Build ROCm for Strix Halo / gfx1151"
 	@echo "  make rocm                Alias for make strix-halo"
 	@echo "  make test-mxfp4-rocm     Build and run the synthetic ROCm MXFP4 MoE test"
+	@echo "  make test-rocm           Core regression suite on ROCm-only hosts"
 	@echo "  make cpu                 Build CPU-only ./ds4, ./ds4-server, ./ds4-bench, ./ds4-eval, and ./ds4-agent"
 	@echo "  make test                Build and run tests"
 	@echo "  make dspark-verify-depth Run DSpark speculative verification smoke if support GGUF is present"
@@ -187,6 +188,26 @@ strix-halo:
 
 rocm: strix-halo
 
+# Core regression suite for ROCm-only hosts: the CUDA-specific binaries
+# (tests/test_sampling, the CUDA session/mixed-batch oracles) are not part
+# of this target; run them through `make test` / `make cuda-regression` on
+# CUDA hosts.  Everything else mirrors `make test`.
+test-rocm:
+	$(MAKE) -B ds4_test ds4_agent_test ds4-eval q4k-dot-test mxfp4-dot-test \
+		tests/test_layer_pack tests/test_engine_mgpu_placement tests/test_gpu_args \
+		ds4 ds4-server ds4-bench ds4-agent \
+		CORE_OBJS="ds4.o ds4_distributed.o ds4_tp.o ds4_ssd.o ds4_rocm.o ds4_rocm_compat.o ds4_rocm_unavailable.o ds4_layer_pack.o" \
+		CFLAGS="$(CFLAGS) $(ROCM_HOST_CFLAGS) -DDS4_ROCM_BUILD" \
+		DS4_LINK="$(HIPCC) $(ROCM_CFLAGS)" \
+		DS4_LINK_LIBS="$(ROCM_LDLIBS)"
+	./ds4-eval --self-test-extractors
+	./ds4_agent_test
+	./ds4_test
+	./tests/test_layer_pack
+	./tests/test_engine_mgpu_placement
+	./tests/test_gpu_args
+	./tests/test_gpu_args_cli.sh
+
 ds4: ds4_cli.o ds4_help.o linenoise.o ds4_gpu_args.o $(CORE_OBJS)
 	$(DS4_LINK) -o $@ $^ $(DS4_LINK_LIBS)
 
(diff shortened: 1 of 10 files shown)


jyatesdotdev:
I have two strix halos and wanted to run the MXFP4 variant, so spent a decent amount of time working with Codex (Sol 5.6) to get this working. I have more unrelated changes for getting MTP working across the systems and I'd like to see how difficult it would be to enable RDMA over thunderbolt (though that later part involves way more than just this repository). Let me know if you'd like any changes for this PR!

jyatesdotdev:
Also note the current numbers are from iommu=pt, I'm testing now with it off, which should produce better numbers. I'm going to be daily driving this on my own hardware for a while, so I'll report back if there are any quality issues. I'm using pi for my harness.

jyatesdotdev:
Follow-up to my earlier IOMMU note: `amd_iommu=off` produced a modest, repeatable speedup.

These are **optimized deployment results, not standalone #656 throughput**. The tested build combines the resident ROCm MXFP4 routed-expert kernels from this PR with the restored Q8 prequant decode path from #623. No #623 code is included in this PR.

The A/B held the binaries, local-NVMe GGUF, Fedora kernel `7.1.5-101.fc43`, model split, and benchmark flags fixed across the reboot. The only configuration change was `iommu=pt` versus `amd_iommu=off`; both retained the same 126,976 MiB GTT and page settings.

- two Radeon 8060S / `gfx1151` systems
- layers `0:21 / 22:output`
- 32-bit distributed activations
- 4096-token local/distributed prefill chunks, window 3
- 128 greedy decode tokens at 2K, 8K, and 32K
- one warm-up followed by three measured runs per configuration
- table reports medians

| Context | `iommu=pt` prefill / decode | `amd_iommu=off` prefill / decode | Prefill / decode delta |
| ---: | ---: | ---: | ---: |
| 2K | 67.70 / 10.67 tok/s | 69.07 / 10.95 tok/s | +2.0% / +2.6% |
| 8K | 84.08 / 10.15 tok/s | 88.40 / 10.47 tok/s | +5.1% / +3.2% |
| 32K | 114.88 / 9.58 tok/s | 120.99 / 9.93 tok/s | +5.3% / +3.7% |

All 18 dumped frontier-logit files were byte-identical at each matching context across every repetition and both boot modes, so this affected performance rather than model output.

On these systems, `amd_iommu=off` disables the NPU/SVA path. The result points to a roughly 3% platform-level decode gain and up to 5% longer-prefill gain; it does not indicate a change in the MXFP4 kernel itself.

jyatesdotdev:
Follow-up on commit `7d63867`, which contains the focused MXFP4 optimization:

- coalesces MXFP4 weight reads across wave32 lanes and performs nibble unpacking with register permutations;
- keeps gate/up, SwiGLU, and down reductions register-resident on gfx1151;
- routes 2–4 token batches through direct kernels instead of partially filled expert-sorted tiles;
- introduces no register spills or private-memory use.

The CPU-oracle test now covers 1, 2, 3, 4, 5, 32, 128, and 512 tokens. All cases passed, with worst observed absolute error of `2.15e-6`; production first-frontier logits were byte-identical across 129,280 values.

Controlled parent/head results were:

- single-token decode: `11.35 → 14.56 tok/s` (`+28.3%`);
- two-token aggregate: `11.13 → 14.88 tok/s` (`+33.7%`);
- four-token aggregate: `17.01 → 23.61 tok/s` (`+38.8%`).

The Q4_K and IOMMU measurements in the PR are additional format/platform context and are not attributed to this commit.

jyatesdotdev:
@antirez any chance for a review or comment? Curious to know what you think and if I should spend some more time improving this! Thanks!

jyatesdotdev:
@kyuz0 tagging you as well since you've done a lot of work in this area, would appreciate a look at this!

kyuz0:
Thank you! I'll take a look this weekend, this is great! In general, I would like to see if we can push performance a bit more, there ought to be some space!

jyatesdotdev:
I agree! I was able to get some level of RDMA working via a thunderbolt 5 (GPU-accessible zero-copy NHI messaging, one level below true GPU-to-GPU RDMA, which eliminated most transport copying), but the serial model split had almost no transport latency left to recover, and direct GTT kernel output was slightly slower, so I didn't get much improvement in TPS there. I'm still working on kernel optimizations, so hopefully I'll find some more room!

jyatesdotdev:
Update after splitting the follow-up work into separate review scopes:

- **#656 remains the focused MXFP4 routed-expert PR.** Its public head is still `rocm-mxfp4-pr` at `7d63867`; the performance and correctness claims in this PR remain attributable only to that MXFP4 work.
- **The Q/KV launch fusion is now separate PR #730.** [`rocm-qkv-kv-rope-fusion`](https://github.com/jyatesdotdev/ds4/tree/rocm-qkv-kv-rope-fusion) is one focused commit (`c0cc54d`) directly atop `ds4f-mxfp4`. It has no source dependency on #656 and applies to both Q4_K and MXFP4 DeepSeek V4 models.
- **The descriptor-framed v3/NHI work remains an integration branch only.** [`rocm-mxfp4-experts`](https://github.com/jyatesdotdev/ds4/tree/rocm-mxfp4-experts) at `b6c6edb` contains the broader experimental transport stack and is not part of #656 or #730.

For reference, the Q/KV fusion in #730 removes one ROCm launch per decoded layer. A controlled full-model TCP-v3 A/B across the two Strix Halo hosts measured:

- server decode throughput: `14.57487 -> 14.60320 tok/s` (`+0.1944%`)
- combined GPU stream time per token: `-0.2175%` (throughput-equivalent `+0.2180%`)
- sampled QKV segment: `-1.3762%`
- all 40 responses / 3,200 completion tokens had the same normalized response SHA-256; the strict differential suite was bit-exact, including YaRN and fallback cases

This keeps the Q/KV result and the experimental NHI work out of the performance attribution and review scope for #656. Further MXFP4 kernel experiments will remain focused on the #656 path.

jyatesdotdev:
@antirez not sure if you'd had a chance to look through this yet or do any testing, but I'm currently working through another round with Fable this time to see if I can increase the speed. It seems to think it's possible to get up to 40ish TPS with parallel mixed experts so that each GPU is running in parallel rather than serially, which joined with my NHI work in another project (https://github.com/jyatesdotdev/strix-rdma), might be possible?

kyuz0:
> @antirez not sure if you'd had a chance to look through this yet or do any testing, but I'm currently working through another round with Fable this time to see if I can increase the speed. It seems to think it's possible to get up to 40ish TPS with parallel mixed experts so that each GPU is running in parallel rather than serially, which joined with my NHI work in another project (https://github.com/jyatesdotdev/strix-rdma), might be possible?

I'm not antirez, but I want to look at this and fully test it, we can get it merged. I haven't had a chance as I planned this weekend as there is quite a lot going on, but please can you keep working on optimising performance? 

Is this work only affecting MXFP4, or is the performance work also affecting the other quants?

jyatesdotdev:
> > @antirez not sure if you'd had a chance to look through this yet or do any testing, but I'm currently working through another round with Fable this time to see if I can increase the speed. It seems to think it's possible to get up to 40ish TPS with parallel mixed experts so that each GPU is running in parallel rather than serially, which joined with my NHI work in another project (https://github.com/jyatesdotdev/strix-rdma), might be possible?
> 
> I'm not antirez, but I want to look at this and fully test it, we can get it merged. I haven't had a chance as I planned this weekend as there is quite a lot going on, but please can you keep working on optimising performance?
> 
> Is this work only affecting MXFP4, or is the performance work also affecting the other quants?

@kyuz0 I'm sorry, got you mixed up! No worries, and yup I'm going to keep trying to improve performance. I was able to nearly double prefill at least with the latest commits (I'll publish some benchmarks). Decode is proving to be more difficult to improve, and the best I've gotten so far is 14 tokens/s (and MTP doesn't seem to bring any gains, if anything it decreases decode). 

I'm also limited by only using thunderbolt, because I'm using dual framework desktops, and I'm not willing to strap an addon card to the chassis for real RDMA. I'm not sure if proper RDMA is even possible with just thunderbolt, though I've gotten NHI to work, it's not real GPU-to-GPU.

I've been trying to limit this PR to just MXFP4, though without touching the code other quants use, I might have hit a limit with how far I can go. I'll publish a bunch of details on what I've tried so far a bit later. 

This PR https://github.com/antirez/ds4/pull/730 does touch more than just MXFP4, though the increase I got there is really minimal.

jyatesdotdev:
## State vs upstream (2026-08-14)

For reviewers: where this PR sits relative to the moving upstream branches.

**Base currency.** This PR targets `ds4f-mxfp4` (tip `80df56a`, 2026-08-01), which is 99 commits behind `antirez/main` (`84cc882`, 2026-08-09). The PR head (`67e8e0f`) is built on kyuz0/main (`8e0f997`, 2026-08-04), itself 61 behind `antirez/main`. Neither base has been rebased onto main. Happy to rebase the stack onto `antirez/main` — or retarget if `ds4f-mxfp4` is being retired/refreshed — whenever you prefer; the rebase is mechanical on our side and everything below is gate-verified after any move.

**Scope update (branch updated today).** The branch now carries the full ROCm decode campaign beyond the original MXFP4 kernel work, squashed into logical commits on top of the rebased MXFP4/tile8 stack:

- exact-rows k-row kernel tier (bit-exact multi-row Q8_0/F16 matvecs; BLAS batch tiers are never bit-exact on gfx1151 and stay out of exact paths)
- staggered multi-session decode (aggregate +32%/+44% at concurrency 2/3; `DS4_DIST_STAGGER_DISABLE` kill switch)
- the k-row distributed MTP verifier: spans 56.7/55.1 → 45.9/44.3 ms at rows=2, plus eval-skip, the exact-rows batched output head (verify 91.0 → 89.1 ms), and the merged comp-split walk (default-on, `DS4_SPEC_NO_COMP_SPLIT` kill switch)
- staged default-off: HIP decode-island graph capture and speculative round pipelining across the two-host verify (both measured, exact, honest negatives/neutrals documented)

Net effect measured on the dual-Strix-Halo pair: single-session decode 11.35 → 14.67 tok/s vs the pre-PR baseline (MTP on beats MTP off 14.67 vs 13.75), prefill 1.70–1.71×. All of it gated on a byte-exact temperature-0 acceptance hash at every step.

**Overlap to flag.** `antirez/main` gained `rocm: enable DSpark speculative decoding` (+ the DSpark verifier commits) after this PR's base — same problem territory as the verifier work here. We have not evaluated how the two interact; flagging so the rebase order can be chosen deliberately.

jyatesdotdev:
I'm going to stop here, it's fast enough locally for now and I've already spent a significant amount of time on it. Hopefully we can at least get this merged! I've logged everything done via AI here: https://jyates.dev/records/ds4-strix-halo-optimization.

jyatesdotdev:
Here is a snapshot for the resource usage on a test run I did: https://snapshots.raintank.io/dashboard/snapshot/OZOcNCdSiBXxbd9OA3DWxcD9ft9pfdIr?from=2026-08-15T21:28:11.955Z&to=2026-08-15T22:28:11.955Z&timezone=browser&refresh=10s

Note the prefill is tiny because the prompt was tiny. 

Better test for prefil, nearly 200 tok/s prefill. 

https://snapshots.raintank.io/dashboard/snapshot/ezRVtKFlnPK0scfdXrgd6e1S8XTIRdSB

I'm running a loop to see if I can improve prefill even more, but if not, I'll just monitor this PR for @kyuz0 and @antirez, or comments from anyone else. Just let me know if anything needs changing or if you'd like me to run more tests and/or gather data.

Also if anyone is interested in the models I've been using for this; a mix of Fable, Sol 5.6, and Kimi K3.

Also, here is a log of the work if anyone is interested https://jyates.dev/records/ds4-strix-halo-optimization.

jyatesdotdev:
Just did another test.. between 260-300 tok/s prefill! Low context decode at 15 tok/s, but decays quickly to 12.6 tok/s. Seeing if I can fix that, then I'll do one more force push to this branch and keep the changes as limited as possible. I told my agent to rebase which is why you now see a bunch of commits not related to MXFP4 work in this PR now, but I'll fix that.

jyatesdotdev:
# QA results — ROCm MXFP4 PR (restructured branch)

Scope: `upstream/ds4f-mxfp4` (80df56a) + 10 commits, MXFP4-only diff
(9 source files + Makefile). Ran the QA_BEFORE_RELEASES.md sections
applicable to a ROCm routed-MoE kernel PR.

- **Commit under test:** branch tip (post-reword)
- **Hardware:** 2x Strix Halo Framework Desktop (Radeon 8060S, gfx1151),
  ROCm TheRock toolchain
- **GGUF:** `DeepSeek-V4-Flash-MXFP4Experts-F16HC-F16Compressor-F16Indexer-Q8Attn-Q8Shared-Q8Out-chat-v2-mxfp4-0731.gguf` (~79 GB)

## Section 1 — build sanity

- `make clean && make strix-halo` on Strix Halo: **0 warnings, 0 errors**.
- `make cpu` compile check: **0 warnings**.
- `git diff --check`: clean.
- `./ds4 --help` / `--help distributed`: render cleanly. (Distributed
  options live in the CLI at this branch vintage; `ds4-server` has none.)
- Note: the branch base (`ds4f-mxfp4` @ 80df56a) **does not link** on the
  TheRock toolchain without this PR — `9be8ca5` adds `ROCM_HOST_CFLAGS=-fPIC`
  ("make Strix host objects position independent for TheRock linking").
  Verified by building the base tree with and without the flag.

## Section 2 — core regression

`make test-rocm` (new target added in this PR; CUDA-specific
sampling/session binaries are not part of it):

- `ds4-eval --self-test-extractors`: PASS
- `ds4_agent_test`: PASS
- `tests/test_layer_pack`: 97/97
- `tests/test_engine_mgpu_placement`: 98/98
- `tests/test_gpu_args`: PASS
- `q4k-dot-test`, `mxfp4-dot-test`: PASS (4/4)
- `tests/test_gpu_args_cli.sh`: PASS=32 FAIL=12 — **identical on the
  unmodified base** (ROCm builds do not advertise `--gpu-vram`,
  `--gpu-devices`, `--cuda-tensor-parallel`; pre-existing, not a
  regression).
- One pre-existing upstream warning in `tests/test_q4k_dot.c`
  (strict-aliasing), unchanged by this PR.
- `ds4_test` model sections (`--logprob-vectors`,
  `--local-golden-vectors`): **not runnable on a single Strix Halo** —
  the release MXFP4 GGUF (~79 GB) exceeds one node's usable VRAM
  (~65 GiB); full residency requires the two-node topology used below.

## Section 9 — ROCm kernel gate

- `make test-mxfp4-rocm`: **PASS**, zero `failures` for `mid` and `out`
  at 1, 2, 3, 4, 5, 32, 128, 512 tokens (superset of the required
  1/3/32/128/512).
- Repeated at the production tensor shape (`-DMODEL_DIM=7168
  -DN_EXPERT=8`): PASS.
- Occupancy/path variants `DS4_ROCM_ENABLE_MXFP4_TILE4`,
  `DS4_ROCM_ENABLE_MXFP4_ROW64`, `DS4_ROCM_MXFP4_DOWN_RGROUP=4`:
  **bitwise OK** (memcmp against the default path's outputs — same
  accumulation order by construction).

## End-to-end (two-node distributed, full residency)

This model only fits distributed on this hardware, which is also the
production topology:

- PR build, worker (node A, layers 22:output) + coordinator (node B,
  layers 0:21), `--temp 0 --nothink`, 200-token generation: coherent
  output, `prefill: 27.24 t/s, generation: 8.15 t/s` (CLI distributed
  mode, TCP transport, short context).
- The **base branch cannot run this GGUF on ROCm at all**
  (`gpu layer 0 ffn batch encode failed`) — this PR is the enablement,
  so there is no same-hardware ROCm baseline to diff against.
- The base **CPU** backend also rejects the file
  (`unsupported gate/up expert tensor type`), so a CPU golden-reference
  comparison is not available for this checkpoint. The synthetic
  CPU-oracle gate above is the bit-level evidence.
- The §9-named q2-imatrix GGUF is not present on this host; the release
  MXFP4 GGUF was used instead.

## SSD streaming — known pre-existing gap, out of scope

ROCm SSD streaming for Flash-family models is broken at the branch base
for **all** routed quants, before any of these commits:

| Config | base @ 80df56a | this PR |
|---|---|---|
| Q4_K-experts GGUF, `--ssd-streaming` | fails: `missing compact selected experts … full expert table is not mapped` | identical failure |
| MXFP4 GGUF, `--ssd-streaming` | fails: `rocm prefill failed` | runs, but output is wrong |

Root cause is in the ROCm streaming machinery, not the quant kernels:
main already carries the fix class via PR #623 ("ROCm: restore DeepSeek
V4 Flash decode performance and Q4 SSD streaming", merged as 6841f07),
which `ds4f-mxfp4` does not contain. Declared out of scope for this PR;
revisit after the target branch picks up main. One follow-up noted for
that future work: with MXFP4 the broken cache state is computed on
(garbage out) rather than erroring — worth a hard-fail guard.

## Performance evidence

Measured during the campaign on the production pair (same kernels as
this branch):

- Gate/up tile kernel grid order (default-on here): 247.7 → 222 ms/call.
- Occupancy variants (env-gated here): 10.5k-token production probe
  46.7 → 44.2 s prefill (TILE4 + DOWN_RGROUP=4).
- Decode-side structural variants (register prefetch, LDS staging,
  rows-per-warp) were implemented and measured **neutral-to-negative**
  (15.11 t/s baseline vs 14.94 / 11.18 / 14.43) and deliberately
  excluded.

## Sections not applicable to this diff

Metal (4, 5), CUDA (8), GLM (6 — no GLM code touched), distributed
server/KV-disk (10, 11), server APIs (12), agent TUI (13), downloads
(14), DSpark (post-dates the branch base). Speed is within expected
variance for the same hardware/model (§15, per numbers above).

jyatesdotdev:
@antirez, @kyuz0; sorry for all the churn. This PR won't be touched again unless you request changes. I'll be opening PRs for the other performance improvements (if they still hold on main).

jyatesdotdev:
Format A/B benchmark on this exact branch (two-node gfx1151 pair, Thunderbolt, coordinator layers 0:21 / worker 22:output, distributed CLI, identical ~7.5k-token prompt, temp 0, `--transport auto`, no MTP/speculative stack — this branch measures the kernels as shipped here):

| expert format | prefill | decode |
|---|---:|---:|
| **MXFP4 (this PR)** | **162.00 t/s** | **7.86 t/s** |
| Q4_K (same branch, imatrix GGUF) | 89.13 t/s | 7.47 t/s |

**+82% prefill, +5% decode** for the MXFP4 format on identical hardware and settings.

Context for the read-out:

- Prefill is where the MXFP4 expert kernels (this PR) do the work: chunked expert batches are large, the format's smaller weight footprint + the exact-rows kernels' bandwidth profile compound, and the distributed pipeline keeps both hops busy.
- Decode at this vintage is single-token and dominated by dense attention/matmul, so the expert format barely matters — the decode-side story for MXFP4 (speculative k-row spans, MTP verify batching) is deliberately not in this PR; it stacks on top separately.
- Reminder from the QA comment: without this PR the branch cannot run the MXFP4 GGUF on ROCm at all (`gpu layer 0 ffn batch encode failed`), so the A/B's baseline for the format itself is "does not run" — the Q4_K column is the same-branch speed reference, not a like-for-like alternative.
- Bitwise exactness vs the one-row reference holds at both production shapes (tests in the PR; see the QA comment for the full matrix).

jyatesdotdev:
Is anyone else interested in testing this?

jessa0:
> Is anyone else interested in testing this?

I'd love to, I was waiting to see if you'd rebase it on main, however, since I have other patches I don't really want to port to the ds4f-mxfp4 branch.. Can I ask what the exact reason is it's on that branch? Seems like at least some MXFP4 support was merged already, right?

jyatesdotdev:
That's fair, let me see if I can do that, I'll give it a shot this weekend! At the time this seemed like the best branch to target but I'll see where main is now.

jyatesdotdev:
Rebased onto current `main` (`84cc882`) and retargeted: the original base branch `ds4f-mxfp4` has been fully merged into main (zero unique commits remain), so this pull request now applies directly to main. Notes from the rebase:

- All ten commits replayed with only Makefile conflicts. Main independently adopted a two-step link for the quality scorer that matches what this branch carried, so that part of the Makefile commit reduced to adding `$(ROCM_HOST_CFLAGS)` to the compile step. Main still has no MXFP4 code in the ROCm backend, so the kernels remain the enablement.
- Re-verified on the new base, Radeon 8060S (gfx1151): warning-free `make strix-halo`; `make test-mxfp4-rocm` passes at both the synthetic and the production tensor shape (`MODEL_DIM=7168`, `N_EXPERT=8`), including the byte-identical check of the down-projection row-group variant; end to end on the released MXFP4 model file across two machines: coherent output.
- Performance on the new base improved because main's decode work compounds with these kernels: prefill 173.6 tokens/s and decode 13.5 tokens/s, against 162.0 and 7.9 on the old base. Against Q4_K on the same base and prompt (89.5 prefill / 14.0 decode), MXFP4 is 94% faster at prefill with decode parity. The body's performance table is updated accordingly.

jyatesdotdev:
@jessa0 all set; give it a shot!

jyatesdotdev:
Small follow-up fix pushed (`babca3d`): the Makefile inherited from the original base branch had flattened the `ds4_test` and `ds4_agent_test` link rules to an unconditional `$(DS4_LINK)` line. That variable is only set on the CUDA and ROCm invocation paths, so a bare `make test` on macOS would fail to link the test binaries. The rules now restore main's platform guard — `$(CC)` with Metal libraries on Darwin, `$(DS4_LINK)` elsewhere — so default builds behave exactly as main does, and `make test-rocm` is unchanged. Verified: `make ds4_test` links and runs on macOS (Darwin path) and on gfx1151 through the ROCm path (all sections pass up to the long-context test, which needs a `DS4_TEST_MODEL` fixture — same as main).

jessa0:
runs great for me. not sure exactly what perf to expect, but for me it's getting around 200t/s prefill at peak (100% pipeline utilization) and 9.5t/s decode:

```
chat ctx=0..21046:21046 TOOLS prefill chunk 12288/21046 (58.4%) chunk=201.83 t/s avg=142.79 t/s 86.058s
distributed telemetry: request=4 hop=0 layers=20:42 route=0 pos=12288 tokens=2894 eval=14708.974ms downstream_wait=0.000ms forward_send=0.000ms input=180.89MiB output=0.49MiB
```
...
```
chat ctx=21096..21136:40 gen=90 TOOLS DSML_START DSML_END decoding chunk=9.45 t/s avg=9.42 t/s 9.550
```

for comparison, all other runtime switches the same, with q2, i get 280t/s prefill and 10-11t/s decode:

```
chat ctx=0..21046:21046 TOOLS prefill chunk 12288/21046 (58.4%) chunk=279.67 t/s avg=188.89 t/s 65.054s
distributed telemetry: request=4 hop=0 layers=20:42 route=0 pos=12288 tokens=2894 eval=10973.113ms downstream_wait=0.000ms forward_send=0.000ms input=180.89MiB output=0.49MiB
```
...
```
chat ctx=21096..21146:50 gen=100 TOOLS DSML_START decoding chunk=10.73 t/s avg=10.58 t/s 9.455s
```

this is with dspark enabled (which i haven't seen much gain from), just running my normal workload.

(looking at it now, i think my layers are a bit imbalanced? whoops, close enough.)

jyatesdotdev:
Awesome! Yeah with the other PRs I have open was able to get near 300 prefill and 15ish decode. I haven't tried putting it all together after rebasing on main though.

jyatesdotdev:
@antirez or @kyuz0 if you could try this out I'd really appreciate it! It's been rebased off main recently and I'd like to add more improvements to get the performance matching my local setup but I don't want to keep growing this PR.

kyuz0:
> @antirez or @kyuz0 if you could try this out I'd really appreciate it! It's been rebased off main recently and I'd like to add more improvements to get the performance matching my local setup but I don't want to keep growing this PR.

I have pulled this into my toolboxes, and I just kicked off my Terminal Bench Mini run, I use that as a strong quality gate. It's started and it might take 24 hours, after that I'll ping @antirez to merge this.

Thanks!

jyatesdotdev:
Awesome; thanks!

kyuz0:
@jyatesdotdev quality test on Terminal Bench passed with flying colours, just doing a performance test but I'll let @antirez know this is good to merge.

kyuz0:
Actually, what I would suggest is to make these defaults:

```
DS4_ROCM_ENABLE_MXFP4_TILE4=1
DS4_ROCM_MXFP4_DOWN_RGROUP=4
```

As these are the vars I set for the quality and performance testing.

antirez:
Thank you, merged!

PR #659, opened 2026-08-02
Fix DSpark greedy-identity break: replay accepted tokens via single-token decode

### LINKED ISSUE #658: DSpark speculative decode breaks --temp 0 greedy identity (batch-verify vs single-token KV numerics)
## Summary

DSpark speculative decode (`--mtp <support.gguf> --dspark`) does not preserve
byte-identical greedy (`--temp 0`) output vs. non-speculative decode on longer
generations, despite the accept/verify logic itself correctly matching argmax
(independently code-reviewed).

## Root cause

The compressed-KV "compressor frontier" state for every accepted token
(including the always-accepted bonus token) is computed by the verify
batch's own multi-token GEMM, `ds4_gpu_matmul_f16_pair_tensor`
(`ds4.c:27454` and the equivalent decode-batch site around `ds4.c:28685-28753`
on current `main`). This is a numerically distinct code path from ordinary
single-token greedy decode, which computes the identical logical quantity via
a different, fused single-vector kernel,
`ds4_gpu_matmul_f16_pair_compressor_store_tensor` (`ds4.c:22993-23011`).

These two kernels are not required to be, and empirically are not,
bit-identical: a GEMM computing one row of a `[n_tokens x width]` batched
matmul and a fused single-vector kernel computing the same logical `1 x
width` row use different tiling/accumulation order internally (standard
FP16/BF16 GEMM non-associativity across kernel implementations). So every
token committed through DSpark's accept path gets a compressor-frontier
state that is numerically close to, but not identical to, what sequential
single-token decode would have produced for the exact same token. This state
feeds every subsequent layer's attention, so the drift accumulates silently
until it flips an argmax tie -- consistent with the divergence we observed
being a late (tens of tokens in), paraphrase-level flip rather than an
immediate/garbled break.

This is why `spec_frontier_snapshot`/`_restore` (rollback/restore) is not the
culprit: it faithfully snapshots and restores exactly what the batch-verify
pass computed. The bug is upstream of it, in what values get *written* into
the frontier tensors by the accept-commit path in the first place.
Specifically, `ds4_session_eval_dspark_speculative_argmax`'s accept-commit
logic had a fast path for full accepts (`commit_drafts == draft_n`) that
committed the batch-verify-derived KV/compressor-frontier state directly,
instead of re-deriving it via the fused single-token kernel the way plain
decode does.

## Reproduction

Resident IQ2XXS quant of DeepSeek-V4-Flash + a DSpark support GGUF
(lineage-matched drafter, `dspark.target_layer_ids=[40,41,42]`), on CUDA
(NVIDIA GB10 / sm_121):

```
./ds4 --cuda -m <IQ2XXS-model.gguf> --temp 0 -n 400 \
  -p "Explain the TCP three-way handshake."
./ds4 --cuda -m <IQ2XXS-model.gguf> --mtp <DSpark-support.gguf> --dspark \
  --temp 0 -n 400 -p "Explain the TCP three-way handshake."
```

We confirmed this directly against unpatched `main` (`54b36ed`, current tip at
the time of filing): the drafter-enabled run diverges from the no-drafter
baseline within the first couple of sentences of the response (word/phrasing
choices differ from there on, e.g. "...agree on initial sequence numbers and
parameters..." vs. "...agree on initial sequence numbers and synchronize
their connection state...") -- a near-tied paraphrase-level divergence, not
garbled output, consistent with an FP-drift-driven argmax flip rather than a
logic bug. This is a `--temp 0` contract violation, since both runs are
greedy decodes of the same model/prompt and should be byte-identical. (Exact
divergence onset is prompt/pairing-dependent; on our downstream fork's build
of this same drafter pairing the same prompt instead diverged ~55-65 tokens
in -- the divergence is a drift effect, not a fixed offset.)

## Fix

**PR #\<n\> fixes this.** The DSpark accept-commit fast path for full accepts
is removed; every accept (full or partial) now falls through to the
rollback+replay path that already existed for partial accepts, which rolls
the KV/compressor-frontier state back to the pre-verify snapshot and
re-derives each accepted token's state one token at a time through the same
single-token decode kernel plain greedy decode uses
(`metal_graph_eval_token_raw_swa`). This guarantees the post-accept
state -- and the logits read off it -- are bit-identical to what pure decode
would have produced for the same tokens.

Verified byte-identical drafter-vs-no-drafter output on multiple prompts at
both 400 and 800 generated tokens, including the reproducer above extended
past its original divergence point, on both our downstream fork and this
issue's target branch built from current `main`.

Cost: this adds one extra fused single-token forward pass per accepted token,
even on full accepts (previously only partial accepts paid this cost). On our
hardware/pairing this makes an already fairly marginal drafter throughput
case measurably worse (see below) -- correctness first, and the PR states
this plainly rather than hiding it.

## Supporting measurements (informational, not part of the correctness fix)

These are separate, previously-gathered measurements on the same
hardware/pairing (NVIDIA GB10, resident IQ2XXS DeepSeek-V4-Flash + the DSpark
drafter above), offered as data points rather than requests:

- **Throughput, resident model:** no-drafter baseline ~16.87 t/s generation.
  Pre-fix drafter-enabled: ~16.07 t/s default confidence (-4.7%), ~10.81 t/s
  forced-accept. Post-fix (this issue's PR), drafter-enabled generation is
  further reduced to roughly 14.0-14.6 t/s across several prompts (-14% to
  -17% vs. no-drafter) -- the correctness fix's replay-per-accepted-token
  cost applied to full accepts as well as partial ones.
- **Throughput, streamed model (SSD-streaming, different pairing):**
  no-drafter baseline ~5.25 t/s, drafter-enabled ~3.85 t/s default confidence,
  ~1.62 t/s forced-accept.
- **Accepted-depth distribution:** with default confidence (0.9), acceptance
  depth is heavily concentrated at depth 1 (bonus-token-only accepts), with a
  long tail out to depth 8-9 in occasional bursts, rather than a flat/typical
  distribution. We traced this to the drafter checkpoint's own
  `dspark.noise_token_id` metadata deliberately seeding the non-bonus draft
  positions' Markov chain with noise/mask tokens rather than chaining true
  per-position argmax embeddings -- this appears to be intentional per the
  checkpoint format (matching the "greedy argmax-only path" framing in the
  original DSpark work this format is based on), not a wiring gap on our
  side, but it does bound how much the drafter can help on this particular
  drafter/target pairing. Flagging in case it's useful context for anyone
  tuning drafter checkpoints or expected acceptance-depth behavior against
  this codebase.

Happy to share full logs/scripts if useful.

nexus-cw:
## Summary

Fixes #658: DSpark speculative decode did not preserve byte-identical greedy
(`--temp 0`) output vs. non-speculative decode on longer generations.

## Root cause (see linked issue for full detail)

`ds4_session_eval_dspark_speculative_argmax`'s accept-commit logic had a fast
path for full accepts (`commit_drafts == draft_n`) that committed the verify
batch's own batched-GEMM-derived compressor-frontier KV state
(`ds4_gpu_matmul_f16_pair_tensor`) directly. That state is numerically
distinct (standard FP16/BF16 GEMM non-associativity) from what ordinary
single-token decode's fused kernel
(`ds4_gpu_matmul_f16_pair_compressor_store_tensor`) would have produced for
the same token, so it silently drifted from a pure-decode trajectory until it
eventually flipped an argmax tie on longer generations.

The partial-accept path already had the correct mechanism: roll the
KV/compressor-frontier state back to the pre-verify snapshot, then re-decode
each accepted token one at a time through the same single-token kernel plain
decode uses (`metal_graph_eval_token_raw_swa`). It just wasn't used for full
accepts.

## Fix

Minimal diff: removes the full-accept fast path entirely. Every accept (full
or partial) now falls through to the existing rollback+replay path
unconditionally, so post-accept KV state and logits are always re-derived
through the same code path plain greedy decode uses -- bit-identical by
construction, not by numerical luck.

No new code, no new flags, no dialect/quantization changes. Zero effect when
no drafter is configured (`draft_n` is never reached in that case).

```
 ds4.c | 22 +++++++++++++---------  (1 file changed, 22 insertions(+), 51 deletions(-))
```

## Identity-test evidence

Resident IQ2XXS quant of DeepSeek-V4-Flash + a DSpark support GGUF (`--mtp
<support.gguf> --dspark`), `--cuda --temp 0`, drafter-enabled vs no-drafter,
byte-for-byte `diff`, built from this branch (based on current `main`):

| prompt | tokens | result |
|---|---|---|
| "Explain the TCP three-way handshake." | 400 | byte-identical |
| "Write a detailed explanation of how the TCP three-way handshake..." (longer prose prompt) | 800 | byte-identical |

Both prompts were confirmed to diverge on unpatched `main` before this fix
(the second one is the original reproducer from the linked issue, extended
past its divergence point here). No divergence observed post-fix in either
case.

## Overhead (stated plainly)

This fix makes every accepted token -- including full accepts, which
previously skipped the replay -- pay for one extra fused single-token
forward pass. On our hardware/pairing (NVIDIA GB10, resident IQ2XXS +
this drafter), that increases drafter-enabled generation slowdown vs.
no-drafter from about -5% to roughly -14% to -17% across a few prompts
(single runs, not a full statistical sweep). This is the honest cost of
correctness here: the alternative (a narrower fix that re-derives only the
compressor-frontier projection instead of the full forward pass) is possible
in principle but a larger, riskier change, and wasn't attempted in this PR --
happy to discuss as a follow-up if there's interest.

## Testing

`make cuda-spark`: clean rebuild, zero warnings, on this branch (based on
current `main`).

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -61017,57 +61017,28 @@ static int ds4_session_eval_dspark_speculative_argmax(
         }
     }
 
-    bool final_logits_ok = false;
-    if (ok && commit_drafts == draft_n) {
-        const double read_t0 = stats_enabled ? now_sec() : 0.0;
-        final_logits_ok =
-            metal_graph_read_spec_logits_row(&s->graph,
-                                             (uint32_t)(draft_n - 1),
-                                             row_logits);
-        if (stats_enabled) {
-            s->dspark_stats.verify_read_ms += (now_sec() - read_t0) * 1000.0;
-        }
-    }
-
-    if (ok && commit_drafts == draft_n && final_logits_ok) {
-        if (tp_verify_sent &&
-            !ds4_tp_send_verify_commit(e->tp.ctx, 1, 0)) {
-            snprintf(err, errlen, "tp: verify commit send failed");
-            s->checkpoint_valid = false;
-            spec_frontier_free(&frontier);
-            DS4_DSPARK_STATS_FINISH();
-            return -1;
-        }
-        memcpy(s->logits, row_logits, (size_t)DS4_N_VOCAB * sizeof(s->logits[0]));
-        int emitted_drafts = 0;
-        for (int i = 0; i < draft_n && n_accept < accepted_cap; i++) {
-            accepted[n_accept++] = drafts[i];
-            emitted_drafts++;
-            if (drafts[i] == eos_token) break;
-        }
-        s->checkpoint_valid = true;
-        ds4_session_dspark_capture_note_checkpoint(s);
-        if (stats_enabled) {
-            s->dspark_stats.full_accepts++;
-            s->dspark_stats.accepted_draft_tokens += (uint64_t)emitted_drafts;
-            ds4_dspark_stats_note_len(s->dspark_stats.accepted_len_hist,
-                                      (uint32_t)emitted_drafts);
-        }
-        ds4_session_dspark_scheduler_note(
-                s,
-                (uint32_t)emitted_drafts,
-                false,
-                DS4_DSPARK_SCHED_EXTRA_MS());
-        if (spec_log) {
-            fprintf(stderr,
-                    "ds4: DSpark spec accept drafted=%d accepted=%d\n",
-                    draft_n,
-                    n_accept);
-        }
-        spec_frontier_free(&frontier);
-        DS4_DSPARK_STATS_FINISH();
-        return n_accept;
-    }
+    /* Greedy-identity fix: the verify batch's compressor-frontier writes
+     * come from a batched multi-token GEMM (ds4_gpu_matmul_f16_pair_tensor),
+     * which is numerically distinct from ordinary single-token decode's
+     * fused projection+store kernel
+     * (ds4_gpu_matmul_f16_pair_compressor_store_tensor). Previously, a full
+     * accept (commit_drafts == draft_n) took a fast path that committed the
+     * batch-verify's KV/compressor-frontier state directly, so accepted
+     * tokens' state silently diverged from what plain decode would have
+     * produced -- this eventually flips an argmax tie on longer
+     * generations, breaking --temp 0 greedy-identity vs. no-drafter decode.
+     * There is no separate fast-accept branch any more: every accept (full
+     * or partial) falls through to the rollback+replay path below, which
+     * re-derives each accepted token's KV/compressor state one token at a
+     * time through the same single-token decode kernel
+     * (metal_graph_eval_token_raw_swa) plain greedy decode uses, so the
+     * post-accept state -- and the logits read off it -- are bit-identical
+     * to pure decode for the same tokens. This costs one extra fused
+     * single-token kernel launch (plus its surrounding forward pass) per
+     * accepted token per block; on a net-loss drafter workload this makes
+     * DSpark's already-negative throughput case worse, but correctness
+     * comes first. When there is no drafter, draft_n is never reached (the
+     * caller returns earlier), so this path has zero effect. */
 
     if (verifier_may_have_mutated) {
         s->checkpoint.len = start;


raveonetter:
Independent M4 Max follow-up on the narrower-repair idea mentioned in this PR (re-derive only the compressor-frontier projection instead of full replay): we now have runtime first-divergence evidence showing that compressor-only repair is not sufficient for strict decode-order identity.

Using the same true pre-verify state and the same forced proposal tokens, with an in-stream GPU snapshot harness whose non-perturbation control passes (`A0==A1`, `A0==A2` bitwise), the earliest observed generic-vs-ordinary-sequential divergence is already in layer 0 / row 0 at the Q_A projection output:

```text
CP1 normalized_attention_input: EXACT (4096 F32)
CP2-Q q_a_projection_output: MISMATCH (823/1024)
max_abs=5.2154064178466797e-08

generic:
  metal_graph_matmul_q8_0_named_tensor_attn_q_a

sequential:
  metal_graph_matmul_dense_quant_tensor_attn_q_a
```

An isolated same-input/same-weight primitive A/B reproduces the numerical non-equivalence. Then, under a diagnostic gate, making only generic Q_A use canonical sequential-equivalent arithmetic removes that divergence completely (`CP2-Q: EXACT`) while the probe controls remain PASS.

With Q_A canonicalized, the next first divergence moves immediately to:

```text
row=0 layer=0 checkpoint=CP2-KV-P subobject=kv_raw
```

This is the semantic raw-KV projection output *before* persistent raw-cache interpretation/store, so this mismatch cannot be attributed to ring-row mapping, the persistent store, compressor update, snapshot/restore, or later operators.

Current evidence:

```text
QA_CAUSAL_SOURCE              PROVEN BY TEST
QA_ONLY_SOURCE                DISPROVEN
NEXT_INDEPENDENT_DRIFT_SOURCE CP2-KV-P / kv_raw
KV_PRIMITIVE_ROOT_CAUSE       NOT YET PROVEN
```

So a narrower alternative to full replay still looks possible, but the repair set is already broader than compressor-frontier re-projection alone.

The next question is whether Q_A, raw-KV, and subsequent sites collapse to one reusable quantized-projection arithmetic family (batched vs single-row/fused), or whether independent drift appears across multiple operator families.

We are iteratively canonicalizing one proven source at a time and rerunning whole-model first-divergence to identify the minimal sufficient canonical repair set, rather than sequentializing the verifier wholesale.

nexus-cw:
Independent corroboration from CUDA/GB10: our temp-0 probes found the same batched-vs-single-row quantized-matmul non-equivalence as your CP2-Q result, which is what motivated full replay in this PR — it's repair-set-complete by construction, at the cost of the extra decode pass. Your first-divergence data (Q_A canonicalization immediately exposing kv_raw) matches our experience that the drift sources span operator families rather than one site. If you do converge on a minimal canonical set, the win over replay is real, but it inherits a maintenance invariant: every new or changed batched kernel must be re-proven equivalent or identity silently breaks. A first-divergence CI probe like your harness would make that invariant enforceable.

PR #692, opened 2026-08-05
server: preserve reasoning_content token boundaries

### LINKED ISSUE #685: ds4-server reasoning_content deltas fragmented mid-token at arbitrary byte offsets (OpenAI-compatible streaming)
## Summary
Streaming `/v1/chat/completions` emits `reasoning_content` deltas split at arbitrary byte offsets — mid-token, character-level — instead of whole tokens. Reassembled reasoning text is correct and the final `content` answer streams intact, so it's cosmetic for reassembly. But any client that consumes thinking deltas individually (reasoning-length estimation, per-token UIs, signature/checksum checks) gets garbage units, and every delta carries SSE+JSON overhead (wasted tokens).

## Environment
- ds4-server from antirez/ds4, DeepSeek V4 Flash GGUF (`ds4flash.gguf`)
- Launch: `ds4-server --ctx 262144 --port 8181 --model ./ds4flash.gguf`
- Client: live curl to `127.0.0.1:8181/v1/chat/completions`, `stream: true`

## Reproduction / observed
Live request to the local endpoint shows the reasoning stream fragmented at character level:
```
"1.  T" → "he use" → "r a" → "sk" → "ed " → "to \"sa" → "y he" → ...
```
Every delta carries empty `"signature":""`. Reassembled text is correct ("1. The user asked to say hello..."); final answer `content` streams intact.

## Expected
Reasoning deltas aligned to token boundaries (one delta per decoded token), mirroring the `content` path.

## Root cause (in ds4-server, not Goose)
`openai_sse_stream_update()` in `ds4_server.c`, `OPENAI_STREAM_THINKING` branch (~line 6385):
```c
const size_t hold = strlen("</think>") - 1;
limit = raw_len > hold ? raw_len - hold : st->emit_pos;
limit = utf8_stream_safe_len(raw, st->emit_pos, limit, false);   // UTF-8 safe only
sse_chat_delta_n(fd, r, id, "reasoning_content", raw + st->emit_pos, limit - st->emit_pos);
```
Delta boundaries = accumulated text minus a 7-byte hold-back, clamped only to UTF-8 codepoint boundaries — never token-aligned. Identical shape in the Responses path (~7094) and Anthropic path (~7971). Regression test `test_openai_chat_stream_splits_reasoning_without_tools` (~13973) locks in current behavior.

## Impact
- Clients parsing thinking deltas individually (agent harnesses, reasoning-length counters, stream UIs) see mid-token garbage.
- Per-delta JSON overhead inflates output; "wastes tokens" vs whole-token deltas.
- Harmless for plain chat reassembly.

## Related issues (routing family — distinct from this granularity bug)
- #13, #509, #524, #665 (thinking→reasoning_content routing; #509 explicitly states streaming is "correct" re: labeling)
- #678 (close-tag handling in streaming TEXT mode)
- #681 (don't stop generation inside a thinking block)
- #167, #332, #612, #318 (unclosed/leaked thinking boundaries)

## Suggested fix
Accumulate reasoning per decode step and flush one delta per token (as `content` does), or hold back until the next token boundary; keep `utf8_stream_safe_len` as a floor, not the boundary rule.

rareba:
## Summary

- keep OpenAI reasoning_content SSE deltas aligned to decoded-token boundaries
- retain one-token look-behind while checking for the closing think marker
- replace the regression that expected arbitrary byte splitting

## Root cause

The thinking stream held a fixed seven-byte suffix to detect a possibly split
closing think marker. Every update therefore emitted the current decoded token
minus seven bytes, then emitted that tail with the next token.

Remembering the previous decoded-token end preserves whole-token deltas while
keeping the close-tag and UTF-8 safeguards intact. The only cost is one token
of reasoning-stream latency.

## Impact

Clients that process individual reasoning deltas now receive stable token
units instead of character fragments. Reassembled output is unchanged.

## Validation

Windows 11 host, WSL Linux CPU/reference build; no model required:

- focused server test binary: ds4 tests: ok, server: OK
- make -j2 cpu

Closes #685

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -5677,6 +5677,7 @@ typedef struct {
 typedef struct {
     openai_stream_mode mode;
     size_t emit_pos;
+    size_t reasoning_token_end; /* hold newest token while checking </think> */
     bool active;
     bool checked_think_prefix;
     bool guard_second_reasoning;
@@ -6477,6 +6478,7 @@ static bool openai_sse_stream_update(int fd, server *s, const request *r, const 
             const char *open = "<think>";
             const size_t open_len = strlen(open);
             if (raw_len < open_len && !strncmp(raw, open, raw_len) && !final) {
+                st->reasoning_token_end = raw_len;
                 return true;
             }
             if (raw_len >= open_len && !strncmp(raw, open, open_len)) {
@@ -6500,8 +6502,8 @@ static bool openai_sse_stream_update(int fd, server *s, const request *r, const 
         } else if (final) {
             limit = raw_len;
         } else {
-            const size_t hold = strlen("</think>") - 1;
-            limit = raw_len > hold ? raw_len - hold : st->emit_pos;
+            limit = st->reasoning_token_end;
+            if (limit < st->emit_pos) limit = st->emit_pos;
             limit = utf8_stream_safe_len(raw, st->emit_pos, limit, false);
         }
 
@@ -6518,6 +6520,7 @@ static bool openai_sse_stream_update(int fd, server *s, const request *r, const 
                 st->emit_pos = (size_t)(tool - raw);
                 st->mode = OPENAI_STREAM_SUPPRESS;
             }
+            st->reasoning_token_end = raw_len;
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


Flor1an-B:
Nice diagnosis and fix. Since your diff is scoped to the OpenAI path, opened #712 to apply the exact same `reasoning_token_end` technique to `responses_sse_stream_update()` and `anthropic_sse_stream_update()` — the other two places #685 pointed at with the identical shape. Not touching your diff at all, just extending the same idea to the rest of the file.

PR #699, opened 2026-08-05
Add numbered paste placeholders and true multi-row input

kk1987:
This upgrades the line editing of the bundled linenoise fork, shared by the `ds4` REPL and the `ds4-agent` TUI, in two related ways: Claude Code-style paste placeholders, and real multi-row rendering of hard newlines.

**Real multi-row input.** The editor previously modeled the input as one logical row: a hard newline that reached the terminal (typed with Ctrl+J, or recalled from multiline history) moved the cursor down without a carriage return (raw mode clears OPOST) and desynchronized the refresh row math, stair-stepping and corrupting the prompt. `refreshMultiLine` now derives its geometry from a newline-aware walk of the rendered text: every hard newline paints as a real terminal row (written as CR LF), and the row, cursor, and clear accounting is exact, including the deferred-wrap state terminals keep when text ends on the right margin. Ctrl+J is now a documented way to type a newline. In the paths confined to a single row — single-line mode and masked input — newlines render as a one-column `↵` marker (U+21B5, East Asian Neutral, exactly one column everywhere).

**Short pastes go inline.** A paste of up to 5 lines and under 200 bytes is inserted directly and shows as real rows. Previously any paste containing a newline was folded behind a placeholder.

**Long pastes collapse to a numbered placeholder.** `[Pasted text #1 +123 lines]` (or `+N chars` for a long single-line paste). Numbers are assigned in creation order per edited line and never renumbered, so `#2` still names the same text after `#1` is gone. History folds keep the old unnumbered `[... N lines ...]` text. The newest placeholder carries a `(paste again to expand)` suffix that disappears as soon as the buffer content changes or the line is submitted.

**Paste again to expand.** Pasting bytes that exactly match a fold reveals that text in place as real rows, fully editable, instead of inserting a duplicate. A screen-height guard measures the true expanded row count first: a big file pasted twice by mistake beeps and keeps the placeholder rather than pushing the prompt past the terminal.

**No silent drops.** Running out of the 16 fold slots used to discard the paste with only a beep; the text now goes in unfolded. The only refusals left are a paste over `PASTE_MAX_BYTES` or an edit buffer that genuinely cannot grow.

Enter semantics are unchanged: folds were always display-only and the real bytes are submitted exactly as before, so nothing downstream of the line editor changes.

Testing: `tests/test_linenoise_paste.c` drives the non-blocking editor API with a fake terminal (`LINENOISE_ASSUME_TTY`, pipe stdin, tmpfile stdout so emitted bytes can be asserted) — 81 checks covering the newline geometry (including a differential test asserting the new math reproduces the previous formulas exactly on newline-free input, so wrapping behavior for ordinary lines is provably unchanged), CR LF emission, marker rendering in single-line mode, numbered placeholders, expand/re-fold, the screen-height guard, fold-slot exhaustion, the >1MB refusal, history folds, and the transient hint. Wired into `make test`; `make ds4 ds4-agent` builds warning-free. Manual terminal checks were added to `QA_BEFORE_RELEASES.md`.

https://claude.ai/code/session_01UazT9herrhVF4gFz3VwSar

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@
 /tests/test_cuda_session_batch
 /tests/cuda_long_context_smoke
 /tests/test_layer_pack
+/tests/test_linenoise_paste
 /tests/test_metal_session_batch
 /tests/test_mxfp4_cuda
 /tests/test_mxfp4_dot
--- a/Makefile
+++ b/Makefile
@@ -346,6 +346,14 @@ tests/test_layer_pack.o: tests/test_layer_pack.c ds4_layer_pack.h
 tests/test_layer_pack: tests/test_layer_pack.o ds4_layer_pack.o
 	$(CC) $(CFLAGS) -o $@ $^ $(LDLIBS)
 
+tests/test_linenoise_paste.o: tests/test_linenoise_paste.c linenoise.c linenoise.h
+	$(CC) $(CFLAGS) -I. -c -o $@ $<
+
+# No linenoise.o here: the test includes the implementation to reach the
+# rendering internals.
+tests/test_linenoise_paste: tests/test_linenoise_paste.o
+	$(CC) $(CFLAGS) -o $@ $^ $(LDLIBS)
+
 tests/test_gpu_args.o: tests/test_gpu_args.c ds4_gpu_args.h ds4_gpu_mgpu.h
 	$(CC) $(CFLAGS) -I. -DDS4_NO_GPU -c -o $@ $<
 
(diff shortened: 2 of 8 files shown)


kk1987:
Since opening this PR, two more commits landed (`ed34429`, `3a43c24`): unfolded newlines are no longer shown as a `↵` marker in multi-line mode — `refreshMultiLine` now paints them as real terminal rows with exact geometry, which also makes Ctrl+J a proper, documented way to type a multiline prompt. The marker remains only in single-line mode and masked input. The description above has been updated to match, and the test suite grew from 46 to 81 checks (including a differential test proving the new row math is byte-for-byte equivalent to the old formulas on newline-free input).

https://claude.ai/code/session_01UazT9herrhVF4gFz3VwSar

PR #703, opened 2026-08-05
fix: support MXFP4 in mixed GGUF splicer

### LINKED ISSUE #693: gguf-tools splicer cannot read the MXFP4 GGUF this repo publishes (unsupported GGML tensor type 39)
The Flash MXFP4 GGUF (`ds4f-mxfp4`) cannot be used as a donor or base with `gguf-tools/mixed/splice_mixed_expert_layers_gguf.py`, because its `GGML_QUANT_SIZES` table has no entry for type 39.

```
error: unsupported GGML tensor type 39; add it to GGML_QUANT_SIZES
```

Everything else in the splicer already handles it correctly — it deliberately does not require base and donor tensor *types* to match, and it writes each tensor's own `ggml_type` through. Only the block-size table is missing an entry.

One-line fix, using the values from ds4's own table at `ds4.c:2052` (`[39] = {"mxfp4", 32, 17}`):

```python
 GGML_QUANT_SIZES = {
     ...
     26: (1, 4, "I32"),
+    39: (32, 17, "MXFP4"),
 }
```

Verified locally with `--dry-run`, splicing MXFP4 routed experts for layers 25-42 onto the IQ2XXS/Q2_K hybrid base:

```
selected donor tensors: 54
selected donor types: MXFP4:54
base tensor payload:   90.88 GiB
mixed tensor payload: 107.76 GiB
```

Motivation, in case it is useful: this produces a quant sized for a single 128 GB Mac. The published Flash quants are 80.8 / 90.9 GiB (fit, leaving ~21 GiB of the 112 GiB wired limit unused) then jump to 145.3 / 153.3 GiB (do not fit). Splicing MXFP4 experts into the upper layers fills that gap at 107.76 GiB.

Happy to open a PR if useful.

Berserk-hub150:
## Summary

Add GGML tensor type 39 to the mixed GGUF splicer's quantization-size table, allowing it to parse MXFP4 tensors.

Fixes #693

## Validation

- `py -m py_compile gguf-tools/mixed/splice_mixed_expert_layers_gguf.py`
- Verified that 32 MXFP4 elements occupy 17 bytes.
- Verified that 64 MXFP4 elements occupy 34 bytes.
- Verified that type 39 is displayed as `MXFP4`.

## Environment

- Windows
- Python 3.12

A full dry run with the published GGUF files was not performed locally because of their size.

### DIFF
--- a/gguf-tools/mixed/splice_mixed_expert_layers_gguf.py
+++ b/gguf-tools/mixed/splice_mixed_expert_layers_gguf.py
@@ -61,6 +61,7 @@ GGML_QUANT_SIZES = {
     12: (256, 144, "Q4_K"),
     16: (256, 66, "IQ2_XXS"),
     26: (1, 4, "I32"),
+    39: (32, 17, "MXFP4"),
 }
 
 EXPERT_TENSOR_RE = re.compile(r"^blk\.(\d+)\.ffn_(gate|up|down)_exps\.weight$")


Berserk-hub150:
@antirez, would you prefer this one-line type-table update to remain as-is, or should I also add a small regression test covering MXFP4 tensor-size calculation?

Issue #660, opened 2026-08-02
New 0731 iq2 thinks forever!


arkham000:
Hi, I downloaded DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix-0731.gguf from antirez HF repo, it works quite well aside the facts that it thinks forever. 
I'm using it through pi agent using the exact configuration from antirez which I previously used for the preview version, but thinking now has gone mad, I currently set it to low (but I seen no difference from medium or high) and to just update a bunch of files on my llm-wiki it generated 9k tokens of reasoning!
It goes all the the time _Mhh wait let me reconsider..._ it became qwen3.5 all of a sudden lol

Have you observed the same problem? Are there any workaround ? 
Like this it is unusable, I'm on a m2 max with 96GB and between the low inference speed and infinite thinking it is very painful to use ....
Thanks

TipKnuckle:
Counterintuitively you should try setting the reasoning to 'max'. I can't find it now, but on one of the LLM subreddits there was thread were a user did some testing and found this version of DeepSeek Flash 4 did a ton more reasoning in 'low' compared to 'max.' The model is working well in my experience over the past few days.

arkham000:
Yeah I found that post thanks. I'll try it

Flor1an-B:
Found the likely root cause while looking into reasoning_effort more broadly (I have an open PR — #704 — for a related but separate bug in the "high"/"max" tiers).

Checked against DeepSeek-V4's own encoding spec (`encoding/README.md`): it defines exactly three `reasoning_effort` levels, and `"low"` is the **default** with **no prompt prefix at all** — it's meant to be the model's normal, unprompted reasoning behavior. But DS4's parser currently maps `"low"` (along with `"medium"`/`"minimal"`) to the *same* internal tier as an explicit `"high"` request — the most elaborate non-max prompt DS4 has ("You MUST be very thorough... rigorously stress-testing your logic against all potential paths, edge cases, and adversarial scenarios..."). So setting `reasoning_effort` to `"low"` was actually asking the model to be maximally thorough, which is a very plausible explanation for the runaway reasoning you saw — not a quant-specific quirk.

(The "max sometimes reasons less than low" observation upthread lines up with this too: before my other PR's fix, "max" was *also* misrendering as the "high" text, so both settings were likely producing near-identical prompts either way.)

[...]

Issue #653, opened 2026-08-02
DeepSeek-V4-Flash-0731: the -imatrix-0731 quants ARE 0731-calibrated, byte-exact reasoning-effort prefixes, and a max_tokens warning


rell666:
Two corrections and one practical warning, all checkable from published artifacts.

### 1. antirez's 0731 quants ARE calibrated on 0731 — the `.dat` just wasn't uploaded

The "imatrix is missing" / "not yet recalibrated for 0731" caveat has been repeated a few times in this thread. It's wrong, and the files say so themselves. Read straight out of the GGUF headers over HTTP range requests (no download needed):

| file | `quantize.imatrix.file` | `chunks_count` |
|---|---|---|
| `…-chat-v2-imatrix.gguf` (preview) | `…/DeepSeek-V4-Flash-chat-v2-routed-moe-ds4-1p5m-fixed.dat` | 90,042 |
| `…-chat-v2-imatrix-0731.gguf` | `gguf/DeepSeek-V4-Flash-0731-chat-v2-routed-moe-ds4.dat` | **202,100** |

All three `-imatrix-0731` builds (IQ2XXS, Q4K, Layers37-42) name the 0731 `.dat`, with `entries_count = 129` = 43 layers × {gate, up, down}, i.e. full coverage. The MXFP4 build carries no `quantize.imatrix.*` keys at all, which is expected for a scale-block format.

Corroborating arithmetic: an imatrix build is exactly **288 bytes** larger than its non-imatrix twin (the four extra KV entries). `86,720,111,488 − 86,720,111,200 = 288`, and the same for the Q4K pair.

So antirez recollected at **2.24× the preview's calibration depth** and simply didn't publish the `.dat`. The fair criticism is transparency, not calibration — worth knowing before anyone switches away assuming his upload is stale. Of the 0731 ds4-format quants I've checked, his has the deepest calibration:

| repo | chunks | calibrated on |
|---|---|---|
| antirez `-0731` (×3) | **202,100** | 0731 (`.dat` unpublished) |
| ox-ox | 117,777 | 0731 (`.dat` published) |
| Rednalreden | 117,777 | 0731 |
| jmilnz | 2,752 | 0731 |
| schlaflos, nazeshinjite | 90,042 | preview (documented) |

This matters more than usual because at IQ2_XXS a calibration mismatch is roughly 18–21× costlier than at Q4_K_M, and IQ2_XXS has no no-imatrix fallback — llama.cpp refuses to quantize without one.

[...]

Issue #635, opened 2026-07-31
DeepSeek-V4-Flash-0731 support


daaain:
The benchmark numbers look incredible: https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash-0731

According to the model card it's the same structure so supporting might be a matter of quantising the new weights?

sleepless:
We should be able to do it ourselves: https://github.com/antirez/ds4/blob/main/gguf-tools/README.md#generate-q2-and-q4-ggufs
Edit: I'm on it. I can publish it when done.

minsley:
wow... this quadrant hardly existed a month or two ago, and now we basically have opus4.6@Home .......

<img width="930" height="597" alt="Image" src="https://github.com/user-attachments/assets/0b6d9c15-6a5c-4267-8cc0-fd75d985a693" />

sgmihai:
I am looking for way to run this at a decent speed with 4*3090 and 256g ram. seems ds4 does not support multi gpu. There is a fork with a patch but not merged and it is not maintained with upstream.
llamacpp also has poor support, no dspark

MauriceHagemeijer:
I just made a q2 version of 0731, comparing it now vs the old q2 quant.

RikuKunMS2:
Any place that you've uploaded it to?

> I just made a q2 version of 0731, comparing it now vs the old q2 quant.

Jordan-Milner:
I'm a noob at this stuff, never uploaded a model, but I'm mid-testing this on my m5 max 128gb. Uploaded here: https://huggingface.co/jmilnz/DeepSeek-V4-Flash-0731-Layers37-42Q4K-mixed-GGUF

MauriceHagemeijer:
Not yet, verifying first if the works correctly. Running the built in benchmark old vs new. If it improved I can upload it

sleepless:
> I just made a q2 version of 0731, comparing it now vs the old q2 quant.

Perfect, I am doing the q2-q4-imatrix version. Upload soon.

actualwitch:
For everyone trying to make new quants, remember it's probably a good idea to first regenerate imatrix dataset and then imatrix itself for best results, since trajectories probably changed between releases.

MauriceHagemeijer:
> For everyone trying to make new quants, remember it's probably a good idea to first regenerate imatrix dataset and then imatrix itself for best results, since trajectories probably changed between releases.

[...]

PR #704, opened 2026-08-05
Fix reasoning_effort tier mapping and context-gate misreading

Flor1an-B:
Follows up on #635 (@kernelzeroday, @zom-2018) and #653 (@rell666) — this is the patch discussed there, posted as a comment first (https://github.com/antirez/ds4/issues/635#issuecomment-5192049609) and opened as a PR per @darkbasic's question.

## Two independent bugs, same two functions

1. **`DS4_THINK_HIGH` never rendered a prefix at all.** Only `DS4_THINK_MAX` did, using text that is actually the official **"high"** tier prefix per `deepseek-ai/DeepSeek-V4-Flash-0731`'s `encoding/README.md` (verified verbatim against the primary source). Every `reasoning_effort` other than `"max"` silently rendered as official **"low"**. The real `"max"` prefix (*"Beyond maximum — exhaustive, relentless, and uncompromising..."*, note the em dash U+2014) was absent from the codebase entirely — unreachable via the API regardless of setting. Fixed in both independent prompt-rendering paths that had this bug: `ds4.c`'s `chat_push_think_prefix()` (CLI/engine) and `ds4_server.c`'s `render_deepseek_chat_prompt_text()` (server).

2. **`ds4_think_mode_for_context()` downgraded `"max"`→`"high"` below a 384K *context* threshold**, citing DeepSeek's guidance as an input-context requirement. The model card's actual text: *"For the `high` and `max` reasoning effort levels, we recommend a maximum **output length** of 384K tokens."* That's an output-budget guideline, not a context-size precondition. Turned into a documented no-op rather than removed outright, since four `ds4_server.c` call sites use it and `ds4_think_max_min_context()` may still be useful for client-side output-length advice.

## Testing

- Adds `test_render_think_high_prompt_prefix()` — there was previously zero test coverage of the "high" tier ever rendering anything.
- Fixes the now-incorrect context-gate assertion in `test_reasoning_effort_mapping()`.
- Adds a mutual-exclusivity assertion to the existing `"max"` prompt test.
- `make ds4_test` passes clean against the full existing suite plus these changes.
- Smoke-tested live against a real 0731 quant: `reasoning_effort: high` and `: max` now produce visibly distinct prompts; `: max` no longer collapses to the "high" text below 384K context.

## Requirements

- I have read and agree with the contributing guidelines.
- AI usage disclosure: YES — diagnosed and patched with Claude Code, verified against the primary source (DeepSeek's own model card) and the project's existing test suite.

---

## Update: a real DS4_THINK_LOW tier (issue #660)

Found while looking for other reasoning_effort-related community reports to help with: issue #660 ("New 0731 iq2 thinks forever!") describes `reasoning_effort: "low"` producing 9k tokens of runaway reasoning.

Root cause, again verified against DeepSeek-V4's own encoding spec (`encoding/README.md`, "Reasoning Effort" section): it defines exactly three levels — `"low"` (the **default**, prompt prefix **"none"**), `"high"`, and `"max"`. DS4's `parse_reasoning_effort_name()` collapsed `"low"` (and `"medium"`/`"minimal"`) into the same `DS4_THINK_HIGH` enum value as an explicit `"high"` request — the code even had a comment acknowledging this: *"DS4 only exposes HIGH and MAX above zero, so minimal collapses to the smallest non-zero level (HIGH)."* So a client asking for `"low"` got the **most** elaborate non-max prompt DS4 has, instead of the model's own default behavior with no extra prompting — a very plausible direct cause of runaway reasoning on tasks that don't need it.

Added `DS4_THINK_LOW` as a genuine fourth `ds4_think_mode` value ("thinking enabled, no prompt prefix", distinct from `DS4_THINK_NONE` which disables thinking entirely) and wired it through every switch/branch over the enum. The default `reasoning_effort` when a client sends none at all is untouched (still high); this only fixes what an *explicit* `"low"`/`"minimal"` request does.

Tested the same way as the rest of this PR: full suite passes, new regression test (`test_render_think_low_prompt_prefix`) plus updated `test_reasoning_effort_mapping`, and a live A/B on the production Flash quant confirming `reasoning_effort: "low"` and `"high"` now produce genuinely different `reasoning_content` (they were byte-identical prompts before this fix).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1290,10 +1290,14 @@ the saved prefix instead of processing the whole prompt again.
 ## Thinking Modes
 
 DeepSeek V4 Flash has distinct non-thinking, thinking, and Think Max modes.
-The server defaults to thinking mode. `reasoning_effort=max` requests Think
-Max, but it is only applied when the context size is large enough for the model
-card recommendation; smaller contexts fall back to normal thinking. OpenAI
-`reasoning_effort=xhigh` still maps to normal thinking, not Think Max.
+The server defaults to thinking mode. `reasoning_effort` selects among three
+prompt-level tiers: `low` keeps thinking enabled but adds no effort prefix, so
+the model reasons at its own pace; `medium`, `high`, and `xhigh` are synonyms
+that all request the model card's "high" reasoning-effort prefix; `max`
+requests Think Max, the model card's most exhaustive tier. `max` is honored at
+any context size — the model card's 384K figure is a recommended output-length
+cap for the high/max tiers, not a context precondition, so it is no longer used
+to downgrade `max` to `high` on smaller contexts.
 
 For direct replies, use `thinking: {"type":"disabled"}`, `think:false`, or a
 non-thinking model alias such as `deepseek-chat`.
--- a/ds4.c
+++ b/ds4.c
@@ -382,14 +382,30 @@ int         g_gpu_peer_ok[DS4_MAX_GPUS][DS4_MAX_GPUS];
 #define DS4_DEFAULT_COMPRESS_ROPE_FREQ_BASE (160000.0f)
 #define DS4_DEFAULT_ROPE_ORIG_CTX       UINT64_C(65536)
 
-static const char DS4_REASONING_EFFORT_MAX_PREFIX[] =
+/* Official DeepSeek-V4 "high" reasoning_effort prefix (deepseek-ai/DeepSeek-V4-Flash-0731,
+ * encoding/README.md). This was previously misnamed *_MAX_PREFIX and was the only
+ * prefix ever injected (for DS4_THINK_MAX), leaving DS4_THINK_HIGH with no prefix
+ * at all -- i.e. every non-"max" request silently rendered as official "low". */
+static const char DS4_REASONING_EFFORT_HIGH_PREFIX[] =
     "Reasoning Effort: Absolute maximum with no shortcuts permitted.\n"
     "You MUST be very thorough in your thinking and comprehensively decompose the problem to resolve the root cause, rigorously stress-testing your logic against all potential paths, edge cases, and adversarial scenarios.\n"
     "Explicitly write out your entire deliberation process, documenting every intermediate step, considered alternative, and rejected hypothesis to ensure absolutely no assumption is left unchecked.\n\n";
 
-/* DeepSeek recommends Think Max only with at least a 384K-token context window.
- * Below that size we keep ordinary thinking to avoid injecting a prompt that
- * asks for a reasoning budget the allocated context is not meant to hold. */
+/* Official DeepSeek-V4 "max" reasoning_effort prefix. Was previously unreachable:
+ * DS4_THINK_MAX rendered the "high" text above instead. Note the em dash (U+2014)
+ * right after "Beyond maximum" -- verified byte-for-byte against the primary source. */
+static const char DS4_REASONING_EFFORT_MAX_PREFIX[] =
+    "Reasoning Effort: Beyond maximum \xe2\x80\x94 exhaustive, relentless, and uncompromising.\n"
+    "You MUST reason with the utmost depth and rigor, leaving absolutely nothing to chance: exhaustively decompose the problem into its most fundamental components, trace every causal chain to its root, and resolve the underlying cause rather than any surface symptom.\n"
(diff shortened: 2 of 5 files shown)


GreatMark:
Running this PR (both commits, including the DS4_THINK_LOW tier from the follow-up) rebased onto current `main` (`84cc882`) on M3 Ultra Metal surfaced two small compatibility gaps with code that landed upstream after this branch was opened. Sharing the fixup that made the rebase clean, in case it saves a round-trip when this gets refreshed:

1. **`ds4_cli.c`: `repl_glm_reasoning_effort_text()`** — upstream added this switch over `ds4_think_mode`; with the new `DS4_THINK_LOW` enum value it no longer compiles (unhandled case). LOW behaves like the DeepSeek default: thinking enabled, no effort prefix, so it returns NULL like NONE.
2. **`ds4_server.c`: `render_glm_chat_prompt_text()`** — the GLM render does `effort ? effort : "Reasoning Effort: Max"`, so a LOW request (which intentionally has no prefix text) silently renders as **Max** — the exact "low means the biggest prompt" failure mode this PR fixes for DeepSeek, resurfacing on the GLM path. Guarding the `<|system|>` block on `effort != NULL` fixes it.

```diff
--- a/ds4_cli.c
+++ b/ds4_cli.c
@@ -1333,6 +1333,7 @@ static const char *repl_glm_reasoning_effort_text(ds4_think_mode mode) {
     switch (mode) {
     case DS4_THINK_HIGH: return "Reasoning Effort: High";
     case DS4_THINK_MAX:  return "Reasoning Effort: Max";
+    case DS4_THINK_LOW:  return NULL;
     case DS4_THINK_NONE: return NULL;
     }
     return NULL;
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -2559,8 +2559,10 @@ static char *render_glm_chat_prompt_text(const chat_msgs *msgs,
     buf_puts(&out, "[gMASK]<sop>");
     if (think) {
         const char *effort = ds4_glm_reasoning_effort_text(think_mode);
-        buf_puts(&out, "<|system|>");
-        buf_puts(&out, effort ? effort : "Reasoning Effort: Max");
+        if (effort) {
+            buf_puts(&out, "<|system|>");
+            buf_puts(&out, effort);
+        }
     }
```

Plus a server-side regression test mirroring the existing prefix tests: a LOW render must contain no `Reasoning Effort:` text while still ending thinking-enabled (`<|assistant|><think>`), so it can't silently fall back to the GLM Max default again.

With that fixup: warning-free build, full `make test` green on M3 Ultra Metal, and the PR's own reasoning-effort mapping tests pass unchanged. Happy to push the fixup somewhere if useful.

Flor1an-B:
Thanks for running it rebased and sharing the fixups — both were real and both were mine: adding the `DS4_THINK_LOW` enum value left two pre-existing GLM consumers un-updated.

- `repl_glm_reasoning_effort_text()` (ds4_cli.c): confirmed the `-Wswitch` warning by compiling the object standalone; added `case DS4_THINK_LOW: return NULL;` (low carries no effort prefix, like the DeepSeek default).
- `render_glm_chat_prompt_text()` (ds4_server.c): exactly as you said — `ds4_glm_reasoning_effort_text()` already returns NULL only for LOW, so `effort ? effort : "Reasoning Effort: Max"` silently rendered a LOW request as **Max**, resurrecting the "low means the biggest prompt" bug on the GLM path. Guarded the `<|system|>` block on `effort != NULL`.

Added `test_render_glm_low_effort_has_no_prefix` (a LOW GLM render must contain no `Reasoning Effort:` text while still ending thinking-enabled), so it can't silently fall back to Max again. I also swept the remaining `DS4_THINK_*` decision sites to confirm those two were the only ones the enum addition broke — they were. Warning-free build, `./ds4_test --server` green. Pushed as 4d6b2ed.

kmike:
I haven't checked code in detail, but I think analysis in the ticket is valid:

* 0731 has 2 different prompt for high and max; high matches current 'max' in ds4, and max is not here
* "low" handling is not right in main
* "384K context" thing is a recommendation, not  requirement, so it's good to remove it.

UPD: I've also used this version a bit, and can confirm it makes ds4 use different thinking levels for low/high/max (i.e. the output has different amount of thinking on average).

kmike:
It seems this PR can benefit from README update, to bring example configurations and implementation in line.

PR #707, opened 2026-08-05
Cancel server work when clients disconnect

mcollina:
## Summary

- detect client FIN/RST while requests are queued or being processed
- detach abandoned queued/assigned jobs and cooperatively cancel prefill and decode work
- preserve in-flight decode and stack-owned job/socket lifetime during cancellation and shutdown
- prevent cancelled requests from publishing responses or stale live continuation state
- add no-model regressions for socket detection, queue ownership, progress cancellation, and decode accounting

A graceful peer FIN, including a write-side half-close, is intentionally treated as request cancellation for this one-request HTTP/1.1 server.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -8427,6 +8427,7 @@ struct server_slot {
     int continued_last_store_tokens;
 
     job *assigned;
+    job *running;
     bool busy;
     bool prefill_waiting;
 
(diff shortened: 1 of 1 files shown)


mcollina:
Most harness have short timeouts and quick retries. This change avoid that a busy queue become a self-dos.

Issue #691, opened 2026-08-05
KV cache reuse breaks for OpenAI tools clients that don't replay reasoning_content


hennas-waifson:
Hello,
I don't have any coding experience, just used Claude Fable to 'solve' an issue - pre-fill was triggered on some inputs. Then I asked it to describe the fix (Fable fixed it), in case it is useful. Here's the page:

[notes-for-antirez.md](https://github.com/user-attachments/files/30727298/notes-for-antirez.md)

Flor1an-B:
Thanks for the notes-for-antirez writeup (and for testing this yourself first) — the diagnosis matched what I found reading the current source, right down to the exact functions involved. Opened a PR with the fix, tests, and a fuller writeup of the mechanism: #709

Kept it scoped to the DeepSeek/DSML path, since I couldn't verify the equivalent for GLM's history renderer (it keeps the opening `<think>` tag either way, so the toolless mechanism's byte-shape assumption doesn't transfer directly — flagged in the PR rather than guessing).

PR #709, opened 2026-08-05
server: remember a tools-aware visible KV checkpoint after tool-call turns

Flor1an-B:
Fixes the root cause reported in #691 (and the same-shaped tool-continuation gap discussed in #609/#611, for the OpenAI-compatible chat/completions path once a conversation actually has tools).

## Root cause

`chat_history_uses_tool_context()` makes `render_deepseek_chat_prompt_text()` always keep the opening `<think>` tag for a historical assistant turn once any tool schema is present — `<think>{reasoning}</think>` — rendering `{reasoning}` from whatever `reasoning_content` the client echoes back for that turn. Most OpenAI-compatible chat/completions clients never replay `reasoning_content`, so that renders as a literal `<think></think>`, while the live KV session actually contains the real sampled reasoning bytes for that turn.

The existing visible-checkpoint mechanism already handles exactly this mismatch for the toolless case (`remember_thinking_checkpoint()` / `should_remember_thinking_checkpoint()`), but both are explicitly gated off whenever `request.has_tools` is true, and the tool-call-finish path (`canonicalize_tool_checkpoint()`, or the raw-DSML-replay fallback when `should_canonicalize_tool_checkpoint()` is false) only ever calls `thinking_live_clear()` afterwards. So a tools-enabled conversation never gets a live-continuation match once a non-replaying client's rendered history diverges from the live tokens — which happens on essentially every turn, since `thinking_live_visible_prefix_prompt()` (the matching side) is already engine-agnostic and untouched by `has_tools`; only the *remembering* side was gated off.

## Fix

Mirrors the toolless mechanism instead of introducing a new one:

- `should_remember_thinking_checkpoint()` only excludes `has_tools` for GLM, whose history renderer keeps the opening `<think>` tag either way (`append_glm_assistant_message_prefix()`) and isn't affected by this bug.
- New `build_tool_context_thinking_visible_text()` mirrors `build_toolless_thinking_visible_text()` but keeps the opening `<think>` tag, matching what `chat_history_uses_tool_context()` will actually render for this turn once it becomes history (the toolless helper strips it, which is correct only when tool context is absent).
- New `remember_tool_thinking_checkpoint()` remembers `prompt_text + build_tool_checkpoint_suffix(..., reasoning=NULL, ...)` as the visible key after a tool-call finish, instead of unconditionally clearing it. A client that *does* replay real reasoning still gets an ordinary token-prefix hit first (before this fallback is ever consulted), so this is purely additive — it doesn't change behavior for clients that already work today.

Both new builders/functions are DeepSeek/DSML-syntax only (gated on `model_syntax != SERVER_MODEL_SYNTAX_GLM`); GLM keeps its current toolless-only behavior unchanged, since I haven't verified this mechanism against GLM's history-rendering shape and didn't want to guess at a model family I can't test.

## Testing

- Full existing `ds4_test` suite passes unchanged.
- Two new regression tests: `test_tool_context_thinking_visible_text_keeps_think_tag` (verifies the two builders produce different shapes for the identical request, which is the whole point of not reusing the toolless one) and `test_tool_checkpoint_suffix_omits_reasoning` (verifies the remembered suffix omits reasoning while still reproducing the exact tool-call DSML).
- Updated `test_thinking_checkpoint_remember_gate` to cover the new GLM/non-GLM split explicitly.

Happy to add a live A/B log-based repro (`thinking live continuation match=visible-prefix` vs `token-mismatch`) if useful, similar to what was posted on #611 — didn't include one here since the existing regression tests already pin down the exact byte-shape logic directly.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -10513,7 +10513,14 @@ static bool continue_after_invalid_dsml(server *s, server_slot *slot,
 static bool should_remember_thinking_checkpoint(const request *r,
                                                 const thinking_state *thinking,
                                                 const char *finish) {
-    if (!r || r->kind != REQ_CHAT || r->has_tools) return false;
+    if (!r || r->kind != REQ_CHAT) return false;
+    /* GLM's history renderer always keeps the opening <think> tag, even for
+     * turns with no reasoning to show (append_glm_assistant_message_prefix()
+     * emits "<think></think>" either way). build_tool_context_thinking_visible_text()
+     * below mirrors the DeepSeek/DSML renderer's shape instead, so only apply
+     * the has_tools relaxation to non-GLM syntaxes; leave GLM's existing
+     * (toolless-only) behavior alone. */
+    if (r->has_tools && r->model_syntax == SERVER_MODEL_SYNTAX_GLM) return false;
     if (r->prompt_preserves_reasoning) return false;
     if (!ds4_think_mode_enabled(r->think_mode)) return false;
     if (finish && (!strcmp(finish, "error") || !strcmp(finish, "length"))) return false;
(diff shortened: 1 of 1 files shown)


Flor1an-B:
Cross-link: #714 fixes the same "cold prefill after tool turns" family on the Responses API path — clients that **replay** the final-answer reasoning summary, the mirror image of the omit-reasoning case here — and independently arrives at the same "multiple visible-key shapes at one KV frontier" idea.

It generalizes it into a primary + alternate store: `thinking_live_remember()` gains a `visible_text_alt`, `build_tool_checkpoint_suffix()` gains `include_final_reasoning`, and `responses_live_visible_prefix_len_locked()` matches either key. The reasoning-free tool key this PR remembers is naturally one of those alternates, so the two should be unified rather than run as parallel mechanisms — whichever lands second plugs into the other's alternate-key slot (noting #714 currently keeps exactly one alternate, so covering both cases needs room for more). Full write-up on #714.

Flor1an-B:
Cross-link: #789 (@fradav) ports this visible-checkpoint idea to chat/Anthropic tool turns much more broadly — tool-call turns, Anthropic, waypoints stored through tool calls, and checkpoint preservation on disconnect/stream error.

The one axis where the two differ is the reasoning in the remembered key (`build_tool_checkpoint_suffix()` renders `reasoning + "</think>"`): #789 keeps `parsed_reasoning`, matching clients that **replay** reasoning (gated on `prompt_preserves_reasoning`), whereas this PR passes `reasoning = NULL` for the **strip**-reasoning case (#691). On current `main` that strip case is covered by neither — `should_remember_thinking_checkpoint()` still excludes `has_tools`, and #789's `should_remember_tool_context_checkpoint()` requires `prompt_preserves_reasoning`. So they're complementary halves, not duplicates.

The consolidation would select the reasoning by `prompt_preserves_reasoning` in a single path (`NULL` when stripped = this PR's key, `parsed_reasoning` when preserved = #789's), plus #789's broader scope. Full write-up on #789. Note this PR is based on the pre-`prompt_preserves_reasoning` `main`, so it needs a rebase/reconcile onto that structure regardless.

Issue #668, opened 2026-08-03
Live KV cache cannot rewind when a repeated prompt is an exact prefix of the current session (GLM 5.2)


Amajiro:
## Title

Live KV cache cannot rewind when a repeated prompt is an exact prefix of the current session

## Summary

`ds4-server` performs a full prefill when the same chat request is sent again, even when the incoming prompt is an exact token-for-token prefix of the current live KV state.

This affects regeneration/retry workflows. The live cache contains:

```text
prompt + previous generated response
```

The repeated request contains:

```text
same prompt
```

`ds4-server` correctly reports that the entire incoming prompt matches the live prefix:

```text
live_prompt_common == prompt_tokens < live_tokens_before
```

However, it treats this as `token-mismatch`, reuses zero tokens, and performs a fresh prefill instead of rewinding the live KV state to the prompt boundary.

This can be very expensive for large prompts. In my real workload, a repeated 48K-token prompt takes roughly 158 seconds to prefill on an M3 Ultra, despite the whole prompt already existing at the start of the live KV cache.

## Environment

* macOS on Apple M3 Ultra, 512 GB unified memory
* `ds4-server`
* GLM 5.2
* Model: `GLM-5.2-UD-Q4_K_RoutedQ4K.gguf`
* Metal backend
* Context: 131072
* Issue reproduced with direct `curl`, without SillyTavern
* Thinking explicitly disabled

## Minimal reproduction

Start the server with tracing enabled:

```bash
./ds4-server \
  -m /path/to/GLM-5.2-UD-Q4_K_RoutedQ4K.gguf \
  --ctx 131072 \
  --host 127.0.0.1 \
  --port 8080 \
  --trace /tmp/ds4-trace.log
```

Send this request:

```bash
curl http://127.0.0.1:8080/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{
    "model": "glm-5.2",
    "messages": [
      {
        "role": "user",
        "content": "Say hello"
      }
    ],
    "thinking": {
      "type": "disabled"
    }
  }'
```

Then send the exact same request again.

## Trace output from the second request

```text
kind: chat
model: glm-5.2
stream: 0
tools: 0
think_mode: none
prompt_tokens: 8
effective_prompt_tokens: 8
cached_tokens: 0

--- cache decision ---
live_tokens_before: 17
prompt_tokens: 8
live_prompt_common: 8
memory_token_reusable: 0
memory_miss_reason: token-mismatch
cache_source: none
[...]

PR #710, opened 2026-08-05
server: extend live-prefix rewind to Flash on Metal, bounded by raw-cache budget

### LINKED ISSUE #698: Extend live-prefix rewind (#668) beyond GLM: DeepSeek V4 Flash on Metal still cold-prefills exact-prefix retries
## Summary

#668 added live-KV rewind for repeated exact-prefix prompts (commit 7694112), but `live_prefix_rewind_target` is gated on `ds4_engine_is_glm_dsa()`, so DeepSeek V4 Flash on the Metal backend still does a full cold prefill when a client resends a prompt that is an exact token-level prefix of the live session.

## Reproduction / evidence

Agent client (Hermes Agent, OpenAI chat/completions) aborts a stream mid-generation (user redirect). The slot's live KV then holds `prompt + partial generation`. The resent prompt is an exact prefix of that state — and is treated as a total miss:

```
ds4-server: live kv cache miss live=175286 prompt=175249 common=175249 reason=token-mismatch
ds4-server: chat ctx=0..175249:175249 prompt start
...
ds4-server: chat ctx=0..175249:175249 prompt done 643.560s
```

`common == prompt_len` (the whole 175,249-token prompt matches the live prefix; only the 37 partially-generated tokens diverge), yet the slot is discarded and ~10.7 minutes of prefill are redone. This happened 4 times in 15 minutes in a real session.

## Why this seems generalizable

`ds4_session_rewind()` itself looks engine-agnostic (truncates the checkpoint, invalidates MTP/DSpark capture). The GLM gate appears to guard the dense-cache capping (`ds4_session_glm_cap_dense_cache`). Question: what needs validation to enable the rewind for the Flash/Metal session path — is there a Metal-graph or streamed-expert invariant that a truncated checkpoint violates?

## Environment

- macOS, Apple Silicon
- `ds4-server -m ds4flash.gguf -c 262144 --host 127.0.0.1 --port 8001 --kv-disk-dir /tmp/ds4-kv --kv-disk-space-mb 8192 --batched-session 4 --mtp .../DeepSeek-V4-Flash-DSpark-support.gguf --dspark`
- Model: DeepSeek V4 Flash
- Client: Hermes Agent (also reproduced the pattern described in #609)

Related: #609 / PR #611 (OpenAI tool-round continuation), #444 (disk KV eviction). Happy to test any branch — the repro environment is live here.

Flor1an-B:
Fixes #698.

## Background

#668 added live-KV rewind for repeated exact-prefix prompts (retry/regenerate workflows), but gated it to GLM only via `ds4_engine_is_glm_dsa()`. #698 reports that DeepSeek V4 Flash on Metal still does a full cold prefill in the same scenario, even when the incoming prompt is an exact token-prefix of the live session (e.g. a 175K-token prompt re-prefilled from scratch after a client-side stream abort).

## Why the gate exists (this isn't a small oversight)

Read `metal_graph_encode_token_raw_swa()` rather than guessing: Metal's raw SWA cache is a genuine ring buffer, `raw_row = pos % raw_cap`. `ds4_session_rewind()` only truncates `checkpoint.len` — it does not shift or revalidate that ring. GLM's dense KV cache has no such ring (`ds4_session_glm_cap_dense_cache()` keeps it consistent unconditionally on rewind), but naively enabling rewind for Flash risks reusing a raw row that the *discarded* continuation has already overwritten — silently conditioning the resumed generation on stale/wrong tokens. That's a correctness bug, not just a missed cache hit, so I didn't want to just flip the gate.

## Fix

New `ds4_session_raw_rewind_budget()`:
- Returns 0 for GLM (unaffected, still gated the existing way).
- For Flash, returns `raw_cap - raw_window` — the margin the graph already reserves specifically so chunked prefill doesn't evict window-relevant rows mid-chunk. That existing margin turns out to be exactly the safe reuse budget for this too.

The rewind is only allowed when the discarded tail (`old_pos - prompt_len`) stays under that budget, which guarantees no raw row the rewound tail depends on has been overwritten.

## Empirical validation, not just the theoretical argument

M5 Max, Metal, production Flash q2-q4 quant:

1. **Positive control**: real `ctx=8192` session, ~1000-token prompt, resend the exact same prompt after a short generation → log shows `rewound Flash live prefix from N to M`, continuation is byte-identical to the first generation.
2. **Negative control**: forced `raw_cap=300` via `DS4_METAL_GRAPH_RAW_CAP` (budget becomes an easy-to-exceed 44 tokens), *with the budget check temporarily removed* to confirm the failure mode is real and not just theoretical — generated 400 tokens (well past budget), resent the same prompt: continuation corrupted into a degenerate `140,140,140,...` repetition loop instead of the correct count, while the unmodified first generation was correct. Same forced `raw_cap`, same over-budget scenario, *with* the guard in place: correctly falls through to a full re-prefill (`reason=token-mismatch` in the log) instead of rewinding, and stays correct.
3. A short within-budget retry under that same forced `raw_cap=300` still rewinds and matches, confirming the guard isn't just conservative to the point of never firing.

Happy to re-run any of this against a different quant/ctx if useful — the scratch negative-control build was throwaway and not included here, only the guarded version in this diff.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -65546,3 +65546,24 @@ int ds4_session_ctx(ds4_session *s) {
 int ds4_session_prefill_cap(ds4_session *s) {
     return s ? (int)s->prefill_cap : 0;
 }
+
+/* Metal's raw SWA cache is a ring buffer of raw_cap rows indexed by
+ * pos % raw_cap (see metal_graph_encode_token_raw_swa()). ds4_session_rewind()
+ * only truncates checkpoint.len; it does not shift or revalidate that ring.
+ * A live-prefix rewind is only safe to reuse those rows verbatim while the
+ * discarded tokens have not yet wrapped a row the rewound tail still needs --
+ * i.e. while (old_pos - new_pos) stays under (raw_cap - raw_window), the
+ * margin the graph already reserves so chunked prefill doesn't evict
+ * window-relevant rows mid-chunk. Returns 0 ("no budget, do not rewind") for
+ * GLM, whose dense KV cache has no such ring and is gated separately via
+ * ds4_engine_is_glm_dsa(), and for CPU-only builds with no raw cache. */
+uint32_t ds4_session_raw_rewind_budget(const ds4_session *s) {
+#ifndef DS4_NO_GPU
+    if (!s || DS4_MODEL_FAMILY == DS4_MODEL_FAMILY_GLM_DSA) return 0;
+    if (s->graph.raw_cap <= s->graph.raw_window) return 0;
+    return s->graph.raw_cap - s->graph.raw_window;
+#else
+    (void)s;
+    return 0;
+#endif
+}
--- a/ds4.h
+++ b/ds4.h
@@ -415,6 +415,7 @@ void ds4_session_rewind(ds4_session *s, int pos);
 int ds4_session_pos(ds4_session *s);
 int ds4_session_ctx(ds4_session *s);
 int ds4_session_prefill_cap(ds4_session *s);
+uint32_t ds4_session_raw_rewind_budget(const ds4_session *s);
 int ds4_engine_routed_quant_bits(ds4_engine *e);
 bool ds4_engine_has_output_head(ds4_engine *e);
 bool ds4_engine_has_mtp(ds4_engine *e);
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -11170,9 +11170,18 @@ static void generate_job(server *s, server_slot *slot, job *j) {
                    "Anthropic continuation state is not available; retry by replaying the full messages history");
         return;
     } else if (cached == 0) {
+        const bool is_glm = ds4_engine_is_glm_dsa(s->engine);
+        /* GLM's dense KV cache can always rewind: ds4_session_glm_cap_dense_cache()
+         * keeps it consistent. Flash's raw SWA cache is a ring buffer instead
+         * (see ds4_session_raw_rewind_budget()), so only rewind it while the
+         * discarded tail is still guaranteed not to have wrapped a row the
+         * rewound tail will need. */
+        const uint32_t raw_budget = is_glm ? 0 : ds4_session_raw_rewind_budget(slot->session);
+        const bool can_rewind = is_glm ||
+            (raw_budget > 0 && j->req.prompt.len < old_pos &&
+             (uint32_t)(old_pos - j->req.prompt.len) < raw_budget);
         const int rewind_to = live_prefix_rewind_target(
-            ds4_engine_is_glm_dsa(s->engine), old_pos,
-            j->req.prompt.len, common);
+            can_rewind, old_pos, j->req.prompt.len, common);
         if (rewind_to >= 0) {
             pthread_mutex_lock(&s->inference_mu);
             ds4_session_rewind(slot->session, rewind_to);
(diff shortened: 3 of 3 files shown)


overstrike:
I tried this on top of current main HEAD and found it drastically speeds up long context prefill on subsequent turns. Tried several clients.

Flor1an-B:
Re-validated on current `main` (84cc882), i.e. **after the recent Metal decode refactor** (`0ad494e` make exact fast paths automatic, `42033ee` pipeline DFlash verification, `0658dac` fix long-context prefill/decode correctness). A clean textual rebase doesn't prove the raw-SWA ring semantics this PR depends on are unchanged, so I re-ran the controls rather than assume.

Rebased onto `84cc882` (no conflicts), `make ds4-server ds4_test` builds `-Wall -Wextra` with **0 warnings**, and `./ds4_test --server` passes (incl. `test_live_prefix_rewind_target`). Apple M5 Max, Metal, production Flash q2-q4, `--ctx 8192` (`raw_kv_rows=4352`).

**Positive control** — same prompt sent twice, temp 0:
```
req1: cache_source=none        cached_tokens=0    (cold, 29-tok prompt; live grows to 89)
req2: cache_source=memory-rewind cached_tokens=28
      ds4-server: rewound Flash live prefix from 89 to 28; final prompt token will be reevaluated
```
Both completions **byte-identical** (`cmp -s`).

**Negative control** — forced `DS4_METAL_GRAPH_RAW_CAP=300` (budget becomes ~44), generated 400 tokens (well over budget), resent the same prompt:
```
req2: memory_miss_reason=token-mismatch  cache_source=none  cached_tokens=0
      (no "rewound ..." line — the budget guard declined and fell through to full re-prefill)
```
Completion still **byte-identical** — the guard correctly refuses to reuse a raw row the discarded tail may have overwritten, so correctness holds exactly where the earlier scratch build showed corruption without it.

GitHub reports this MERGEABLE / CLEAN against current main. Nothing else to rebase.

GreatMark:
Verified on Mac Studio M3 Ultra (80-core GPU, 512 GB), Metal, MXFP4 Flash 0731 fully resident, merged on `84cc882` (stacked with #765, #611, #704, #767 — see note below).

## Exact-prefix retry scenario

~5.3K-token prompt (16K chars), 42 tokens generated, then the identical request re-sent (streaming, `temp 0`):

| | main | this PR |
|---|---|---|
| retry log | `live kv cache miss live=5356 prompt=5314 common=5314 reason=token-mismatch` — common equals the whole prompt and it still cold-prefills | `rewound Flash live prefix from 5356 to 5313`, then `ctx=5313..5314:1` |
| retry TTFB | 5.05 s | **0.045 s** |
| retry total (42 gen tokens) | 11.85 s | 1.19 s |

The baseline miss line is a nice self-contained proof of the gap this closes: `common=5314` on a 5,314-token prompt, discarded anyway.

Generation after the rewind matched the first run's output. `./ds4_test --server` passes, and a `ds4-bench` 2K–16K sweep is within ±1% of baseline (the `ds4_session_raw_rewind_budget` addition doesn't touch the hot path).

## Rebase note for whoever lands this after #765

#765 moves the rewind probe into `slot_probe_reuse_locked()`. The port that worked here: compute `can_rewind` (GLM unconditional; Flash gated by `ds4_session_raw_rewind_budget()` and the discarded-tail bound, exactly as this PR's `generate_job_inner` hunk) inside the probe where `live_prefix_rewind_target()` is called, and generalize the `rewound GLM live prefix` log line to GLM/Flash. The ds4.c/ds4.h budget function applies unchanged.

*Measured locally; an AI agent helped run the A/B harness. Log lines are ds4-server's own.*

PR #711, opened 2026-08-05
kvstore: don't let a shared fixed payload cost bias eviction toward big files

### LINKED ISSUE #444: Disk KV eviction preferentially removes short-prefix cache files causing disk cache miss & full pre-fill when Disk KV Store is full
## Background
As the good balance of speed & granularity, I am running a Disk KV configuration of 4096 tokens (--kv-cache-continued-interval-tokens 4096, --kv-cache-boundary-align-tokens 4096) backing OpenClaw. DS4 writes a new disk checkpoint every 4,096 tokens during prefill:

```
0–4,096 tokens   → one .kv file  (~76 MiB)
0–8,192 tokens   → one .kv file  (~130 MiB)
0–12,288 tokens  → one .kv file  (~184 MiB)
```

## Issue
I was running a 256Gb Disk KV Cache store limit which I ran out of today. When the disk budget runs out, DS4 must evict files to make room. The eviction score formula is:
```
score = (effective_hits + 1.0) * tokens / file_size
```
When there are no hits, `tokens / file_size` is token density — tokens saved per byte of disk used. Because all KV files for the same model have roughly the same per-token size, density increases slightly with file size due to fixed header overhead. Small files always score lower than large files and are evicted first.

This creates a self-defeating cycle when disk KV store hits the limit:

1. New turn arrives. DS4 starts prefilling a ~35,000-token prompt.
2. At each 4,096-token boundary it writes a small checkpoint file.
3. Disk is full → the small checkpoint (76 MiB, 4,096 tokens) is immediately evicted to make room for the next larger one.
4. By the end of prefill, only large files survive (e.g. the 481 MiB full-context snapshot).
5. Next turn: the new prompt differs slightly from the saved one — the client strips metadata from old messages as they become history, so the rendered text diverges around token ~34,000 out of ~35,000.
6. DS4 checks disk. The only surviving file is the 481 MiB snapshot, but its stored text extends past the divergence point, so the SHA1 doesn't match.
7. The smaller files (e.g. 0–32,768 tokens) would have matched, since their text falls entirely within the stable prefix — but they were evicted in step 3.
8. Full re-prefill from scratch: ~100–200 seconds instead of ~8 seconds.

## Temporary Workaround
I nuked the kv store & started with a fresh 400 Gb store to avoid these eviction issues. This will again stop working when KV store hits its limit but at least now I know what to monitor.

## Pre-Post Results
State | Turn-start time | Disk KV loaded
-- | -- | --
Fresh disk / well under budget | ~8–12 s | 32,768 tokens from disk ✓
Disk at capacity (256 GB, 441 stale files) | 100–208 s | 0 ✗
After nuke + budget increase to 400 GB | ~8–12 s | 32,768 tokens from disk ✓

## Long Term Fix needed
We need to revisit the kv-eviction score formula to re-prioritize what should be evicted. I will also try experimenting with some options.

Flor1an-B:
Fixes #444.

## The reported symptom

Under disk-KV pressure, small "continued" checkpoints (the early, still-reusable prefixes of a growing conversation) get evicted before large ones. Eventually only the largest (full-length) snapshot survives, and once the incoming prompt diverges anywhere before its stored length, it no longer matches — forcing a full re-prefill that a smaller, earlier checkpoint would have avoided. The issue's own math pointed at "token density increases slightly with file size due to fixed header overhead."

## Where the fixed cost actually comes from

Not the 48-byte file header (negligible) — `session_payload_live_tensor_bytes()` caps the persisted raw SWA cache at `raw_window` rows regardless of checkpoint length (`session_raw_live_rows()`). So every checkpoint beyond `raw_window` tokens — the common case for anything past the disk cache's default `min_tokens` floor — pays roughly the *same* fixed number of raw-cache bytes no matter how many tokens it actually holds. That shared fixed cost is a much bigger fraction of a short checkpoint's `file_size` than a long one's, so `tokens/file_size` scores short checkpoints lower than long ones at equal hit counts — not because they're less useful, but purely because they're short.

## Fix

`ds4_kvstore` now carries an optional `fixed_overhead_bytes` estimate (defaults to 0, so `ds4-agent` and any other caller that doesn't set it are completely unaffected — this is opt-in). `ds4_kvstore_entry_eviction_score()` subtracts it from `file_size` before computing density, so the score reflects each checkpoint's own marginal bytes rather than a cost every checkpoint pays once regardless of length.

New `ds4_engine_checkpoint_fixed_overhead_bytes()` computes the estimate from the engine's own `raw_window`/layer-count/head-dim shape (returns 0 for GLM, whose dense KV cache scales with checkpoint length and has no equivalent fixed cost — I didn't want to apply an unverified correction there). `ds4-server` wires it in once at `kv_cache_open()` time, right next to the existing `budget_bytes` setup — no new config surface.

## Testing

- Full existing `ds4_test` suite passes unchanged (`fixed_overhead_bytes` defaults to 0, so behavior is bit-for-bit identical when unset).
- New regression test `test_kv_cache_eviction_ignores_shared_fixed_overhead`: two entries with the exact same marginal (per-token) cost but different sizes — scores differ without the fix (reproducing the reported bias) and become equal with it — plus an edge-case check that an overestimated overhead can't divide by zero or go negative.

I considered a couple of alternatives before landing here — worth saying why I didn't:
- Estimating the overhead by fitting it from the existing on-disk entry population (no cross-module engine knowledge needed) — works but adds real complexity/fragility to a hot eviction path for what the engine already knows exactly.
- Actually reworking "value" to account for short prefixes being *more* robust to future divergence than long ones (not just equally dense) — a real effect, but a deeper design change than this specific bug needs; happy to discuss separately if useful.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -57271,20 +57271,40 @@ bool ds4_engine_is_glm_dsa(ds4_engine *e) {
 
 /* Estimate for the disk KV cache's eviction scoring (see
  * ds4_kvstore_entry_eviction_score()): the fixed, checkpoint-length-independent
- * portion of a saved session payload. session_payload_live_tensor_bytes()
- * caps the persisted raw SWA cache at raw_window rows regardless of how many
- * tokens the checkpoint actually holds (session_raw_live_rows()), so every
- * checkpoint beyond raw_window tokens -- the common case for anything past
- * the disk cache's default min_tokens floor -- pays roughly this many bytes
- * of raw-cache overhead no matter its length. GLM's dense KV cache scales
- * with checkpoint length instead and has no equivalent fixed cost. */
+ * portion of a saved session payload. Mirrors session_payload_live_tensor_bytes()'s
+ * per-layer terms, keeping only the ones that don't scale with checkpoint
+ * length:
+ * - the raw SWA cache, capped at raw_window rows regardless of how many
+ *   tokens the checkpoint actually holds (session_raw_live_rows()), so every
+ *   checkpoint beyond raw_window tokens -- the common case for anything past
+ *   the disk cache's default min_tokens floor -- pays roughly this many bytes
+ *   of raw-cache overhead no matter its length;
+ * - each compressed layer's attention/index *state* (layer_attn_state_bytes(),
+ *   layer_index_state_bytes()), which depends only on that layer's fixed
+ *   compress ratio, not on the number of compressed rows actually stored
+ *   (g->layer_n_comp[]/layer_n_index_comp[] scale with tokens and are
+ *   deliberately excluded here).
+ * GLM's dense KV cache scales with checkpoint length instead and has no
+ * equivalent fixed cost. */
 uint64_t ds4_engine_checkpoint_fixed_overhead_bytes(ds4_engine *e, int ctx_size) {
 #ifndef DS4_NO_GPU
     if (!e || DS4_MODEL_FAMILY == DS4_MODEL_FAMILY_GLM_DSA) return 0;
     uint32_t raw_window = DS4_N_SWA;
     if (raw_window > (uint32_t)ctx_size) raw_window = (uint32_t)ctx_size;
     if (raw_window == 0) raw_window = 1;
-    return (uint64_t)DS4_N_LAYER * raw_window * DS4_N_HEAD_DIM * sizeof(float);
+    uint64_t bytes = 0;
+    for (uint32_t il = 0; il < DS4_N_LAYER; il++) {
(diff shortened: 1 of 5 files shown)


Flor1an-B:
Self-review caught an incompleteness: the fixed-overhead estimate only accounted for the raw SWA cache's contribution, missing \`layer_attn_state_bytes()\`/\`layer_index_state_bytes()\` — per-compressed-layer terms that \`session_payload_live_tensor_bytes()\` also adds and that depend only on each layer's fixed compress ratio, not checkpoint length. Pushed a fix that mirrors the full per-layer loop (excluding only the genuinely token-scaling terms). Full test suite still passes.

PR #714, opened 2026-08-06
fix(server): retain Responses cache after tool turns

blake-gj:
## Problem

Stateless Responses clients replay the typed output items from the previous
turn. When `reasoning.summary` is enabled, that replay includes the emitted
final-answer reasoning summary.

DS4's live visible key intentionally omitted final-answer summaries for older
clients that do not replay them. After a tool-output continuation, however, the
live sampled KV can tokenize differently from a freshly rendered prompt at the
tool boundary. A client that did replay the summary then missed both the exact
token prefix and the visible-prefix fallback, forcing a cold prefill on its
next user turn.

Four consecutive production occurrences at 18k-39k input tokens stopped at the
preceding tool-call frontier and spent 12.6-26.9 seconds cold-prefilling.

## Change

- Make the primary live visible key mirror the emitted final reasoning summary.
- Retain one summary-free alternate key at the same KV frontier for older
  clients.
- Accept either key in the existing visible-prefix continuation path.
- Cover both replay shapes and the live-frontier guard in the server tests.

This adds no protocol or configuration flag. Durable disk KV continues to use
only the primary, faithful replay shape.

## Validation

Clean branch based on `b030961`:

- `git diff --check`
- `make clean && make` — warning-free Metal build on an Apple M4 Pro
- `make cpu` — warning-free CPU build
- `make ds4_test && ./ds4_test --server`
- `./ds4-eval --self-test-extractors`
- Agent, Q4_K, MXFP4, layer-pack, multi-GPU placement, GPU-argument, and CLI
  suites all passed

The aggregate `make test` reached the default model-backed long-context group
and stopped because this Mac has no `ds4flash.gguf`; all remaining
model-independent targets were run directly and passed.

The same change was also built for CUDA `sm_120` on an RTX 5090 and RTX PRO
6000, and for ROCm `gfx1151` on a Ryzen AI MAX+ 395. A three-host distributed,
streamed `/v1/responses` regression used a DeepSeek-V4-Flash hybrid Q2/Q4 model
at 256K context:

Those GPU builds emitted the pre-existing `cache_source` `-Wstring-compare`
warning and no new warning from this diff.

| Request | Input | Cached | Evaluated | Prefill |
| Initial tool request | 419 | 0 | 419 | 1.100 s |
| Tool-output continuation | 524 | 505 | 19 | 0.130 s |
| Next user turn after final answer | 563 | 557 | 6 | 0.058 s |

The last request matched `responses-visible` and retained 98.9% of its prompt;
this is the transition that cold-prefilled before the change.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -8400,6 +8400,12 @@ typedef struct {
      * Anthropic currently uses only the call-id side of the state. */
     char *visible_text;
     size_t visible_len;
+    /* Some Responses clients replay emitted final-answer reasoning summaries,
+     * while older clients omit them.  Both transcripts describe the same live
+     * KV frontier, so retain one optional alternate key instead of forcing one
+     * client shape to cold-prefill after a tool continuation. */
+    char *visible_text_alt;
+    size_t visible_alt_len;
     /* Tool-call ids generated at the same live frontier. A following tool
      * result for these ids is a direct protocol continuation and should not
      * trigger prompt-prefix matching or checkpoint canonicalization. */
@@ -8691,6 +8697,9 @@ static void live_tool_state_clear_locked(live_tool_state *st) {
     free(st->visible_text);
     st->visible_text = NULL;
     st->visible_len = 0;
+    free(st->visible_text_alt);
+    st->visible_text_alt = NULL;
+    st->visible_alt_len = 0;
     st->valid = false;
     st->live_tokens = 0;
 }
@@ -8738,12 +8747,19 @@ static void thinking_live_remember(server *s, server_slot *slot,
 
 static void responses_live_remember(server *s, server_slot *slot,
                                     const char *visible_text,
+                                    const char *visible_text_alt,
                                     const tool_calls *calls) {
     if (!s || !slot || !visible_text || !visible_text[0]) return;
     pthread_mutex_lock(&s->tool_mu);
     live_tool_state_clear_locked(&slot->responses_live);
     slot->responses_live.visible_text = xstrdup(visible_text);
     slot->responses_live.visible_len = strlen(visible_text);
+    if (visible_text_alt && visible_text_alt[0] &&
+        strcmp(visible_text_alt, visible_text) != 0)
+    {
+        slot->responses_live.visible_text_alt = xstrdup(visible_text_alt);
+        slot->responses_live.visible_alt_len = strlen(visible_text_alt);
+    }
     if (calls) {
         for (int i = 0; i < calls->len; i++) {
             id_list_push_unique(&slot->responses_live.call_ids, calls->v[i].id);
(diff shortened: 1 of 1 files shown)


kmike:
Hey! Faced the same issue (pi agent, configured to use ds4 via Responses API, q2, macbook m5 128gb). 

I think the fix is valid, and approach looks good 👍 . It seems there is an assumption in the current codebase that pi agent doesn't resend the final-answer reasoning summary, but it does; the fix here preserves support for both client behaviors, so looks safe.

Checked it locally - I can reliably reproduce the problem without the fix, and it goes away with the fix. So, consider it validated independently; it'd be great to have it merged to main :)

---

AI-generated repoduction summary:

  **Repro / validation** — pi → ds4 `openai-responses`, `reasoning.summary=auto`, thinking=high, q2; tested on `main` @ `84cc882`.

   Shape: a tool-call turn → `finish=stop` answer → a short follow-up. The follow-up turn is where it breaks.

   `main` (no fix), follow-up turn:
 ```

 live kv cache miss RESPPROTO live=4582 prompt=4595 common=4505 reason=token-mismatch
 chat ctx=4414..4595:181 RESPPROTO TOOLS prompt done 0.759s

 ```
   `common=4505` is the tool-result boundary: the `responses-visible` byte-prefix key fails because `main` omits the final-answer reasoning summary from the key while pi replays it, so the token-prefix fallback
 diverges there. At production ctx sizes `cold_max=30000 < ctx`, so this becomes a full from-zero prefill — the minutes-long stalls from the issue.

   `main` + this PR, same follow-up turn:
 ```

 responses live continuation RESPPROTO match=visible-prefix ids=0 cached=4553 prompt=4564
 chat ctx=4553..4564:11 RESPPROTO TOOLS prompt done 0.552s

 ```
   0 cache misses across the run; only the 11-token user-message suffix gets prefilled.

Flor1an-B:
Cross-referencing #709, which fixes the same "cold prefill after tool turns" symptom from a complementary angle — flagging it because the two overlap mechanically, not just thematically.

**Complementary intent.** #709 targets the OpenAI-compatible chat/completions path (DeepSeek/DSML only) and the *opposite* client behaviour: clients that **omit** the in-`<think>` `reasoning_content` of the tool-call turn, so DS4 remembers a **reasoning-free** visible key. This PR targets the Responses API and clients that **replay** the final-answer reasoning summary, remembering a summary-mirroring primary key plus a summary-free alternate. Different protocol, different bytes — both needed for full coverage.

**Mechanical overlap.** They touch the same two helpers, so they'll conflict on merge and, more to the point, on design:

- `thinking_live_remember()` — this PR adds a `visible_text_alt` parameter; #709 calls it (and adds its own caller for the tool-turn checkpoint).
- `build_tool_checkpoint_suffix()` — this PR adds `include_final_reasoning`; #709 calls it with `reasoning=NULL`.
- The "remember after a tool turn" region of `generate_job()` — this PR adds the alternate-key remember there; #709 adds `remember_tool_thinking_checkpoint()` in the same block.

**The real convergence:** both PRs independently introduce "accept more than one visible-key shape at the same KV frontier." This PR makes it a general primary + alternate store (`responses_live_visible_prefix_len_locked` matching either); #709's reasoning-free tool key is naturally one of those alternates. Whichever lands second should plug into this PR's `visible_text_alt` slot rather than run a parallel mechanism — with the caveat that covering both cases at once needs room for more than a single alternate.

Happy to do the reconciliation on the #709 side if this lands first.

brutuscat:
I think this fix is valid too. At least that issue with finish=stop type logs now turned into a responses live continuation:

```
0812 15:54:01 ds4-server: chat ctx=54408..54497:89 gen=781 RESPPROTO TOOLS finish=stop 34.301s
0812 15:57:44 ds4-server: responses live continuation RESPPROTO match=visible-prefix ids=0 cached=55278 prompt=56357
0812 15:57:44 ds4-server: chat ctx=55278..56357:1079 RESPPROTO TOOLS prompt start
0812 15:57:44 ds4-server: chat ctx=55278..56357:1079 RESPPROTO TOOLS prefill chunk 0/1079 (0.0%) chunk=0.00 t/s avg=0.00 t/s 0.000s
0812 15:57:55 ds4-server: chat ctx=55278..56357:1079 RESPPROTO TOOLS prefill chunk 1079/1079 (100.0%) chunk=98.33 t/s avg=98.32 t/s 10.974s
0812 15:57:55 ds4-server: chat ctx=55278..56357:1079 RESPPROTO TOOLS prompt done 10.974s
```
Thank you @kmike for bringing this up and @blake-gj for the attempt at fixing this.

PR #617, opened 2026-07-27
Add GLM 5.2 inference (SSD streaming and distributed) on ROCm for GLM-5.2-UD-IQ2_XXS_RoutedIQ2XXS_blk78Q2K.gguf

kyuz0:
This PR adds inference support for GLM 5.2 on AMD Strix Halo systems using the ROCm backend for for `GLM-5.2-UD-IQ2_XXS_RoutedIQ2XXS_blk78Q2K.gguf`. This is both for SSD streaming and for distributed inference.

The model can now be split by layer across two nodes and kept resident in memory. Each node maps and loads only its assigned layers, so distributed inference does not require SSD streaming.

It uses the ROCm fixes and optimized attention, projection, MoE, and decode paths needed to make GLM 5.2 inference correct and practical on `gfx1151`.

The FP16 hipBLAS attention approach was informed by [mgiustiniani's GLM 5.2 work](https://github.com/mgiustiniani/ds4/commit/d2487c808b6bca9bb983adbb78f43273dfa722f0).

## Running

The following example splits the 78 transformer layers across two Strix Halo systems:

- coordinator: layers `0:37`, address `192.168.100.2`
- worker: layers `38:output`, address `192.168.100.1`

Start the coordinator:

```sh
DS4_GLM_MEMORY_GUARD_RESERVE_GB=8 \
ds4-server \
  -m /path/to/GLM-5.2-UD-IQ2_XXS_RoutedIQ2XXS_blk78Q2K.gguf \
  --ctx 30000 \
  --host 0.0.0.0 \
  --port 8000 \
  --role coordinator \
  --layers 0:37 \
  --listen 192.168.100.2 8081 \
  --dist-prefill-chunk 256 \
  --dist-prefill-window 2
```

Start the worker:

```sh
DS4_GLM_MEMORY_GUARD_RESERVE_GB=8 \
ds4-server \
  -m /path/to/GLM-5.2-UD-IQ2_XXS_RoutedIQ2XXS_blk78Q2K.gguf \
  --ctx 30000 \
  --role worker \
  --layers 38:output \
  --coordinator 192.168.100.2 8081
```

`--dist-prefill-chunk 256 --dist-prefill-window 2` is the tested prefill configuration.

## Testing

This was tested on two Strix Halo systems with 128 GiB of unified memory each, using ROCm 7.2.4 and:

```text
GLM-5.2-UD-IQ2_XXS_RoutedIQ2XXS_blk78Q2K.gguf
```

Validation included:

[...]

PR #717, opened 2026-08-06
glm: make GLM 5.2 work on CUDA with SSD streaming, including IQ2_XXS

JonasColmsjo:
This is the CUDA counterpart of #617, which added GLM 5.2 SSD streaming and
distributed inference for ROCm using the same
`GLM-5.2-UD-IQ2_XXS_RoutedIQ2XXS_blk78Q2K.gguf`. On CUDA those paths were either
unreachable or silently wrong.

Tested on a DGX Spark (GB10, 121 GB unified). Before this: the 262 GB Q2_K build
generated token soup, the 211 GB IQ2_XXS build could not load at all, and loading
either **without** `--ssd-streaming` wedged the host badly enough to need a power
cycle — with nothing logged to say why. It cost four power cycles to work out.

### 1. Routed MoE never used the streaming expert compaction on CUDA

`ds4_gpu_glm_stream_expert_cache_begin_selected_load_tensor` exists and works, but
its call sites are ROCm (`#ifdef DS4_ROCM_BUILD`), Metal and CPU only — and neither
`ds4_gpu_glm_routed_moe_batch_tensor` nor `..._one_tensor` reads
`g_stream_selected_cache`.

So the batch dispatch asked `cuda_resolve_weight_ptr` for **all `n_total_expert`
experts as one contiguous span**. Streaming can never satisfy that, and on
single-GPU the resolver answers a miss *without an error* — the multi-GPU path
logs `selective-cache miss`, the single-GPU path just returns a pointer. The
kernel then multiplied against memory the model was never written into.

Now wired to the same compaction the DeepSeek MXFP4 path uses at
`ds4_cuda.cu:23727`, and it refuses the layer if compaction is unavailable rather
than reading unpopulated memory.

### 2. IQ2_XXS routed experts had no CUDA path

GLM's routed-MoE kernels are Q2_K-only, so IQ2_XXS was rejected at the type gate —
even though the vendored MMQ tier already implements IQ2_XXS MoE and DeepSeek uses
it on the very same hardware.

The MMQ contract matches GLM's layout exactly: `W` is `[n_experts, M, K]` with
per-expert slabs stacked, `ids` is `[n_tokens, n_expert_used]` row-major, and `out`
is column-major over (token, slot) — the layout GLM already asserts via
`mid_token_stride`. `K=6144` is a multiple of 256 and `n_expert_used=8` is one of
the specialised widths.

Gate/up now dispatch through `ds4_mmq_iq2_xxs_moe_pair`, then SwiGLU and the router
weights via the existing kernel, then down per (token, slot). Down is dispatched on
its own type, so the `RoutedIQ2XXS_blk78Q2K` layout works. The fused
`gate_up_mid_vec` entry cannot be used — it is hardcoded to `n_expert_used == 6`.
Decode needs no separate work: `..._one_tensor` delegates to the batch entry.

### 3. IQ2_XXS layers were diverted to the generic routed path

`glm_graph_layer_uses_generic_routed_moe()` sent them to
`ds4_gpu_routed_moe_batch_tensor`, which needs `g->batch_routed_*` buffers that are
not wired up when a shared prefill workspace is in use — so every routed layer
returned 0 before reaching a kernel. That predicate existed because the GLM path
could not do IQ2_XXS, which (2) changes. Metal and ROCm keep the generic path.

### 4. The memory guard could not see host memory on Linux

`glm_graph_host_memory_bytes()` was macOS-only and returned `0` elsewhere, so the
guard fell back to a GPU working-set estimate. On an integrated GPU that is the
wrong constraint: device allocations come out of host RAM, so an oversized model
does not fail an allocation — it takes the machine down. Now reads the real total
via `sysconf(_SC_PHYS_PAGES)` and, when refusing, prints the shortfall and the
remedy.

### 5. Diagnostics

None of the above was visible. `glm_graph_indexed_prefill_trace_enabled()` and
`_trace_all()` were hardcoded `return false`, making every
`glm_graph_indexed_prefill_tracef()` call site dead code; they now read
`DS4_GLM_INDEXED_PREFILL_TRACE` / `_TRACE_ALL`, cached since they are consulted at
every stage boundary of every layer. The silent entry guards name the precondition
they rejected instead of returning a bare `false` out of a twelve-condition `if`.
Routed-MoE dispatch failures report the expert tensor types — which is what
identifies a quant-support problem. `DS4_CUDA_VERIFY_WEIGHT_PTR` reports any weight
not resolving to device memory, and an unbounded weight cache on an integrated GPU
now says so.

### Verification

On GB10 with SSD streaming:

| | |
|---|---|
| IQ2_XXS (211 GB) | *capital of France* → `Paris`; *reverse a string* → `def reverse_string(s): return s[::-1]` |
| | 0.46 t/s prefill, **0.49 t/s** decode |
| Q2_K (262 GB) | correct, 0.25 t/s — IQ2_XXS is ~2× faster, streaming 51 GB less |
| no `--ssd-streaming` | refuses in seconds, memory untouched: `need 249.56 GiB but only 89.63 GiB is available … retry with --ssd-streaming` |
| DeepSeek V4 Flash | unaffected, resident and streaming (regression-checked) |

Trace and verify flags are off by default, so output is unchanged unless the env
vars are set.

Supersedes #683 and #701, which covered subsets of this.

## Regression Testing Evidence

**PR #717: glm: make GLM 5.2 work on CUDA with SSD streaming, including IQ2_XXS**

### Test Environment
- Machine: gpu1 (NVIDIA DGX Spark, GB10, CUDA compute_121a/sm_121)
- Date: 2026-08-10
- Branch: `glm-cuda-complete`
- Models: DeepSeek V4 Flash (regression baseline), GLM 5.2 IQ2_XXS & Q2_K (PR target)

### Portability: CPU Build
```
make cpu
```
✅ PASSED — CPU-only binaries build without errors (verified locally).
New memory-guard code properly guarded with `#elif defined(_SC_PHYS_PAGES)` and fallback `#else` for non-Linux.

### CUDA Compilation & Build
```
make cuda-spark
```
✅ PASSED — CUDA binaries compile and link successfully on GB10 (sm_121 architecture).
All MMQ-tier changes integrated correctly with existing CUDA paths.

### Correctness Regression Tests
**Commands run on gpu1:**
```
make cuda-regression
./ds4_test --logprob-vectors
./ds4_test --long-context
```
✅ PASSED — Verified DeepSeek V4 Flash unaffected:
- cuda-regression smoke test: all checks pass
- Logprob vectors: continuations correctly predicted (no tokenization or attention regression)
- Long-context: fact-recall intact; context handling unchanged

### GLM 5.2 CUDA Correctness (IQ2_XXS & Q2_K)
**As documented in PR body:**
- IQ2_XXS (211 GB): "capital of France" → `Paris`; "reverse a string" → correct Python; 0.49 t/s generation
- Q2_K (262 GB): correct output; 0.25 t/s (unoptimized streaming path)
- Full-residency refusal (no `--ssd-streaming`): correctly refuses with memory estimate & remedy suggestion
- SSD streaming: stable decode with expert cache compaction confirmed (no "selective-cache miss" errors on CUDA)

### Memory Guard (Linux-Specific Fix)
New `glm_graph_host_memory_bytes()` implementation:
- Correctly reads system RAM via `sysconf(_SC_PHYS_PAGES)` on Linux
- Prevents integrated-GPU wedging by enforcing budget check upfront
- Prints diagnostic message with shortfall and remedy on refusal

✅ Verified: oversized models now refuse cleanly instead of OOM-killing the host.

### Diagnostics
`DS4_GLM_INDEXED_PREFILL_TRACE` / `_TRACE_ALL` env vars now functional (were hardcoded false before).
Stage boundaries and layer failure reasons now visible when enabled.

### Speed Regression
DeepSeek V4 Flash (regression baseline):
- Resident decode: ~20 tok/s (unchanged)
- SSD streaming: ~1.2–1.4 tok/s (unchanged)
No throughput penalties in non-GLM code paths.

### Conclusion
GLM 5.2 now fully functional on CUDA with SSD streaming and IQ2_XXS quant support.
All existing regressions tests pass. DeepSeek unaffected. No performance regressions in shared code paths.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -35053,11 +35053,22 @@ static void glm_debug_dump_prefill_logits(const float *logits) {
 }
 
 static bool glm_graph_indexed_prefill_trace_enabled(void) {
-    return false;
+    /* Cached: this is consulted at every stage boundary of every layer. */
+    static int cached = -1;
+    if (cached < 0) {
+        const char *v = getenv("DS4_GLM_INDEXED_PREFILL_TRACE");
+        cached = (v && v[0] && strcmp(v, "0") != 0) ? 1 : 0;
+    }
+    return cached != 0;
 }
 
 static bool glm_graph_indexed_prefill_trace_all(void) {
-    return false;
+    static int cached = -1;
+    if (cached < 0) {
+        const char *v = getenv("DS4_GLM_INDEXED_PREFILL_TRACE_ALL");
+        cached = (v && v[0] && strcmp(v, "0") != 0) ? 1 : 0;
+    }
+    return cached != 0;
 }
 
 static uint32_t glm_graph_indexed_prefill_trace_slow_ms(void) {
@@ -37909,6 +37920,15 @@ static uint64_t glm_graph_host_memory_bytes(void) {
     size_t len = sizeof(mem);
     if (sysctlbyname("hw.memsize", &mem, &len, NULL, 0) != 0) return 0;
     return mem;
+#elif defined(_SC_PHYS_PAGES) && defined(_SC_PAGESIZE)
+    /* Without this the guard falls back to a GPU working-set estimate, which on
+     * an integrated GPU is not the constraint that actually matters: device
+     * allocations come out of host RAM, so a model that does not fit takes the
+     * whole machine down rather than failing. */
+    const long pages = sysconf(_SC_PHYS_PAGES);
+    const long page_size = sysconf(_SC_PAGESIZE);
+    if (pages <= 0 || page_size <= 0) return 0;
+    return (uint64_t)pages * (uint64_t)page_size;
 #else
     return 0;
 #endif
@@ -38149,6 +38169,21 @@ static bool glm_graph_memory_guard_for_compact_cap(
             ctx_size,
             mem.comp_cap,
             phase ? phase : "before Metal graph allocation");
+    fprintf(stderr,
+            "ds4:   need %.2f GiB but only %.2f GiB is available\n",
+            glm_graph_bytes_to_gib(required),
+            glm_graph_bytes_to_gib(budget));
+    if (!ssd_streaming) {
+        /* The remedy is not obvious from the numbers alone, and proceeding on an
+         * integrated GPU wedges the host rather than failing an allocation. */
+        fprintf(stderr,
+                "ds4:   this model does not fit in memory; retry with "
+                "--ssd-streaming (optionally --ssd-streaming-cache-experts NGB)\n");
+    } else {
+        fprintf(stderr,
+                "ds4:   already streaming; lower --ctx or "
+                "--ssd-streaming-cache-experts\n");
+    }
     if (ssd_streaming && model_bytes != model->size) {
         fprintf(stderr,
                 "ds4:   streamed active model map: %.2f GiB "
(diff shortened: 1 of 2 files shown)


vincenzopalazzo:
Concept ACK, this solves the problem for me

image

PR #718, opened 2026-08-06
Add a minimal web chat UI served by ds4-server

phierru:
I added a minimal web chat interface, served directly by `ds4-server` at `/`, so a
model can be used from a browser without any extra tooling. There is no build step
and nothing to install: the page is embedded in the binary as a string literal, and
`chat-ui.html` is the same page as a standalone file if you prefer to open it from
disk.

## Trying it

```sh
./ds4-server -m ./ds4flash.gguf --ctx 32768
```

Then open <http://127.0.0.1:8000> — the UI is served by `ds4-server` itself on its
usual port, not by a separate process.

Two prompts that exercise the rendering:

1. `Show me the Maxwell equations`
2. `Explain briefly how these equations imply electromagnetic waves`

The first comes back with headings, bold text and display equations; the follow-up
uses inline math inside bullet lists.

## What it does

- Streams responses over the existing `/v1/chat/completions` SSE endpoint.
- Renders Markdown: headings, nested lists, tables, blockquotes, links, horizontal
  rules, fenced and inline code.
- Renders LaTeX with KaTeX: `$$..$$` and `\[..\]` as display math, `$..$` and
  `\(..\)` inline. Inline `$..$` requires a non-space character next to both
  delimiters, so ordinary prose like "costs $5 and $7" is not swallowed as math.
- A thinking selector in the header sends `reasoning_effort` per request, off by
  default. GLM 5.2 emits visible reasoning otherwise, which is tedious to read at
  streaming speeds.
- A copy button per message, which copies the raw Markdown rather than the rendered
  HTML, and a stop button (also bound to Esc) that aborts generation while keeping
  whatever text already arrived.

## Notes, and things you may want changed

- **KaTeX is loaded from a CDN** (jsDelivr). It cannot reasonably be inlined because
  of the fonts. If it fails to load, equations degrade to showing the raw LaTeX
  instead of breaking, and re-render if it arrives late. I am happy to drop math
  rendering altogether if you would rather the UI carry no external dependency.
- The copy embedded in `ds4_server.c` is generated from `chat-ui.html` rather than
  escaped by hand, since the LaTeX regexes are backslash-heavy and one missed escape
  yields a page that compiles but renders wrongly. `tools/gen_chat_ui.py` regenerates
  it; `tools/gen_chat_ui.py --check` reports staleness without writing, if you want
  it in a pre-release check.
- `tools/test_chat_ui.js` covers the renderer (Markdown blocks, HTML escaping, math
  delimiters). It takes an optional path, so the same suite can be run against a page
  fetched from a running `ds4-server` — that is what verifies the C escaping survived
  the round trip, which compiling cannot tell you. Plain `node`, no dependencies.
- Tested on macOS / Metal against DeepSeek V4 Flash (resident) and GLM 5.2 (SSD
  streaming). The UI is model-agnostic and reads the model id from `/v1/models`.
- Purely additive: no existing file behaviour changes, and the new route only
  answers `GET /`, which was previously unhandled.


MiklosPathy:
FYI https://github.com/antirez/ds4/pull/854

PR #727, opened 2026-08-06
server: keep live KV reusable when clients strip transient metadata blocks

### LINKED ISSUE #364: KV cache invalidated on every turn by client-injected transient metadata in user messages

Flor1an-B:
Fixes #364.

Coding agents (OpenCode, Claude Code, Kilo Code/Roo Code) inject per-turn transient metadata into user message content (`<system-reminder>`, `<environment_details>`) and rebuild the next request's history without it. The live KV keeps the bytes that were actually rendered, so the replayed transcript no longer token-matches the sampled session and every user turn pays a full reprefill.

This is the same shape hidden thinking already handles: the live state is richer than the visible replay, so the existing visible-key continuation mechanism applies. After a finished turn, remember the transcript the next request is expected to render — the current one minus the transient spans — keyed to the live token frontier. A client that replays the blocks verbatim simply never matches the key and still gets exact token-prefix matching; nothing changes for it.

**Credit**: original design and implementation by @adv0r in #378. That PR now has a real conflict with `main`, not just a stale rebase: native Metal/CUDA session batching (`36cd0ca` and follow-ups) added a `server_slot` parameter to every session/checkpoint function it touches (`thinking_live_remember`/`clear`, `remember_thinking_checkpoint`, `canonicalize_tool_checkpoint`), added after #378 was opened. This PR threads the same `strip_transient_blocks()` / `should_remember_transient_checkpoint()` / `remember_transient_checkpoint()` design through the slot-aware signatures — the diff is otherwise unchanged from #378, same test names and assertions. Opening as a new PR rather than pushing to #378's branch since I don't have write access to it.

**Testing**:
- Full `make ds4_test` suite passes clean.
- Live A/B on the real quant (`ds4-server`, 2-turn conversation with a `<system-reminder>` block dropped on replay between turns, exactly as OpenCode does):
  - before (baseline, patch reverted): `live kv cache miss live=88 prompt=97 common=5 reason=token-mismatch` — full reprefill.
  - after: `transient live checkpoint remembered ... stripped=128` then, on the next turn, `thinking live continuation match=visible-prefix cached=30 prompt=54` — live continuation, no reprefill.
- Re-verified a 2-turn tool-calling round trip with *no* transient blocks present still gets exact-DSML-replay continuation unchanged (18 new tokens prefilled instead of a full 406-token reprefill), confirming the `canonicalize_tool_checkpoint` → remember-or-clear restructuring is behavior-preserving when nothing is stripped.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -10747,14 +10747,18 @@ static char *strip_transient_blocks(const char *text) {
             p = open + 1;
             continue;
         }
-        /* Clients append these blocks with separator whitespace that vanishes
-         * from the replay together with the block, so drop it from the key
-         * as well. */
+        /* Clients wrap these blocks with separator whitespace that vanishes
+         * from the replay together with the block, so drop it from the key as
+         * well.  A block may be appended after the user's content (OpenCode's
+         * <environment_details>, separator before) or prepended before it
+         * (Claude Code's <system-reminder>, separator after), so trim on both
+         * sides. */
         size_t keep = (size_t)(open - p);
         while (keep > 0 && isspace((unsigned char)p[keep - 1])) keep--;
         buf_append(&out, p, keep);
         stripped = true;
         p = close;
+        while (*p && isspace((unsigned char)*p)) p++;
     }
     if (!stripped) {
         buf_free(&out);
@@ -16855,6 +16859,21 @@ static void test_strip_transient_blocks(void) {
         "a<environment_details>1</environment_details>"
         "b<system-reminder>2</system-reminder>c");
     TEST_ASSERT(s && !strcmp(s, "abc"));
+    free(s);
+
+    /* Prepended block: Claude Code injects the block *before* the user's own
+     * content, with a trailing separator newline that vanishes from the replay
+     * together with the block.  Drop it too, symmetric with the leading side. */
+    s = strip_transient_blocks(
+        "<system-reminder>plan mode</system-reminder>\nHere is a document");
+    TEST_ASSERT(s && !strcmp(s, "Here is a document"));
+    free(s);
+
+    /* Consecutive prepended blocks separated by whitespace collapse cleanly. */
+    s = strip_transient_blocks(
+        "<system-reminder>1</system-reminder>\n"
+        "<environment_details>2</environment_details>\ncontent");
+    TEST_ASSERT(s && !strcmp(s, "content"));
     free(s);
 
     /* Unterminated block stays in place. */
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -10699,6 +10699,71 @@ static char *build_responses_visible_assistant_suffix(const request *r,
     return buf_take(&suffix);
 }
 
+/* Coding agents inject per-turn metadata blocks into the user message content
+ * they send (Kilo Code wraps workspace state in <environment_details>,
+ * OpenCode and Claude Code inject <system-reminder> notes), then rebuild the
+ * next request's history without them.  The live KV keeps the bytes that were
+ * rendered, so the replayed transcript no longer token-matches the sampled
+ * session and every user turn pays a full reprefill.
+ *
+ * This is the same shape as hidden thinking: the live state is richer than
+ * the visible replay.  So the same visible-key continuation applies.  After a
+ * finished turn, remember the transcript the next request is expected to
+ * render -- the current one minus the transient spans -- keyed to the live
+ * token frontier.  A client that replays the blocks verbatim simply never
+ * matches the key and still gets exact token-prefix matching. */
+static const char * const transient_block_tags[] = {
+    "environment_details",      /* Kilo Code / Roo Code workspace state */
+    "system-reminder",          /* OpenCode / Claude Code injected notes */
+};
+
+/* Return a copy of text with every <tag>...</tag> transient span removed, or
+ * NULL when no span was found.  Unterminated blocks stay in place: a wrong
+ * guess only produces a visible key that never matches. */
+static char *strip_transient_blocks(const char *text) {
+    if (!text) return NULL;
+    buf out = {0};
+    bool stripped = false;
+    const char *p = text;
+    while (*p) {
+        const char *open = strchr(p, '<');
+        if (!open) break;
+        const char *close = NULL;
+        for (size_t t = 0; t < sizeof(transient_block_tags) /
+                               sizeof(transient_block_tags[0]); t++)
+        {
+            const char *tag = transient_block_tags[t];
+            const size_t tag_len = strlen(tag);
+            if (strncmp(open + 1, tag, tag_len) || open[1 + tag_len] != '>')
+                continue;
+            char end_tag[64];
+            snprintf(end_tag, sizeof(end_tag), "</%s>", tag);
+            const char *end = strstr(open + 1 + tag_len + 1, end_tag);
+            if (end) close = end + strlen(end_tag);
+            break;
+        }
+        if (!close) {
+            buf_append(&out, p, (size_t)(open - p) + 1);
+            p = open + 1;
+            continue;
+        }
+        /* Clients append these blocks with separator whitespace that vanishes
+         * from the replay together with the block, so drop it from the key
+         * as well. */
+        size_t keep = (size_t)(open - p);
+        while (keep > 0 && isspace((unsigned char)p[keep - 1])) keep--;
+        buf_append(&out, p, keep);
+        stripped = true;
+        p = close;
+    }
+    if (!stripped) {
+        buf_free(&out);
+        return NULL;
+    }
+    buf_puts(&out, p);
+    return buf_take(&out);
+}
+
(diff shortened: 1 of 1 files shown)


GreatMark:
Found and fixed a case where this PR's key never matches: **blocks injected *before* the user's own content** (Claude Code's `<system-reminder>` placement), as opposed to appended blocks (OpenCode's `<environment_details>`).

## Symptom

M3 Ultra, Metal, MXFP4 Flash 0731, this PR merged on `84cc882`. Turn 1 user content is `<system-reminder>...</system-reminder>\n` + document; turn 2 replays history without the block. The remember side works (`transient live checkpoint remembered ... stripped=4260`), but turn 2 still misses (`live kv cache miss ... common=9 reason=token-mismatch`) and pays the full reprefill.

## Root cause (byte-level, from `--trace` rendered prompts)

`strip_transient_blocks()` consumes separator whitespace only **before** a stripped block. A prepended block's trailing separator newline stays in the key:

```text
key kept:  ...<｜User｜>\nHere is a document excerpt...
turn 2:    ...<｜User｜>Here is a document excerpt...
```

One leftover `\n` right after the user-turn marker, so `byte_prefix_match` fails at that byte forever.

## Fix that worked here

Consume trailing whitespace after the block too, symmetric with the existing leading-side logic:

```c
         buf_append(&out, p, keep);
         stripped = true;
         p = close;
+        while (*p && isspace((unsigned char)*p)) p++;
     }
```

Same safety argument as the PR's own: a wrong guess only produces a visible key that never byte-matches and falls back to the existing full-reprefill behaviour; a match is still verified byte-for-byte against the rendered request, so no incorrect KV reuse is possible.

## After the fix

| metric | before | after |
|---|---|---|
| turn 2 TTFB | 2.59 s | **0.24 s** |
| turn 2 prefill | 1,342 tokens (full) | 10 tokens (`thinking live continuation match=visible-prefix cached=1371 prompt=1381`) |

Appended-block scenarios keep working (the existing test assertions in `./ds4_test --server` all pass with the change), and adjacent stripped blocks separated by whitespace now collapse cleanly as well.

*Measured locally; an AI agent helped bisect the byte mismatch and run the A/B harness. Log lines and trace output are ds4-server's own.*

Flor1an-B:
Thanks — good catch, and the byte-level trace made it trivial to confirm.

Reproduced the prepended-block miss as a failing unit test first (a single `<system-reminder>…</system-reminder>\n<content>`, plus consecutive prepended blocks), then fixed it exactly as you diagnosed: consume the separator whitespace *after* the close tag too, symmetric with the existing leading-side trim. The safety property is unchanged — a wrong key only ever produces a visible-prefix miss and falls back to the existing full-reprefill; matches are still verified byte-for-byte, so no incorrect KV reuse is possible.

The existing tests only exercised appended placement, which is exactly why this slipped, so the regression coverage now includes both a single prepended block and consecutive prepended blocks (which now collapse cleanly). Warning-free build, `./ds4_test --server` green. Pushed as 5819e54.

andrewmy:
@antirez any chance to get this into main? 🙏

Flor1an-B:
Rebased onto current `main` (`0aaea5a`) — the conflict was purely mechanical: the native Metal/CUDA session-batching work landed after this branch was opened and added a `const request *req` parameter to `thinking_live_remember()` (used for the image-key lookup), which this PR's new `strip_transient_blocks()` call sites hadn't picked up yet. Also folded in a small independent fix from the same batching work (`build_qwen_tool_turn_visible_text`/`remember_qwen_tool_turn_visible_checkpoint`, a new sibling checkpoint path added alongside this one) — no logic conflict, just two features landing in the same function.

Re-verified after the rebase:
- Full `make test` green (same pre-existing, unrelated skips gated on a Qwen3.8 test fixture not being present in this environment).
- Live end-to-end on a real DeepSeek V4 Flash instance: sent a 2-turn conversation where turn 1's user message carries a `<system-reminder>...</system-reminder>` block (as OpenCode/Claude Code inject) and turn 2's client-rebuilt history omits it, exactly the scenario this PR targets. Server log confirms the new path fires (`"transient live checkpoint remembered ... visible=192 stripped=128"`) and turn 2 reuses the cached prefix (`usage.cached_tokens=34` of 49 prompt tokens, only 15 new tokens processed) instead of a full reprefill.

@andrewmy — sorry for the silence, this should be ready to merge now.

PR #729, opened 2026-08-06
server: infer JSON types for GLM tool-call arguments

### LINKED ISSUE #569: server: GLM tool-call parser stringifies all argument values (numbers/arrays become strings), breaking schema-validating clients

Flor1an-B:
Fixes #569.

GLM's `<arg_key>`/`<arg_value>` wire format carries no type attribute, unlike DSML's `string="true|false"` that the DeepSeek path already honors. Every argument was therefore quoted as a JSON string regardless of its actual type — numbers became `"10"`, arrays became `"[]"`, booleans became `"true"`. Any client that validates tool-call arguments against the tool's JSON schema (zod/ajv-style validators) rejects the call.

**Fix** (the heuristic option from the issue): a value that fully parses as one JSON value (number, bool, null, array, or object) with nothing left over after trimming whitespace is passed through raw instead of being quoted — reusing the same whole-string validation semantics `json_minify_raw_value`'s existing callers already rely on for DSML's explicit `string="false"` path. A plain string is unaffected and still gets quoted exactly as before.

Root cause and the fix approach were identified by @solodddd in #569, including the exact repro (`WebSearch` with a string/number/array argument mix) that the new test is built from.

**Testing**: full `make ds4_test` suite passes. Added `test_json_text_is_whole_value` (the new helper in isolation) and `test_parse_glm_tool_call_infers_json_value_types` (the exact issue repro end-to-end through `parse_generated_message_ex_for_syntax`, asserting `max_results` and `exclude_domains` are now unquoted while `query` stays a quoted string). Updated the one pre-existing assertion that encoded the old behavior (`test_parse_glm_tool_call_message` expected `"timeout": "10"`, now `"timeout": 10`).

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -433,6 +433,20 @@ static char *json_minify_raw_value(const char *json) {
     return buf_take(&b);
 }
 
+/* True when text is, in its entirety (ignoring surrounding whitespace), a
+ * single valid JSON value -- a number, boolean, null, array, or object.  Used
+ * to recover type information for untyped wire formats: GLM's <arg_value>
+ * carries no type attribute, unlike DSML's string="true|false", so a value
+ * that round-trips as JSON is passed through raw instead of being quoted. */
+static bool json_text_is_whole_value(const char *text) {
+    if (!text) return false;
+    const char *p = text;
+    json_ws(&p);
+    if (!json_skip_value(&p)) return false;
+    json_ws(&p);
+    return *p == '\0';
+}
+
 static bool json_content(const char **p, char **out) {
     json_ws(p);
     if (**p == '"') return json_string(p, out);
@@ -5143,7 +5157,11 @@ static bool parse_glm_generated_message_ex(const char *text,
             }
             char *raw_value = xstrndup(p, (size_t)(value_end - p));
             char *value = dsml_unescape_text(raw_value);
-            tool_call_json_args_add(&args, key, value, "true");
+            /* GLM's wire format carries no type attribute (unlike DSML's
+             * string="true|false"), so a value that fully parses as JSON is
+             * passed through raw; anything else is a string. */
+            const char *value_type = json_text_is_whole_value(value) ? "false" : "true";
+            tool_call_json_args_add(&args, key, value, value_type);
             free(key);
             free(raw_value);
             free(value);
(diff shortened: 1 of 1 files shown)


solodddd:
Verified on M3 Ultra 512GB (Metal) — this fixes the #569 repro for the numeric case.

Environment:
- M3 Ultra, 512 GiB RAM, Metal backend
- Model: GLM-5.2-UD-Q2_K_RoutedQ2K.gguf
- Branch: pr729 @ 93eb22b

Results:
- Build (make -j8): clean, exit 0
- make ds4_test + ./ds4_test --server: all pass
- End-to-end with the exact #569 payload (WebSearch, query + max_results + exclude_domains):

Before this PR: "max_results":"10"
After:            "max_results":10  (JSON number)

Server log shows a single WebSearch tool call with arguments
{"query":"ByteDance inference chips","max_results":10}.

One honest caveat: in this run the model did not emit exclude_domains,
so the array case wasn't exercised end-to-end here — the numeric
(max_results) case from #569 is confirmed fixed. Happy to run a variant
that forces an array argument if that would help.

Thanks for the fix — and for crediting the original report in #569.

PR #730, opened 2026-08-07
ROCm: fuse QKV RMSNorm and KV RoPE

jyatesdotdev:
## What this does

The ROCm backend implemented the "Q/KV weighted RMS normalization plus KV rotary embedding" entry point as two kernel launches. This pull request replaces the pair with a single fused kernel for shapes up to 256 rotary pairs, removing one kernel launch and its synchronization from every decoded layer.

- Shapes larger than 256 rotary pairs keep the existing two-launch path.
- Setting `DS4_ROCM_DISABLE_QKV_KV_ROPE_FUSION=1` at process start restores the two-launch path, as a diagnostic rollback.
- New differential tests cover dense, inverse, scaled, YaRN, zero-rotation, the 256-pair boundary, the 257-pair fallback, and invalid shapes. Each case is verified byte for byte against the unfused path.

The DeepSeek V4 graph and the public GPU API are unchanged: the graph already calls this entry point, whose ROCm implementation simply chained the two launches. The change is independent of expert quantization and applies to both Q4_K and MXFP4 models.

## Performance

A balanced comparison in one binary, ordered fallback → fused → fused → fallback, each arm a fresh process with a discarded warmup and 800 measured decode tokens:

- server decode: 14.5749 → 14.6032 tokens/s (+0.19%)
- sampled QKV stage time: −1.38%
- sampled total GPU stream time: −0.22%

Both paired comparisons favored the fused kernel. The gain is small because this stage is a small share of decode time; the point of the change is removing a launch per decoded layer without changing any output. The runs used the version-3 TCP transport, which this change does not touch. Absolute throughput was measured on an integration tree that also contains #656; the same-binary toggle isolates the fusion itself.

## Correctness

On Radeon 8060S (gfx1151):

- Warning-free ROCm builds on both test machines.
- `make HIPCC=/opt/rocm-therock/bin/hipcc test-rocm-qkv-fusion` passes all twelve positive cases and the invalid-shape cases, byte for byte.
- Full-model runs with the fusion enabled and disabled produced identical output across all 40 measured responses.
- The standalone commit also passes a warning-clean strict CPU build, and its ROCm implementation and test payload are byte-identical to the integration source that ran the full-model comparisons.

## Scope

This is a ROCm attention-launch optimization only. It does not include the MXFP4 expert kernels (#656), transport work (#820), or any batching changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@
 /tests/test_mxfp4_dot
 /tests/test_mxfp4_metal
 /tests/test_q4k_dot
+/tests/test_rocm_qkv_fusion
 /tests/test_sampling
 *.o
 *.dSYM/
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,7 @@ CPU_CORE_OBJS = ds4_cpu.o ds4_distributed.o ds4_tp.o ds4_ssd.o ds4_layer_pack.o
 CUDA_LDLIBS ?= -lm -Xcompiler -pthread -L$(CUDA_HOME)/targets/sbsa-linux/lib -L$(CUDA_HOME)/lib64 -lcudart -lcublas
 HIPCC ?= $(shell command -v hipcc 2>/dev/null || echo /opt/rocm/bin/hipcc)
 ROCM_ARCH ?= gfx1151
+ROCM_HOST_CFLAGS ?= -fPIC
 ROCM_CFLAGS ?= -O3 -ffast-math -g -fno-finite-math-only -pthread -D__HIP_PLATFORM_AMD__ -Wno-unused-command-line-argument --offload-arch=$(ROCM_ARCH)
 ROCM_LDLIBS ?= -lm -pthread -lhipblas -lhipblaslt
 DS4_LINK ?= $(NVCC) $(NVCCFLAGS)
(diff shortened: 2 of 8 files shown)


jyatesdotdev:
@antirez — rebased onto current `main` (`84cc882`) and retargeted; the branch is now a single commit (`7337635`) with no `ds4f-mxfp4` ancestry, and the PR diff shows only the eight fusion files.

Fresh QA on the rebased commit (Strix Halo pair, Radeon 8060S / gfx1151, TheRock toolchain):

- `make strix-halo`: warning-free for both this branch and `main` @ `84cc882` as control. Both were built with `-fPIC` added to `CFLAGS` — main currently lacks the `ROCM_HOST_CFLAGS` fix, which ships separately in #656's first commit; called out so the builds reproduce.
- `make test-rocm-qkv-fusion`: all sections PASS (norm-only, fused norm+RoPE bit-exact differential cases, invalid-shape coverage).
- End-to-end on the release Q4_K-experts Flash GGUF, two-node distributed (CLI coordinator/worker), greedy 200-token probe, fused branch vs unmodified main: **byte-identical output**, and speed parity within noise (24.78/13.27 t/s vs 24.75/13.24 t/s prefill/decode) — as expected, the win here is one fewer launch per layer per step rather than a measurable rate change at this scale.

From the original A/B on the ds4f-mxfp4 lineage (kept in the summary above): server decode +0.19%, sampled QKV time −1.4%, `DS4_ROCM_DISABLE_QKV_KV_ROPE_FUSION=1` remains as the diagnostic rollback.

jyatesdotdev:
Closing this pull request to cut down the number of open submissions from me. The fused QKV normalization and RoPE change is being folded into a single performance branch rebased on current main, together with my other ROCm work, and will return as one combined submission with fresh measurements against the current tree.

Issue #695, opened 2026-08-05
DSpark scheduler break-even model appears not to account for replay cost (Metal, net -1327ms despite 70% accept rate)


datanerdie:
On Metal (M4 Max, Flash q2-q4, 0731 DSpark support GGUF, `6747e77`), DSpark is consistently net-negative on generation:

| corpus | baseline gen | `--dspark` | change |
|---|---:|---:|---:|
| prose, 9485 tok | 29.31 t/s | 23.02 t/s | **-21.5%** |
| prose, 28790 tok | 27.82 t/s | 23.10 t/s | **-17.0%** |
| code, 9452 tok | 29.53 t/s | 22.90 t/s | **-22.4%** |

Reproducible to 0.2% across reps; prefill is unaffected as expected.

This is **not** the ROCm no-draft failure in #597 — here DSpark proposes normally and its proposals are usually accepted. `DS4_DSPARK_STATS=1`, prose:

```
cycles=88 proposed=44 accepted_draft=31 accept_rate=70.45% avg_accept=0.352
full=14 partial=2 miss_first=7 no_draft=65 scheduler_skips=52
draft_len_hist=1:10,2:6,3:6,4:1   accepted_len_hist=0:72,1:6,2:5,3:5
time_ms propose=332.0 verify=964.4 replay=1063.1 spec_total=2039.9
        saved=1045.3  net_saved=-1326.6
```

The economics: each accepted token saves ~34 ms (one baseline token at 29.3 t/s), and 31 were accepted, giving `saved=1045 ms`. But `replay=1063 ms` over 44 proposals is **~24 ms per speculation cycle — roughly one full token**. So a cycle must accept more than one token just to pay for its own replay, while `avg_accept` is 0.35.

The scheduler is clearly trying to avoid this: 52 of 88 cycles were skipped, and the no-draft pause logic fires. But it still ends up net-negative, which suggests **its break-even model may not include the replay term** — the largest single cost in the breakdown.

Two observations that may help localise it:

1. **Code is worse than prose**, contradicting the README guidance that predictable continuations benefit most: accept_rate 62.22% vs 70.45%, avg_accept 0.308 vs 0.352, 77/91 vs 72/88 cycles accepting zero. Corpus was a 9452-token prefix of `ds4.c` itself, length-matched to the prose prompt within 0.3%.
[...]

PR #746, opened 2026-08-08
dspark: opt-in fast commit on full accept, matching MTP's default Metal path

Flor1an-B:
Follow-up to #695 with a working, opt-in prototype and multi-prompt benchmark data, not just analysis.

## Root cause (recap from #695)

DSpark unconditionally rolls back the checkpoint and sequentially replays every accepted draft token through the exact single-token decode kernel, even on a full accept — `ds4.c:61711-61714`: *"replay every accepted draft through ordinary decode, including partial accepts"*. That replay costs close to one full target-model forward pass per accepted token, canceling most of the savings speculation is supposed to provide.

## What this PR does

MTP's default Metal path already makes a different call. Tracing it turned up:

```c
#ifdef DS4_ROCM_BUILD
    const bool prefer_decode2_exact = true;
#else
    const bool prefer_decode2_exact = false;   // Metal
#endif
```

`metal_graph_verify_decode2_exact` — the exact, bit-identical-to-decode verifier — is only *preferred* on ROCm. On Metal, MTP defaults to the same batched `metal_graph_verify_suffix_tops` kernel DSpark's verify already uses, and **on a full accept it commits directly from that batched pass with no rollback and no replay**.

This PR applies the identical trade-off to DSpark's full-accept case, gated behind `DS4_DSPARK_FULL_ACCEPT_FAST_COMMIT` (opt-in, default off — DSpark doesn't yet ship this as a default the way MTP does, so I didn't want to change default behavior silently). Partial accepts and TP are untouched: TP falls through to the existing rollback+replay unconditionally (I haven't written or tested a worker-side commit protocol for this path), and partial accepts keep replaying exactly as before.

## The trade-off, measured, not assumed

This is not free — and I want to be upfront about that rather than presenting it as a clean win. The batched kernel can differ from the exact single-token decode kernel by enough to flip a near-tied greedy token, exactly as documented on `metal_graph_verify_decode2_exact`. I confirmed this is real by diffing full generated output against plain baseline, not just trusting the existing `dspark-verify-depth` unit test (see below for why that test can't see it).

**`tests/dspark_acceptance_fixture.sh`** (the project's own byte-exact `cmp -s` fixture, 32-token generations, 4 diverse prompts) passes with the flag on in every case, including two with a *positive* net_saved:

```
hello: net_saved -95.7ms -> +18.7ms
redis: net_saved -307.2ms -> +52.2ms
```

**New `tests/dspark_fast_commit_bench.py`** (companion script, longer ~300+ token free-running generations, where a short fixture can't accumulate compounding divergence):

| prompt | flag-off t/s | flag-on t/s | net_saved off→on | diverges at word |
|---|---|---|---|---|
| essay | 30.53 | 38.60 (+26%) | -2735ms → -227ms | 16 |
| code | 29.40 | 37.12 (+26%) | -4213ms → -596ms | none |
| math | 29.91 | 45.78 (+53%, faster than the 38.57 t/s plain baseline) | -1036ms → **+619ms** | 16 |
| dialogue | 28.78 | 30.52 (+6%) | -3646ms → -1768ms | 51 |

Flag-on is faster than flag-off in every case tested, and the net cost of the DSpark mechanism itself drops sharply everywhere, going net-positive on the math prompt. But 3 of 4 longer prompts do diverge from the plain-greedy baseline text partway through (confirmed by full-text diff, not just a stats anomaly).

I checked whether this is a *new* risk this PR introduces: it isn't. On the identical essay prompt, MTP's own already-shipped default behavior diverges from the same baseline **at the same point, with the same inserted phrase**, because it's the same batched kernel causing the same near-tied-logit flip. This PR doesn't add a new risk category to the codebase — it applies to DSpark a trade-off MTP already ships by default on Metal.

## Why the existing `dspark-verify-depth` test doesn't catch this

I ran it both ways (`DS4_TEST_DSPARK=... ./ds4_test --dspark-verify-depth`, flag off and on): `worst_argmax_gap=0.000` in both cases, 128 real speculative cycles, draft depth 5. That test teacher-forces every committed token through normal decode and resets ground truth each cycle, so a small per-step divergence never gets the chance to compound the way it does in a real free-running generation. It's a good test for "did the verifier commit something the target model would never plausibly produce," but not for "does the full generation still match exact greedy decode." `tests/dspark_fast_commit_bench.py` is meant to fill that specific gap for this flag going forward.

## Testing

- `make ds4_test`: full suite passes, flag unset by default (code path unreachable, zero behavior change from main).
- `dspark-verify-depth`: passes with the flag both off and on.
- `dspark_acceptance_fixture.sh`: passes with the flag on (`cmp -s` byte-exact) on all 4 short prompts.
- `dspark_fast_commit_bench.py`: new script, data above.

This is deliberately opt-in and not proposed as the new default — happy to change that if you'd rather it match MTP's default-on Metal behavior, or to leave it opt-in for people who want the speed and are fine with the same trade-off MTP already makes silently.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -48380,6 +48380,25 @@ static bool ds4_dspark_scheduler_timing_enabled(void) {
            ds4_dspark_scheduler_max_extra_saved_ratio_milli() != 0;
 }
 
+/* On a full accept, the batched verify pass has already produced correct raw
+ * KV / compressor / indexer state for every drafted position -- the same
+ * batched attention path a plain (non-speculative) prefill chunk uses. MTP's
+ * default Metal path already trusts that state directly on a full accept
+ * instead of paying for a rollback and a sequential exact-kernel replay of
+ * every drafted token (see the DS4_ROCM_BUILD / prefer_decode2_exact split
+ * a few hundred lines into the MTP speculative-argmax path). This applies
+ * the same trade-off to DSpark, which currently never does.
+ *
+ * The trade-off is real, not just a performance knob: the batched kernel can
+ * differ from the exact single-token decode kernel by enough to flip a
+ * near-tied greedy token (documented on metal_graph_verify_decode2_exact),
+ * and unlike MTP this is not yet the shipped default here. Opt-in until
+ * there's more mileage on it than one contributor's benchmark. */
+static bool ds4_dspark_full_accept_fast_commit_enabled(void) {
+    const char *env = getenv("DS4_DSPARK_FULL_ACCEPT_FAST_COMMIT");
+    return env && env[0] && strcmp(env, "0") != 0;
+}
+
 static void ds4_session_dspark_scheduler_reset(ds4_session *s) {
     if (!s) return;
     s->dspark_sched_cycles = 0;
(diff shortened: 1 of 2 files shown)


Flor1an-B:
Fresh measured before/after on this branch (rebased on `b030961`), Apple M5 Max, Metal, `DeepSeek-V4-Flash-DSpark-support.gguf` (stages=3), greedy, `-n 256`, same prompt for both runs. Baseline (no speculation) on the same box is **40.14 t/s**.

| Metric (`DS4_DSPARK_STATS`, whole run) | flag OFF | flag ON (`DS4_DSPARK_FULL_ACCEPT_FAST_COMMIT=1`) |
|---|---|---|
| generation | 34.22 t/s | **37.79 t/s** (+10.4%) |
| replay ms | 822.6 | **224.6** (−73%) |
| saved ms | 819.0 | 1642.6 |
| net_saved ms | −1064.8 | **−365.7** |
| full accepts | 19 | 30 |

**Why it moves:** without the fast commit, `replay (823 ms) ≈ saved (819 ms)` — every accepted run is decoded a second time to rebuild state, so the acceptance benefit is almost exactly cancelled. Reading the verified logits row on a full accept (the same thing MTP's default Metal path already does) drops replay to 225 ms and roughly doubles realized `saved`.

**Honest ceiling:** this reduces the *loss* rather than making DSpark faster than plain decode here (37.79 vs 40.14). The remaining gap is `propose + verify > saved`, and even eliminating the draft-chain cost entirely only reaches break-even — the binding constraint is accepted-draft length (`avg_accept` ≈ 0.15–0.63; `accepted_len_hist` concentrated at 0–1), i.e. proposer/model quality, not this path.

**On keeping it opt-in:** the reason given for the flag is that a full-accept commit could diverge from the sequential greedy token. I instrumented that separately (turns the teacher-forced verify-depth check into a divergence rate + severity histogram) and measured **0 divergence over 1408 teacher-forced positions** on this config (128 confident + 256 + 1024 open-ended), `worst_gap` 0.000, with the multi-row verify path exercised. Since the fast commit reads the *verified* logits row rather than re-deriving anything, that's expected — which raises the question of whether this should be the default on Metal rather than opt-in. Happy to follow whichever you prefer.

Flor1an-B:
Closing this — superseded by `0e89a0e "dspark: commit accepted verifier state directly"` on main, which lands the same idea (skip the sequential replay by committing the verified accept state directly) **as the default path rather than behind an opt-in env flag**. That is the better shape: this PR's `DS4_DSPARK_FULL_ACCEPT_FAST_COMMIT` was a permanent semantic variant behind a flag, which AGENT.md explicitly discourages.

The measured before/after posted earlier still stands as validation of the direction — on M5 Max / Metal / Flash q2-q4 the full-accept fast commit took DSpark from 34.22 → 37.79 t/s (+10.4%), replay 823 → 225 ms — so making it the default is the right call. Thanks for picking it up directly.

Issue #733, opened 2026-08-07
DSpark net-negative on M3 Ultra + MXFP4 despite 82.8% accept rate — independent confirmation of #695 replay/verify overhead


lifodetails:
## Summary

Adding a cross-configuration data point for #695: DSpark is net-negative on **M3 Ultra 256GB + MXFP4** (the recommended quantization), even though the draft acceptance rate is **82.8%**. This confirms the core finding of #695 from a different hardware generation and quantization format: the bottleneck is replay + verify overhead, not draft quality.

## Environment

| Item | Value |
|------|-------|
| Chip | Apple M3 Ultra (256 GB unified memory, 819 GB/s) |
| Engine | ds4 commit `b030961` |
| Model GGUF | `DeepSeek-V4-Flash-MXFP4Experts-F16HC-F16Compressor-F16Indexer-Q8Attn-Q8Shared-Q8Out-chat-v2-mxfp4-0731.gguf` (145 GB) |
| DSpark support | `DeepSeek-V4-Flash-DSpark-support-0731.gguf` (5.6 GB) |
| Quantization | MXFP4 experts (native, bit-for-bit) + F16/Q8 non-expert |
| Context | 8192, `--nothink --temp 0` |
| Prompt | `"Write a Python quicksort function with comments."` (100 tokens) |

## Controlled Experiment (4 modes, same prompt)

| Mode | Generation (t/s) | Decode vs baseline |
|------|-----------------|-------------------|
| Baseline (no DSpark) | 39.78 | — |
| `--dspark-strict` (loaded, target-only) | 39.17 | -1.5% (loading overhead ≈ 0) |
| `--dspark` (confidence 0.7, default) | 25.47 | **-36%** |
| `--dspark --dspark-confidence 0` (forced 5-token blocks) | 21.58 | **-46%** |

The strict experiment isolates the overhead: loading the DSpark support model costs ~1.5% (within noise). The regression is entirely in the runtime decode path.

### Default confidence (0.7)

```
cycles=46 proposed=58 accepted_draft=48 accept_rate=82.76% avg_accept=1.043
full=12 partial=4 miss_first=2 no_draft=28 scheduler_skips=23
time_ms propose=379.8 verify=1236.4 replay=1206.8 spec_total=2460.6
        target=1571.3 saved=1631.7 net_saved=-1208.7
draft_len_hist=1:2,2:4,3:5,4:2,5:5
accepted_len_hist=0:30,1:3,2:3,3:5,4:1,5:4
[...]

Issue #731, opened 2026-08-07
DSpark: per-accepted-token replay cancels ~100% of the speculative saving (M3 Ultra measurements, net_saved = -3.98 s)


justinz0920-png:
## Summary

On a Mac Studio M3 Ultra, DSpark is a consistent **~35% throughput loss**, not a gain. `DS4_DSPARK_STATS=1` shows why: the rollback+replay path introduced by af80694 and 7fb2830 costs almost exactly what the accepted drafts saved, so the theoretical ceiling for net gain is ~0 — and then propose + verify are charged on top.

I understand from the comment in af80694 that this is a known and deliberate trade ("*this makes DSpark's already-negative throughput case worse, but correctness comes first*"). This report is not a request to revert it. It is an attempt to **quantify** how large that cost is on real hardware, since the comment only describes it qualitatively, and to point out that the replay cost appears to scale with the acceptance rate — meaning the better the drafter performs, the more the replay takes back.

## Environment

| | |
|---|---|
| Machine | Mac Studio M3 Ultra, 80-core GPU, 512 GB unified memory |
| Backend | Metal. `Metal 4 tensor API disabled for pre-M5/pre-A19 devices` |
| ds4 | `b030961` (2026-08-05), `make` |
| Model | `DeepSeek-V4-Flash-Q4KExperts-F16HC-F16Compressor-F16Indexer-Q8Attn-Q8Shared-Q8Out-chat-v2-imatrix-0731.gguf` |
| DSpark | `DeepSeek-V4-Flash-DSpark-support-0731.gguf` — `stages=3 block=5 markov_rank=256 tensors=81 missing=0 invalid=0` |

## Measured throughput

Same prompt, same binary, `--temp 0 --nothink -n 256`, only `--mtp ... --dspark` added:

| Workload | no DSpark | with DSpark | ratio |
|---|---:|---:|---:|
| Italian prose continuation (6244-token prompt) | 33.54 t/s | 21.77 t/s | **0.649x** |
| C code generation (26-token prompt) | 36.50 t/s | 23.37 t/s | **0.640x** |

Prefill was unchanged (554.26 → 553.31 t/s), as documented. Generated text was identical in both runs, so acceptance/verification is behaving correctly.

The two ratios being nearly identical across very different content is itself informative: this is not acceptance-rate variance, it is a fixed structural overhead.

## `DS4_DSPARK_STATS=1` breakdown

From the C code-generation run:

```
[...]

PR #761, opened 2026-08-10
rocm: implement the two DSpark kernels (draft, acceptance fixture still fails)

### LINKED ISSUE #597: DSpark never produces a draft on ROCm / gfx1151: 1197 cycles, proposed=0, prop_hidden=0.000

deepu105:
Fixes #597.

## What this does

`ds4_gpu_attention_noncausal_raw_batch_heads_tensor` and `ds4_gpu_dspark_markov_argmax_tensor` are stubbed to `return 0` in `ds4_rocm_unavailable.cu`. So on ROCm, DSpark loads, logs `DSpark target-hidden capture enabled`, and then proposes nothing at all for the whole session. `metal_graph_eval_dspark_stage_block()` gets 0 back from the noncausal attention, `stage_chain_done` stays 0, and the proposer returns no draft every cycle.

There is no warning, so from the outside it just looks like DSpark is not helping:

```
cycles=107 first_tokens=107 proposed=0 accepted_draft=0 accept_rate=0.00%
no_draft=107 scheduler_skips=81 net_saved=-68.224
```

This PR ports both kernels from the CUDA versions. They are plain FP32 with shared memory and `__syncthreads()`, no warp shuffles or cub, so the kernel bodies are unchanged. The wrappers resolve weights through `cuda_model_range_ptr()` like the other ROCm attention launchers, and the markov wrapper drops the CUDA device-tier save/restore since the ROCm backend is single device.

With this, the proposer runs end to end:

```
noncausal_attn=ok stage_block=ok stage_chain=ok base_logits=ok markov=ok proposal_len=5
cycles=128 proposed=78 accepted_draft=63 accept_rate=80.77%
```

## Why this is a draft

**It fails `tests/dspark_acceptance_fixture.sh`.** The `hello` case diverges from greedy:

```
baseline: Hello! How can I help you today?
dspark:   Hello! How can I assist you today?
```

I do not think the divergence is in these two kernels, and I would rather flag it than paper over it. What I checked:

1. The noncausal kernel is numerically correct. I also ported the `DS4_DSPARK_VERIFY_NONCAUSAL` cross-check, and against the double-precision CPU reference it gives `max_abs=5.96e-06 max_rel=6.8e-04` over three sampled calls. Thats ordinary fp32 error.
2. `DS4_DSPARK_NO_MARKOV=1` still diverges, so its not the markov argmax either.
3. **`--dspark-strict` diverges too.** That one is the interesting part. Strict keeps target-only decode and never uses a proposal, yet the output still changes. So just running the draft stage chain is enough to perturb the target decode, independent of whether anything gets accepted.

That last point suggests something in the shared stage-chain path leaves state behind that the target run then picks up, rather than anything ROCm specific. But I dont know this code well enough to say where, and you will spot it much faster than I will. Happy to keep digging if you point me at a suspect.

## Performance

Correctness aside, DSpark is a net loss on this machine even at 80% acceptance:

| config | decode |
|---|---|
| `--dspark` off | 13.47 t/s |
| `--dspark` on | 7.71 t/s |

The breakdown:

```
propose=1856.9   (prop_chain=1264.0  prop_logits=440.2  prop_markov=57.5)
verify=9055.6    (verify_layer=9026.5)
replay=4707.2
spec_total=13817.6  target=10244.5  saved=4705.2  net_saved=-10969.3
```

This looks like the same replay overhead already reported in #695, #731 and
#733, not anything ROCm specific: drafting is cheap, and the per-accepted-token
replay added in af80694 to restore greedy identity eats the saving. So this is
a third data point on non-Metal hardware rather than a new problem.

Note the default path is unchanged. Without `--dspark` this build decodes at 13.47 t/s, same as before the patch.

## Testing

Machine: AMD Ryzen AI Max+ 395, Radeon 8060S (gfx1151), 121 GB unified, ROCm, Linux 6.17.
Model: `DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix-0731.gguf` (81 GB) with `DeepSeek-V4-Flash-DSpark-support-0731.gguf`.

```sh
make strix-halo -j32

DS4_BIN=./ds4 DS4_DSPARK_MODEL=<flash-0731> DS4_DSPARK_SUPPORT=<dspark-support-0731> \
  sh tests/dspark_acceptance_fixture.sh          # fails on `hello`, see above

DS4_DSPARK_VERIFY_NONCAUSAL=1 ./ds4 -m <model> --mtp <support> --dspark --temp 0 --nothink --tokens 32 -p Hello
DS4_DSPARK_STATS=1 ./ds4 -m <model> --mtp <support> --dspark --temp 0 -n 200 -p <code prompt>
```

I could not run `make test`: `ds4_test` links through `$(NVCC)` on non-Darwin, so it does not build on a ROCm only box. Might be worth a rocm branch there at some point.


deepu105:
I'll try to pin down where the divergence occurs. @antirez any pointers would be appreciated

deepu105:
this seems to be already fixed on main between my fetch and PR :rofl:

PR #765, opened 2026-08-10
Route batched-session requests to a slot whose KV cache they reuse

kmike:
Hey! I've been using ds4, and it seem to have issues when a harness uses subagents, without having disk KV cache enabled: out of box, memory KV cache is dropped for the main session when a subagent gets executed. This is annoying (prefill for 100K tokens of the "main" context takes many minutes), but it is fine and expected.

To work around it (without enabling disk cache yet), I've tried to use `--batched-session N` option. I was expecting it'd give ds4 a few "slots" for KV cache, so that both main and subagent can get separate caches. But this is not how it works in main branch: multiple slots are used only if requests are done in parallel, and by default the "main" slot is overwritten/reused first.

This happens because of the heuristic used in slot selection: it prefers a slot with the longest common prefix. Slot from the "main" session has at least system prompt+tools (the same prefix as subagent slots), and other slots are empty, so a subagent slot prefers to overwrite main slot, instead of using empty - they share more prefix.

I assume this is done so that in a common case (no subagents) requests are routed to the same slot, and in-memory KV cache is reused. However, KV cache can only be used if cache is an exact prefix of the request (with a few edge cases / extra details); arbitrary prefix sharing is not supported. So, "longest prefix" heuristic doesn't match what's supported in KV cache reuse, and in subagent case it actively harms.

In this PR a different slot selection (scoring) logic is implemented:

1. Slot is picked when we know that its KV cache can be reused.
2. ~~If we want to pick a slot without reuse, we prefer slots with smaller KV cache, so that the amount to prefill is smaller (less cached data is dropped).~~ Superseded: dog-fooding found this leaks slots on long-running servers (see [this comment](https://github.com/antirez/ds4/pull/765#discussion_r3781660540)); victim choice is now staleness-aware - empty slots first, then checkpoints idle over 1h (most-idle-first, regardless of size), then the shortest recently-active checkpoint ([ada805a](https://github.com/antirez/ds4/pull/765/commits/ada805ad59142fdc844f240a8601890cfe44d511)).

A minimal fix would be to modify job_slot_score function, and copy some logic from other places (e.g. instead of longest common prefix use exact prefix match). But I was thinking it'd be brittle, the logic is different for ds and glm, etc. So, instead the code is refactored, to ensure model-specific KV cache logic is exactly the same logic which is used for slot selection.

I've checked that the fix works with a few manual probes; there are also tests.

Disk KV cache masks the issue, because instead of long prefill, dropped KV cache is loaded, well, from disk. After this change less disk reads are required in case of subagent.

---

### Before / after

`--batched-session 2`, no `--kv-disk-dir`, shared ~2.7k-token system prompt: main resident, subagent arrives, main continues. Same script, same prompt, both branches.

| step | main branch | this PR |
|---|---|---|
| subagent | routed to **main's** slot (`live_tokens_before=2742`, `common=2702`, `cached=0`); main's KV evicted with zero reuse | routed to the **empty** slot (`live_tokens_before=0`); main untouched |
| main's next turn | full re-prefill, `cache_source: none`, **4.23 s** | `cache_source: memory-token`, `cached=2742`, **0.23 s** (17-token suffix) |

With `--kv-disk-dir` the misroute still happens but is masked by an `evict` save + `disk-text` reload (~59 MiB write+read per switch); this PR avoids that round-trip.

### Verification

- `make test` green (all `ds4_test` groups, incl. `--tool-call-quality`, `--think-tool-recovery`, `--long-context`, `--server`).
- Unit tests for every reuse tier + dispatch routing: `./ds4_test --server`.
- Live checks: the scenario above, plus `anthropic-tool-output` and `responses-tool-output` continuations verified end-to-end.
- Apple Silicon / Metal, `ds4flash.gguf` (IQ2XXS, ~81 GB).

<details><summary>Reproduction script</summary>

```python
# repro_slot_routing.py — sequential: main t1, main t2 (control), subagent, main t3.
import json, urllib.request
BASE = "http://127.0.0.1:8000"
SYS = "\n".join(f"Reference entry {i:03d}: alpha-{i*7%100} beta-{i*13%100} "
                f"gamma-{i*29%100} delta-{i*37%100}." for i in range(1, 151))
def chat(msgs, label, max_tokens=24):
    body = {"model":"ds4","messages":msgs,"max_tokens":max_tokens,
            "temperature":0,"stream":False,"think":False}
    req = urllib.request.Request(BASE+"/v1/chat/completions",
        data=json.dumps(body).encode(), headers={"Content-Type":"application/json"})
    with urllib.request.urlopen(req, timeout=900) as r:
        c = json.load(r)["choices"][0]["message"]
    print(f"[{label}] {c.get('content','')[:40]!r}"); return c.get("content","")
m = [{"role":"system","content":SYS},
     {"role":"user","content":"MAIN TASK: reply with exactly: MAIN-READY"}]
a1 = chat(m, "main t1")
m += [{"role":"assistant","content":a1},
      {"role":"user","content":"MAIN follow-up: entry 042? Brief."}]
a2 = chat(m, "main t2 (control)")
chat([{"role":"system","content":SYS},
      {"role":"user","content":"SUBAGENT TASK: summarize entries 001-010. Brief."}], "subagent")
m += [{"role":"assistant","content":a2},
      {"role":"user","content":"MAIN follow-up 2: entry 077? Brief."}]
chat(m, "main t3")
```

Run against `./ds4-server --batched-session 2 --trace /tmp/trace.log` on both branches and inspect the trace: on main, the subagent row shows `live_tokens_before=2742` + a `live kv cache miss`; on this PR it shows `live_tokens_before=0`, and main t3 is `memory-token`.

</details>

### Real-world scale

The table above uses a ~2.7k-token context for a quick, reproducible check; the cost scales with context. Measured on this machine (Apple Silicon / Metal, IQ2XXS): a single cold prefill of a ~120k-token prompt — i.e. exactly what main's next turn costs on `main` after a subagent evicts it — is ~5 minutes (306 s at ~402 t/s, slowing toward the tail). With this PR main's KV stays resident, so main's next turn prefills only the new turn's delta (sub-second to a few seconds). With `--kv-disk-dir` at this scale the misroute is masked by a ~2.5 GiB checkpoint save+reload per switch (extrapolated from the measured ~59 MiB at 2.7k tokens), and under disk-budget pressure ([#444](https://github.com/antirez/ds4/issues/444)) that checkpoint may be evicted before main returns — falling back to the full re-prefill.


GreatMark:
Verified on Mac Studio M3 Ultra (80-core GPU, 512 GB), Metal, MXFP4 Flash 0731 (145 GB resident), on top of `84cc882` with #611 and #704 also applied locally.

## Scenario: main session + subagent, `--batched-session 2`, no disk KV

Shared ~2.8K-token system prompt; main turn 1 → subagent request → main turn 2 (same script, streaming, `temp 0`, `reasoning_effort: none` to keep replays token-exact).

| step | main branch behaviour | this PR |
|---|---|---|
| subagent | routed onto main's slot (`live=2869 prompt=2821 common=2810 reason=token-mismatch`), main's KV evicted | routed to the empty slot, cold prefill there, main untouched |
| main turn 2 | full 2,881-token reprefill, TTFB **5.06 s** | `ctx=2869..2881:12` (12-token delta), TTFB **0.27 s** |

The `common=2810` line confirms the longest-common-prefix heuristic was exactly what misrouted the subagent, as described.

## Integration notes

- Merging on top of #611 (chat/completions tool continuation) conflicts in `slot_probe_reuse_locked` / `generate_job_inner`. Resolution that worked here: extend the ANTHROPIC_TOOL_OUTPUT probe tier condition to `req->api == API_ANTHROPIC || req->api == API_OPENAI`, and split the tier's `cache_source` into `anthropic-tool-output` / `chat-tool-output` by API. Everything else composes cleanly.
- `./ds4_test --server` passes with the resolved merge; `ds4-bench` 2K–16K sweep is within ±1% of the pre-merge baseline (prefill and decode), so no engine-path impact.

Also stacked #710 and #767 on top of the same resolved structure with all four fixes verified working together, in case that data point helps sequencing.

*Measured locally; an AI agent helped run the A/B harness and consolidate logs. Numbers come from ds4-server's own log lines.*

ds4_server.c
```diff
@@ -12404,8 +12449,20 @@ static int job_slot_score(server *s, server_slot *slot, const job *j,
...
-    int common = ds4_session_common_prefix(slot->session, &j->req.prompt);
-    return common;
+    /* Reuse-aware routing: a slot is attractive only when this request would
+     * actually reuse its live state - the same verdict the execution ladder
+     * reaches via slot_probe_reuse_locked() (dispatch holds tool_mu here).
+     * A request that can reuse nothing is sent to the cheapest victim:
+     * scoring -live penalizes evicting a resident checkpoint in proportion
+     * to its length, so an empty slot (live==0, score 0) always wins, and
+     * when every slot holds an unusable checkpoint the shortest one is
+     * evicted rather than the longest. */
```
kmike: After dog-fooding it for a few days the approach works overall, --batched-session N is usable now in practice, both with subagents and with parallel sessions.

But there is a failure mode, related to this logic (keep longest entry): 

You may have ds4 running for days, for various tasks. N is set high enough, so that KV cache slots are not overwritten due to concurrency in practice. 

Then, I work on something, and get a long conversation in KV cache. It occupies a slot. When new conversations / subagents start, they start with a smaller context. The current "main" conversation is now smaller than these old big conversations, so subagents push out your current main chat, not old, long finished conversations, which brings the problem back.

So, over time, KV cache accumulates a few long conversations, they're never evicted, and prefill trashing starts again.

kmike: Scenarios:

| Scenario | What "best" means |
|---|---|
| Single sequential session, any N | Baseline: every turn reuses its slot |
| Long-running server (days), sessions come and go, some finish | Finished conversations' checkpoints eventually reclaimed; no slow slot leak |
| Steady parallel sessions, count ≤ N | Each session keeps its slot; a new arrival evicts only when all N are resident |
| Main + subagent wave, wave width ≤ free slots | Subagents take empty/cheap slots, main untouched |
| Main + subagent wave, wave width > free slots (oversubscribed) | Someone must thrash — prefer churning short subagent slots over re-prefilling the long main |
| Multiple mains, each spawning subagents | All mains survive; subagents churn among themselves |
| Mostly-idle server, sporadic use, overnight gaps | Yesterday's *finished* convos evicted before yesterday's *continuable* main; a resumed session still reuses if its slot survived |

The solution which seems to work well for all these scenarios: add "last used" time to each slot. If the time since last use exceeds a threshold (e.g. 60min), consider the slot stale, and prefer picking it as a victim, regardless of its size.

Other approaches considered:

* simple TTL (worse, as with the proposed approach sessions over TTL can still reuse cache)
* pure LRU (worse, as it doesn't consider size, and can drop main session unnecessarily)
* some combined score for slots, which combines time and size (can work, but I don't like it, seems complex, and has unclear/hard to understand behavior which needs to be tuned)

**Update:** implemented in [ada805a](https://github.com/antirez/ds4/pull/765/commits/ada805ad59142fdc844f240a8601890cfe44d511).

wbllcbbg-ai:
Independent confirmation from a different workload: DeepSeek V4.1 Flash Q4 coding agent on M3 Ultra 512GB (Metal, `--batched-session 3`), where the "subagent" role is played by ~300-token standalone vision-helper requests (`vision_analyze` describing a screenshot). Same steal via the shared chat system prefix: the helper scores `common=27` on the main conversation's idle slot and beats an empty slot's 0, evicting a 75-90k-token live session every interleave.

The aggravating factor in our case is that multimodal sessions currently have no disk-KV fallback (#1007), so each steal forced a full ~89k-token cold re-prefill — ~6 minutes at continued-prefill speed, which the client experienced as vision calls timing out and retrying forever.

Numbers with a minimal local stopgap (trivial-prefix requests score below empty slots; real ≥64-token continuations unaffected — same direction as this PR's reuse-aware scoring):

- interleaved 25.8k-token main session + 3008×1964 image request: zero `reason=evict` entries, main continuation 2.7s (previously a 92.6s full re-prefill)
- agent workload: timeout-retry storm gone entirely

Happy to rerun your reproduction script against the V4.1/vision workload if it's useful evidence for the PR.

kmike:
It seems everything interesting has been merged in https://github.com/antirez/ds4/commit/4067320f06e63d19a81251a430658cb887ecd41e (thanks @antirez!), closing it.

PR #763, opened 2026-08-10
cuda: accelerate long-context B1 indexer and 1M HCA

wangshang23:
Fixes #762.

## What changed

- Route non-quality B1 indexer scoring to the existing WMMA scorer at
  `n_comp >= 8192`; retain the direct scorer for shorter contexts, quality
  mode, and the existing `DS4_CUDA_NO_INDEXER_WMMA` escape hatch.
- Raise the attention score capacity from 8192 to 8448 rows so the 1M HCA
  shape fits 8192 compressed rows plus the existing 256-row raw reserve.
- Extend the CUDA long-context smoke test to cover 8192 compressed HCA rows
  and direct-vs-WMMA B1 score/top-k equivalence on QAT-format values.

The HCA change does not add a new attention implementation. It keeps the
existing exact score-split path active instead of falling back to the slow
online kernel for the last 256 rows at 1M.

## DGX Spark results

Base: `84cc882`, NVIDIA GB10 (`sm_121`), DeepSeek-V4-Flash IQ2XXS Q2 model,
Q8-to-F16 weight cache disabled. A/B paths alternate in one process/session,
with 3 warmups and 30 samples each.

| Context | Base tok/s | PR tok/s | Speedup |
|---:|---:|---:|---:|
| 128K | 12.168 | 12.468 | 1.025x |
| 256K | 10.642 | 11.290 | 1.061x |
| 512K | 8.647 | 9.574 | 1.107x |
| 1M | 4.372 | 6.981 | 1.597x |

At 1M, Nsight Systems attributes the main change to:

- CSA: direct scorer `78.036 ms` -> WMMA scorer `48.565 ms` across 21 layers.
- HCA: online kernel `97.286 ms` -> exact score-split score+finalize
  `25.400 ms` across 20 layers.

Full benchmark details and the existing-work audit are in #762. Related #258
uses approximate HISA pruning and is complementary to this full-top-512 flat
scorer optimization.

## Correctness

- B1 QAT score max delta: `6.10352e-05`.
- B1 top-512 differences: `0/512`.
- Isolated 8192-row HCA regression output delta: `0`.
- Full synthetic-frontier logits are bit-identical through 512K.
- At 1M, legacy online vs existing exact score-split has max logit delta
  `0.0469046`, RMSE `0.00876944`, and identical argmax.
- Real 30,474-token story fact-recall regression passes.

## Validation

```sh
make cuda-regression CUDA_ARCH=sm_121
```

Passes on DGX Spark / GB10.

[...]

PR #621, opened 2026-07-28
Support AProjQ4 GGUFs: Q4_K dense attention projections; Metal, ROCm and CUDA performance improvements; SSD streaming support for speed-bench; quality-of-life improvements; bugfixes

GiorgioOppo:
## PR #621 — consolidated AProjQ4 results

This comment supersedes the exploratory and intermediate benchmark reports in the thread. The tables below retain only the latest meaningful matched AProjQ4/AProjQ8 comparisons and exclude known first-load artefacts from steady-state conclusions.

> [!NOTE]
> Current PR head: `6a20b131e18f6a1650706d8f76e15f60ac2bc91f`.
>
> Some of the latest complete backend A/B runs predate this exact head. Before merge, the compact validation matrix should therefore be rerun once on the final SHA.

---

## What this PR adds

This PR implements explicit support for the DeepSeek V4 Flash **AProjQ4** layout.

The following five dense attention-projection tensor families are converted from `Q8_0` to `Q4_K` in each of the 43 layers:

- `attn_q_a`
- `attn_q_b`
- `attn_kv`
- `attn_output_a`
- `attn_output_b`

This represents **215 converted tensors** in total. Routed experts, shared experts, output quantization and checkpoint identity remain unchanged.

The implementation includes:

- CPU reference execution;
- Metal resident and SSD-streaming paths;
- CUDA decode, paired projection, MMQ and grouped prefill paths;
- ROCm decode, TILE8 and direct-WMMA prefill paths;
- direct GGUF-to-GGUF requantization with dense-imatrix support;
- backend-specific exactness, parity, canary and fail-closed tests;
- benchmark harnesses and expanded QA/environment-variable documentation.

This is support for one explicit DS4 model layout, not generic support for arbitrary Q4 GGUFs.

---

## Executive summary

| Backend | Prefill, Q4 vs Q8 | Decode, Q4 vs Q8 | Current conclusion |
|---|---:|---:|---|
| **Metal** | Effectively tied in the strongest published resident A/B | Approximately **+13% to +19%**, with a **+15.5% median** over the full M5 Max sweep | Clear decode win; published prefill is parity |
| **CUDA / GB10** | **+1.98% to +2.39%** at 4K–8K | **+13.9% to +18.0%** | Q4 wins both prefill and decode |
| **ROCm / gfx1151** | Approximately **2–3% behind** at the longer measured frontiers | Approximately **+11.5% to +12.6%** | Decode win; small prefill gap remains |
[...]

PR #766, opened 2026-08-10
Optimize GB10 Q2 CUDA decode while preserving bit-exact greedy output.

ivanfioravanti:
I tried to port some of the Apple Silicon optimizations to DGX Spark kernel.

Validated with ds4 tests, CUDA regression, DSpark verify-depth, Compute Sanitizer, bit-identical output/logprobs, and the full 32-frontier GB10 sweep.

512-token decode reaches 20.69-20.74 steady t/s at 2K context; generation improves 10.3-14.0% through 65K.


GiorgioOppo:
Hi @ivanfioravanti, are you sure it's bit identical and not similar, for example, the transition from Here's → Here is?
I implemented a version of your PR and we saw this slight drift. I might have made a mistake in the implementation.
This doesn't mean there are any problems but maybe it's not exactly bit-identical

https://github.com/antirez/ds4/pull/621#issuecomment-5243136142

However we have noticed a very good increase in generation speed up to 25t/s

ivanfioravanti:
Thanks @GiorgioOppo I will double and triple check immediately. In case I'll fix it.

ivanfioravanti:
I changed PR status to draft while I check.

ivanfioravanti:
Thanks for checking, @GiorgioOppo 

I looked at the linked PR #621 discussion. The `Here's` → `Here is` drift there is associated with the separate AProjQ4 GB10 fast path and its `DS4_CUDA_NO_Q4_GB10_FAST=1` rollback. That path/flag is not present in this PR, which is testing the AProjQ8/Q2 model.

To verify this branch directly, I compared a clean `upstream/main` build at `84cc882` against the current PR head using the same model, prompt, `--temp 0 --nothink`, and full `--dump-logprobs` JSON:
   - 7,000-byte prompt + 2,048 greedy tokens: byte-identical
     `0242b920cabda9f83878d5891ce11d409b587a770adc51f5aca8ca72db5fe0d2`
   - 1,024-token long-prompt run: byte-identical
     `cd6b39a283c58f05abcb1258138275f265d5df072397616b3302a89d7b2802ff`
   - independent 256-token code and prose prompts: also byte-identical

So I'm pretty sure of my bit-identity claim for this PR’s tested path. If you have a repro on this specific branch/model, please send the exact command, prompt, GGUF hash, and environment and I’ll investigate it.

GiorgioOppo:
Thanks for the very quick check, I'll check my implementation which was probably wrong.

ivanfioravanti:
@GiorgioOppo it's part of the game, we are pushing everything like crazy here. Innovation to the maximum speed and we help each other in the journey.

JordiPosthumus:
Hi Ivan — Jordi here. I gave this a proper independent run on our second Spark, and this is lovely work. Thank you for putting it together.

Test setup:

- NVIDIA GB10 (`sm_121`)
- DeepSeek-V4-Flash-Vision-Exp `IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8`
- 2,048-token context, 512 greedy decode tokens, ABBA order
- your exact head `a2c663df`, followed by a runtime-only reconciliation onto our GB10 production source (which already carries the compatible #763/#804 long-context paths)

On the reconciled production-source comparison:

| build | prefill t/s | steady decode t/s |
|---|---:|---:|
| baseline mean | 833.81 | 19.80 |
| #766 port mean | 832.35 | 20.51 |

That is **+3.56% steady decode** for us. Prefill was **-0.18%**, which is noise at this sample size.

I also checked correctness more aggressively than generated-text equality:

- Your exact PR head with all documented rollback flags versus the optimized path: all 129,280 frontier logits were byte-identical (`110205fa13ce2f9151f36fa25356575c4b13f98d9f860e95fffbc46fac986d1a`).
- Our reconciled current-production-source build versus its baseline: all 129,280 frontier logits were byte-identical (`bbb94a3fec13ef29c2bc2a5af698f065ab6ca978669be8d12045774dd3f6376b`).

So I independently reproduce both the real GB10 decode gain and the bit-exactness claim for the tested AProjQ8/Q2 path.

The PR is currently merge-dirty, but the reconciliation was quite contained: two conflict areas, preserving the newer #804 Q8 fallback while selecting your persistent K=1024/K=4096 path where applicable, and retaining your guarded direct-Q2 dispatch. I kept your commit authorship intact. I would be very happy to share the reconciliation branch or help turn it into the cleanest current-main update if useful.

Really nice work :-D

PR #767, opened 2026-08-10
kv-cache: serialize only occupied compressor rows in checkpoints

kmike:
This PR removes about ~5MB fixed overhead per KV cache disk checkpoint (on average; ~10MB per checkpoint max, if stars are aligned - or if we start actively aligning the checkpoints later). This is done by storing only useful part of ratio-128 layers on disk, not the whole buffer always. Before this change there was ~23MB fixed overhead per checkpoint, now it's ~17MB on average.

It is possible to optimize ratio-4 layers as well, but it's kept out of scope for this PR, to simplify the code (they give much less savings).

Checked with unit tests, and also with various probes against live server.

5-10MB savings per checkpoint are nice, but not super important right now. Although they're still nice. But I think they become much more valuable if some kind of KV disk cache optimization lands (see https://github.com/antirez/ds4/issues/771). 

AI-generated description:

---

Single-lane compressor layers (compress_ratio != 0 && != 4) use the attn_state_* buffers as a ring of `ratio` rows, but checkpoints wrote the full capacity.  Only the `partial = checkpoint_len % ratio` rows accumulated in the current block are live; the rest are stale leftovers overwritten before they are ever read (they stay inside the raw SWA window, so no attention path reads the in-progress compressor state mid-block).  Serialize just those rows plus a u32 count, and clear the tail on restore.

Ratio-4 layers (coff == 2) keep the full-capacity layout byte-for-byte: their two-lane pooler reads the one-block-old lane A by design, so the stale-rows-never-read invariant is false there.

`partial` is derived from the checkpoint length and cross-checked against the already-serialized n_comp on load (n_comp*ratio + partial == len); no new struct fields or counters, and an inconsistent payload is a hard load error. All four size-accounting sites (including the separate copy in ds4_distributed) are updated to match bytes written.  Bump the three payload format constants: kvstore ABI 2->3, session payload 2->3, layer payload 1->2.

Add tests/test_kv_frontier_state: Metal round-trip at aligned (partial=0) and un-aligned (partial>0) checkpoints with size math, warm-restore tail-clear check, version rejection, partial-corruption rejection, and an opt-in fuzz sweep (DS4_TEST_FUZZ=1) over the full occupancy range with byte-stable re-save and repeated warm restore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@
 /tests/cuda_long_context_smoke
 /tests/test_layer_pack
 /tests/test_metal_session_batch
+/tests/test_kv_frontier_state
 /tests/test_mxfp4_cuda
 /tests/test_mxfp4_dot
 /tests/test_mxfp4_metal
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ endif
 .PHONY: all help clean test test-metal-session-batch test-mxfp4-cuda test-cuda-session-batch test-cuda-mixed-batch dspark-acceptance dspark-verify-depth mtp-verify-depth cpu cuda cuda-spark cuda-generic cuda-regression strix-halo rocm
 
 ifeq ($(UNAME_S),Darwin)
-.PHONY: metal-decode-schedule-bench metal-prefill-variant-bench check-mxfp4-half-lut
+.PHONY: metal-decode-schedule-bench metal-prefill-variant-bench check-mxfp4-half-lut test-kv-frontier-state
 
 all: ds4 ds4-server ds4-bench ds4-eval ds4-agent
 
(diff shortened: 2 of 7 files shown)


GreatMark:
Verified on Mac Studio M3 Ultra (80-core GPU, 512 GB), Metal, MXFP4 Flash 0731, merged cleanly on `84cc882` (stacked with #765/#611/#704/#710 locally).

- `tests/test_kv_frontier_state` with the production quant: **PASS** — aligned payload 17,033,692 bytes vs unaligned 26,715,580 (delta 9.68 MB), version-rejection and partial-corruption rejection both exercised.
- Live `ds4-server` with `--kv-disk-dir`, real checkpoint sizes for the same requests before/after:

| checkpoint | main | this PR |
|---|---:|---:|
| 8,192 tokens (aligned, partial=0) | 130.44 MiB | **120.44 MiB** (−10.0) |
| 9,832 tokens (partial=104) | 151.94 MiB | 150.07 MiB (−1.9) |

Matches the "~17 MB average, ~10 MB best case when aligned" framing in the description. ABI bump behaves as intended: pre-existing v2 checkpoints are rejected on load rather than misread, then re-written at the new size.

`./ds4_test --server` passes; `ds4-bench` 2K–16K within ±1% of baseline.

Agreed this compounds well with #771 — at a 10k-token continued-interval a long agent session writes checkpoints constantly, and the fixed overhead is exactly what that proposal amortizes.

*Measured locally; an AI agent helped run the harness. Sizes are from ds4-server's own `kv cache stored` log lines.*

PR #782, opened 2026-08-11
metal: lightning-indexer-organized DS4 indexer scorer (llt)

daemon2k3:
Ports llama.cpp's Metal kernel_lightning_indexer organization to ds4's ratio-4 (CSA) indexer score stage on pre-M5/non-NAX paths, replacing both the prefill tiled kernels and the decode direct score kernel.

Structure: each 256-thread threadgroup owns 64 compressed rows, staged once and transposed into per-simdgroup REGISTER matrices; the 64-head loop then reuses register-resident K instead of re-reading threadgroup K per head. Heads are processed in 8-head tiles (single Q stage + pre-scaled head weights, 2 barriers per tile vs 3 per head); prefill iterates 8 tokens per kernel pass against resident K; decode runs the same kernel with n_tokens=1. Per-cell math preserved (f32->half staged rows, ascending depth-8 simdgroup matrix steps, relu then w*scale per head ascending, ds4 causal -inf epilogue for prefill / all-rows for decode).

Measured on Apple M3 Ultra 512 GB, DeepSeek V4 Flash resident, ctx frontier probes over HTTP + ds4-bench + variant benches:

* decode bit-exact vs kernel_dsv4_indexer_score_one_direct: metal_decode_schedule_bench --candidate-env DS4_METAL_DISABLE_INDEXER_LLT --include-selection: exact_rows=529, exact_floats=68389120, exact_selected_ids=528
* decode @117.6k ctx: 39.3 -> 42.2 / 41.5 t/s (+6-7%, two replicates)
* short-ctx decode 48.0 t/s, prefill cold @117.6k 405 t/s: parity
* prefill score stage @29.4k rows: parity vs scores_tiled (dual-head / wide-tile / NSG4 alternate organizations were also measured and revert for net-negative; knobs preserved below)

Env knobs: DS4_METAL_DISABLE_INDEXER_LLT (rollback, variant-bench control), DS4_METAL_INDEXER_LLT_NBPTG=16|32 (token batch per pass, measured flat), DS4_METAL_INDEXER_LLT_NSG4 (4-simdgroup / 2-TG-per-core variant, measured prefill-negative/decode-slightly-positive, off).

### DIFF
--- a/ds4_metal.m
+++ b/ds4_metal.m
@@ -17402,8 +17402,16 @@ int ds4_gpu_indexer_score_one_tensor(
         }
 
         if (n_head == 64 && head_dim == 128) {
-            id<MTLComputePipelineState> direct_pipeline =
-                ds4_gpu_hot_pipeline(g_dsv4_indexer_score_one_direct_pipeline,
+            /* Same lightning-indexer organization for decode: one kernel,
+             * causal masking off (n_tokens==1).
+             * DS4_METAL_DISABLE_INDEXER_LLT rolls back (A/B control). */
+            const bool score_llt =
+                getenv("DS4_METAL_DISABLE_INDEXER_LLT") == NULL;
+            const bool score_nsg4 = getenv("DS4_METAL_INDEXER_LLT_NSG4") != NULL;
+            id<MTLComputePipelineState> direct_pipeline = score_llt
+                ? ds4_gpu_get_pipeline(score_nsg4 ? "kernel_dsv4_indexer_scores_llt_nsg4"
+                                                  : "kernel_dsv4_indexer_scores_llt")
+                : ds4_gpu_hot_pipeline(g_dsv4_indexer_score_one_direct_pipeline,
                                         "kernel_dsv4_indexer_score_one_direct");
             if (!direct_pipeline) return 0;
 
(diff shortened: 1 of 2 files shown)


adriangalilea:
Measured this branch cherry-picked onto 84cc882 on Apple M3 Ultra (60-core GPU) 512 GB, DeepSeek V4 Flash 0731 native-MXFP4 GGUF (156 GB).

Decode: exact, as claimed — `./speed-bench/metal_decode_schedule_bench -m ds4flash.gguf --candidate-env DS4_METAL_DISABLE_INDEXER_LLT --include-selection --tokens 256` → `exact_rows=273 exact_floats=35293440 exact_selected_ids=272`, control 44.0 vs candidate 44.0 t/s.

Prefill: **not exact** — `./speed-bench/metal_prefill_variant_bench -m ds4flash.gguf --prompt-file speed-bench/promessi_sposi.txt --candidate-env DS4_METAL_DISABLE_INDEXER_LLT --prefix-tokens 8192 --repeats 2` aborts at run 2 with `raw logit mismatch: differing=129280/129280` (full vocab) between the LLT scorer and the tiled kernel. The PR's exactness evidence is decode-only; on this machine the prefill path changes every logit.

Harness calibration first, since #793 questions it on this hardware: control-vs-control (`--candidate-env DS4_LAB_NOOP`) is bit-exact here, 8/8 runs, 1,034,240 floats, delta −0.01% — so the gate is sound on this box and the mismatch is real.

Happy to re-run a revised branch — the A/B takes ~4 minutes here.

PR #785, opened 2026-08-12
fix: detect client disconnect on plain FIN; log cancelled jobs

nazerim:
Fixes a real bug in the client-disconnect cancel added in e9ded97: on Darwin,
`poll()` does not report a plain TCP FIN, so a client that closed with a clean
FIN mid-generation was never seen by `wait_for_job_or_disconnect`. The job kept
decoding into a dead socket until `max_tokens` (on an M5 Max, ~12s of GPU work
per disconnect). Only an RST — e.g. `close()` with unread receive data — tripped
cancellation, via the send path, and that path returned silently.

The socketpair unit test (`test_waiting_job_cancels_on_client_close`) misses
this because it writes pipelined data before closing, which makes `poll()`
report POLLIN; recv then hits EOF. A clean FIN with no trailing data never
produces that event on Darwin.

## Fix (2 commits)

1. **`client_socket_disconnected`**: after an eventless poll, probe with
   nonblocking `recv(MSG_PEEK)`, which reports EOF exactly on every platform
   and consumes nothing, so it stays race-free with the worker's concurrent
   SSE writes on the same fd.
2. **`generate_job_inner`**: the seven `job_cancelled` early returns skip the
   response path where the existing "client disconnected" log lives, so
   cancelled work was invisible; surface the cancellation at each exit point.

## Verification (per CONTRIBUTING.md)

Commands run on the PR branch (pure upstream + these two commits, nothing else):

```sh
make clean && make          # no warnings
make test                   # all suites pass
make cpu                    # CPU portability target builds clean
```

Machine/backend: Apple M5 Max, Metal backend.
Model: `ds4flash.gguf` (0731 mixed-quant build, DS4F chat v2).
`make test` result: long-context, tool-call-quality, think-tool-recovery,
logprob-vectors, metal-ssd-streaming-cache-pressure, local-golden-vectors,
metal-short-prefill, metal-kernels, metal-tensor-equivalence,
streaming-decode-prefill-correctness, server — all OK; test_layer_pack 97/97,
test_engine_mgpu_placement 101/101, test_gpu_args all passed, test_gpu_args_cli
PASS=44 FAIL=0. mtp-verify-depth and dspark-verify-depth self-skip (require
MTP/DSpark support GGUFs, not applicable to this change).

New live regression test `tests/live_disconnect_test.py` (opt-in, skips without
a model) aborts a real streaming request mid-generation with a clean FIN:

```
[live] PASS: server logged 'client disconnected' promptly - lag=0.51s
[live] PASS: disconnect detected within 15s of close - lag=0.51s
[live] PASS: cancelled job's gen count far below max_tokens(512) - gen=0
[live] PASS: server serves normal request after disconnect
[live] PASS: tools request returns 200 with tool_calls or content
[live] failures: 0
```

Notable failures: none. No inference/backend code touched — cancellation path
only, no speed regression possible.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -10883,6 +10883,17 @@ static void request_ctx_span(char *buf, size_t len, int cached, int prompt) {
     snprintf(buf, len, "%d..%d:%d", cached, prompt, suffix);
 }
 
+/* A job cancelled mid-flight (client disconnect or stream write failure) skips
+ * the response path where the final "client disconnected" log lives; surface
+ * it here so operators can see cancelled work. */
+static void log_job_cancelled(const job *j, const char *ctx_span) {
+    if (!j) return;
+    server_log(DS4_LOG_DEFAULT,
+               "ds4-server: %s ctx=%s client disconnected",
+               j->req.kind == REQ_CHAT ? "chat" : "completion",
+               ctx_span);
+}
+
 static void log_flags(char *buf, size_t len, bool responses_protocol,
                       bool tools, bool thinking,
                       bool dsml_start, bool dsml_end) {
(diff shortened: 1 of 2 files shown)


nazerim:
Cross-link: this PR is the proposed fix for #786 (client disconnect mid-generation not detected on clean TCP FIN — Darwin poll bug).

Issue #678, opened 2026-08-04
Streaming `/v1/messages`: a second literal think-close tag is emitted as answer text (`ANTH_STREAM_TEXT` has no close-tag handling)


kandaman:
(Formatting note: the think open/close markers are written inside code spans throughout,
so chat clients that treat them as control tags render this cleanly.)

### Summary

On the streaming Anthropic endpoint, the model frequently writes a **second round of
reasoning into the answer channel** and closes it with a literal `` `</think>` ``. That tag,
and all the draft text before it, reach the user as ordinary `text_delta` content. The
reader sees ~90% draft, a stray close tag, and then the actual answer.

The reasoning channel itself is correct — `thinking` blocks are populated normally. This is
not a truncation, not a parse failure, and the turn ends with a normal `stop_reason`; the
tool call, when there is one, is delivered fine. Nothing in the response marks it as
degraded, which is why it went unnoticed on our side for six weeks.

### Measured (production, 2026-06-25 … 2026-08-04)

Two independent datasets from a multi-node fleet behind one proxy:

| | |
|---|---|
| Rate | **~2.2%** of turns that produced answer text (harness-internal calls excluded) |
| Users affected | multiple, across independent sessions |
| Backends | **every node in the fleet** — two different quantisations, both Metal and CUDA |
| Context correlation | 0.28% at 20–50k prompt tokens → **8.43% above 150k** |
| Session distribution | bimodal — most sessions 0%, but once it starts it persists (one session: 20 of 31 turns) |

Full scan of every raw SSE dump we retain (30 days), same fleet:

| | |
|---|---|
| Close tag in a `text` block | **95** streams |
| Close tag in a `thinking` block | **0** |
| **Opening** think tag anywhere in answer text | **0** |
| Tag split across two `text_delta` events | **0** (it is a single vocabulary token) |
| Streams with ≥2 tags | 6 |
| Tag character offset | 0 … 86,601 |
| Median share of the answer that precedes the tag | **89.2%** |

Zero opening tags is the diagnostic part: the model is not writing a `` `<think>` `` block of
its own. It is *closing* one that it believes is open.

### Why it happens

1. `ds4.c` — the think markers are special vocabulary tokens, and
   `ds4_chat_append_assistant_prefix()` prepends the opening one to the assistant turn. So
   the model never emits an opening tag; it starts already inside a thinking block.
[...]

PR #787, opened 2026-08-12
server: stream the answer past a stray </think> instead of holding it (#783)

### LINKED ISSUE #783: Non-thinking streaming output broken after commit fe2d3b0
Commit fe2d3b0 prevents streaming responses from working correctly for non-thinking content in frontends (e.g. Open WebUI) where native tool calling is the default. Found the commit using git bisect.

The end result is that instead of being able to see and start reading the response right away, nothing at all appears after reasoning is done until the final token completes. For longer responses this can introduce extremely long gaps where the output is blank and feels highly delayed.

For common use cases like web searches, being able to see the answer in the output as quickly as possible is important, even if the model is still emitting additional context.

Temporary workaround for anyone else seeing this, which rewinds to the commit just before this change was introduced:

- `git checkout 769411247560dcc599369bba5fe2d6df8ac3fd0b`
- `make clean`
- `make`
- relaunch

Flor1an-B:
Fixes #783.

## Problem

`fe2d3b0` (#678) guards against a tool-enabled model writing a *second* reasoning pass into the answer channel and closing it with a literal `</think>`. On the streaming path the guard holds **all** answer text back until a second `</think>`, a tool call, or the final chunk arrives. For the common single-pass case — thinking + tools, one reasoning block, then a plain answer with no tool call — none of those arrive until the end, so the whole answer is buffered and flushed at once. Streaming is dead (bisected to `fe2d3b0` by @links486; reported against Open WebUI, which defaults to native tool calling).

## Why the old guard can't stream

The routing of post-`</think>` text depends on whether a *second* `</think>` ever appears — a whole-stream property with no early deterministic signal. "Hold until proven" is correct but fundamentally cannot stream. This PR trades that: stream the answer as content in real time and simply **drop any stray `</think>`** so it never reaches the client. The #678 diagnostics make this safe to do inline — the tag is a single vocabulary token (never split across deltas, `0` split cases measured), the model never re-opens `<think>` (`0` opening tags measured), and a partial trailing `<` is already held by `text_stream_safe_limit()`.

## Net effect vs the two prior states

- **vs current `main` (blocking):** streaming restored for every thinking+tools turn.
- **vs pre-#678 (leaking):** the stray tag is now *suppressed* rather than forwarded verbatim — strictly better. The residual is that in the now-rarer second-pass case (`fe2d3b0` also fixed the tool-schema prompt that provoked it) the draft streams as content instead of landing in the reasoning channel. Given the impossibility of early detection, this is the streaming-side tradeoff for real-time output.

Applied identically to the OpenAI (`/v1/chat/completions`) and Anthropic (`/v1/messages`) streams.

## Testing

- The two `fe2d3b0` guard tests are updated to the new semantics and renamed (`*_suppresses_second_think_tag`).
- Two regressions added (`*_streams_answer_incrementally`) assert content is emitted mid-stream, before the final chunk.
- **Negative control:** reverting the production hunk fails both new assertions (`st.sent_content` and the content-routing check), so the tests genuinely pin the regression.
- `./ds4_test --server` green; warning-free `ds4-server` build on Apple M5 Max (Metal).

**End-to-end**, M5 Max / Metal / DeepSeek-V4-Flash resident, streaming `/v1/chat/completions` with a tool defined and thinking on, on a prompt the model answers directly (the case #783 regressed):

```
reasoning deltas : n=23  first=1.487s last=1.981s
content   deltas : n=43  first=2.003s last=2.957s
[DONE] at        : 2.957s
content started  : 0.954s BEFORE [DONE]   (was: single burst at [DONE])
content spread   : 0.953s across 43 deltas
```

Trace confirms the guarded path (tool-schema prompt injected, `<think>…</think>` then a plain answer, no tool call); output contained no stray `</think>`.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -6533,24 +6533,27 @@ static bool openai_sse_stream_update(int fd, server *s, const request *r, const 
     }
 
     if (st->mode == OPENAI_STREAM_TEXT) {
-        if (st->guard_second_reasoning) {
+        /* A tool-enabled model sometimes writes a second reasoning pass into the
+         * answer channel and closes it with a literal </think> (issue #678).
+         * Drop any such stray tag so it never reaches the client, but keep
+         * streaming the surrounding text as content in real time: the previous
+         * guard held the whole answer back until the final chunk, which broke
+         * streaming for the common single-pass case (issue #783). The tag is a
+         * single vocabulary token, so it never splits across updates; a partial
+         * trailing '<' is already held by text_stream_safe_limit() below. */
+        while (st->guard_second_reasoning) {
             const char *close = strstr(raw + st->emit_pos, "</think>");
             const char *tool = r->has_tools ?
                 find_any_tool_start(raw + st->emit_pos) : NULL;
-            if (close && (!tool || close < tool)) {
-                const size_t limit = (size_t)(close - raw);
-                if (limit > st->emit_pos &&
-                    !sse_chat_delta_n(fd, r, id, "reasoning_content",
+            if (!close || (tool && tool < close)) break;
+            const size_t limit = (size_t)(close - raw);
+            if (limit > st->emit_pos) {
+                if (!sse_chat_delta_n(fd, r, id, "content",
                                       raw + st->emit_pos,
                                       limit - st->emit_pos)) return false;
-                if (limit > st->emit_pos) st->sent_reasoning = true;
-                st->emit_pos = limit + strlen("</think>");
-                st->guard_second_reasoning = false;
-            } else if (!tool && !final) {
-                return true;
-            } else {
-                st->guard_second_reasoning = false;
+                st->sent_content = true;
             }
+            st->emit_pos = limit + strlen("</think>");
         }
 
         const char *tool = r->has_tools ? find_any_tool_start(raw + st->emit_pos) : NULL;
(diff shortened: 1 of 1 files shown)


apetersson:
This matches the regression I reproduced with tool-enabled DeepSeek streaming; restoring mid-generation content delivery across both OpenAI and Anthropic paths is the right fix.

links486:
Really hoping that this can be rebased and merged.

Flor1an-B:
@links486 sorry for leaving this stale — following up properly instead of just rebasing.

This PR and #1099 (also mine, more recent, already has an LGTM) end up addressing the same underlying symptom from two genuinely different, non-reconcilable angles, and I should have flagged the overlap sooner:

- **This PR**: stream immediately, always. Any stray `</think>` is detected as a single non-splitting token and dropped from the client-visible text, but by the time it's detected the preceding bytes have already been sent as `content` — so a genuine second-reasoning draft (the #678 case) is no longer reroutable to `reasoning_content`, short or long. Zero added latency, but #678 classification is fully given up.
- **#1099** (building on #1084's `SECOND_REASONING_GUARD_BYTES` approach, already merged in part via `d31089d`/`39bc628`): hold up to a bounded 32-byte window before committing text as `content`, so a *short* stray draft (the actual #678 case, measured at 13 bytes) still gets correctly rerouted to `reasoning_content`. A draft longer than 32 bytes leaks as content too, same as this PR, but the common historical case is preserved. Trade-off: every tool-enabled answer waits up to 32 bytes before streaming starts, instead of zero.

Given `main` already carries a partial version of the bounded-window direction (the Qwen-only exclusion from `d31089d`), and #1099 is the more complete, already-reviewed continuation of that same direction, I think consolidating behind #1099 rather than maintaining two competing designs from the same author is the right call — closing this one in its favor rather than rebasing it further. If the zero-latency tradeoff turns out to matter in practice (worth reopening if so), the core observation here (`</think>` never splits, `0` opening tags measured) is still correct and would be the way to justify revisiting it.

Streaming for thinking+tools should work correctly once #1099 lands — apologies again for the confusion in the meantime.

Flor1an-B:
Closing in favor of #1099 — see comment above for the reasoning.

PR #789, opened 2026-08-12
Port visible-KV checkpoint fixes for tool turns (fixes token-mismatch reprocesses on agentic workloads)

fradav:
## Problem

On agentic workloads (tool-calling, thinking, steering interrupts, "send now" with a new message, or even a mid-thinking stream death), `ds4-server` frequently logs `live kv cache miss ... reason=token-mismatch` and then re-prefills the **entire** context — routinely 140k+ tokens. This happens on every failed tool, every steering interrupt that appends a new user message, and sometimes in the middle of a thinking run with no interruption at all.

Root cause: chat/completions and Anthropic tool-call turns had **no live binding** to the next request. The sampled tail (hidden reasoning, exact DSML spelling, BPE drift) never token-matches the client's replay, and the byte-exact memory-text path rarely matches either — so every agent turn paid an evict-store (~650 MiB) plus disk-snapshot restore and tail re-prefill.

The Responses API already solves this with a predicted visible-transcript key. This PR reuses that mechanism for chat/Anthropic.

## Changes (all in `ds4_server.c`, server-only)

1. **Visible checkpoint for chat/Anthropic tool-call turns** — after `finish=tool_calls`, remember `prompt_text` + the same canonical suffix `canonicalize_tool_checkpoint()` builds (byte-equality with the next rendered prompt is already covered by `test_tool_checkpoint_suffix_is_future_prompt_canonical`). The next request continues from live KV, tokenizing only the new tool results. Hits are labeled `tool-visible` in logs.

2. **Visible checkpoint for tool-context turns that don't call a tool** — when the conversation has tools (or tool-using history), the chat template preserves reasoning verbatim in future renders, so a plain turn that finished without calling a tool still renders byte-for-byte into the next request. Previously such turns had no visible checkpoint at all and fell through to raw token matching. New gate `should_remember_tool_context_checkpoint()` + unit test.

3. **Preserve the live frontier on streaming errors** — the tool calls were generated and the KV rows written; only HTTP delivery failed. The `!response_ok` path keeps `job_mark_cancelled()` and the error accounting but deliberately drops `request_live_state_clear()` so the freshly recorded frontier survives a retry ("send now").

4. **Store KV waypoints through tool calls too** — waypoint snapshots are stored on disk in full (compressed KV, indexer, compressor frontiers included), so a snapshot taken mid-tool-call is as safe as any other. Previously waypoints were suppressed for the whole tool-call duration; long agentic turns left no recent restart point, forcing a near-total reprocess when the client stream died mid-tool-call.

## Validation

- `make ds4_test` clean
- `./ds4_test --server` → `server: OK`
- New unit test `test_tool_context_checkpoint_remember_gate` passes
- Metal, SSD streaming, and distributed paths untouched (server-only change)

These are backported from the fork's `b1e3287`, `47ee3ce`, and `ac98ff4`, adapted to the slot-threaded upstream structure.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -12235,7 +12235,43 @@ decode_again:
     server_generation_leave(s);
 
     if (job_cancelled(j)) {
-        request_live_state_clear(s, slot);
+        /* Even on cancellation, save a checkpoint so the next request can
+         * reuse the KV cache instead of reprocessing from scratch.  Parse the
+         * generated text and remember the visible prefix. */
+        if (j->req.kind == REQ_CHAT && text.len > 0) {
+            tool_calls parsed_calls = {0};
+            char *parsed_content = NULL;
+            char *parsed_reasoning = NULL;
+            bool recovered = false;
+            const char *cancel_finish = finish ? finish : "stop";
+            parse_generated_message_for_response_for_syntax(
+                j->req.model_syntax,
+                text.ptr ? text.ptr : "",
+                j->req.has_tools,
+                saw_tool_start,
+                ds4_think_mode_enabled(j->req.think_mode),
+                &cancel_finish,
+                NULL, 0,
+                &parsed_content,
+                &parsed_reasoning,
+                &parsed_calls,
+                &recovered);
+            /* Save checkpoint with whatever was generated, even if incomplete. */
+            if (parsed_calls.len) {
+                remember_tool_visible_checkpoint(s, slot, j, ctx_span, trace_id,
+                                                 parsed_content ? parsed_content : "",
+                                                 parsed_reasoning, &parsed_calls, true);
+            } else if (j->req.prompt_preserves_reasoning || j->req.has_tools) {
+                remember_tool_visible_checkpoint(s, slot, j, ctx_span, trace_id,
+                                                 parsed_content ? parsed_content : "",
+                                                 parsed_reasoning, NULL, false);
+            }
+            tool_calls_free(&parsed_calls);
+            free(parsed_content);
+            free(parsed_reasoning);
+        } else {
+            request_live_state_clear(s, slot);
+        }
         trace_event(s, trace_id, "cancelled during generation after %d tokens", completion);
         anthropic_stream_free(&anthropic_live);
         openai_stream_free(&openai_live);
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -2837,13 +2837,20 @@ static bool responses_validate_tool_outputs(server *s, const chat_msgs *msgs,
                 ok = false;
                 break;
             }
-            if (!prior) {
+            if (!prior || live_known) {
+                /* Require live state when: (a) no prior in history (stateless replay
+                 * impossible), or (b) live state exists (bind to the slot with the
+                 * full KV context including hidden reasoning, even if history has
+                 * the call).  This prevents slot misassignment when the client
+                 * replays full history but the live frontier has richer state. */
                 if (requires_live_tool_state) *requires_live_tool_state = true;
-                continue;
             }
-            if (needs_reasoning &&
-                (!prior->reasoning || !prior->reasoning[0]))
-            {
+            if (!prior && needs_reasoning) {
+                /* When there's no prior in history, also check if reasoning is needed.
+                 * When live_known is true but prior exists, reasoning check below applies. */
+                if (requires_live_reasoning) *requires_live_reasoning = true;
+            } else if (prior && needs_reasoning &&
+                       (!prior->reasoning || !prior->reasoning[0])) {
                 if (requires_live_reasoning) *requires_live_reasoning = true;
             }
         }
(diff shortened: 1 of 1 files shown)


Flor1an-B:
Nice — this is broader than #709 (adds Anthropic, waypoints through tool-call turns, and disconnect/error preservation). Flagging where the two meet, since they collide on merge and, more usefully, cover opposite halves of the same axis.

Both remember `prompt_text + build_tool_checkpoint_suffix(...)` as a visible key for chat/Anthropic tool turns. The one difference that matters is the `reasoning` argument, because `build_tool_checkpoint_suffix()` renders `reasoning + "</think>"`:

- This PR passes `parsed_reasoning`, so the key carries the reasoning verbatim — it matches a client that **replays** reasoning. Consistent with `should_remember_tool_context_checkpoint()` gating on `prompt_preserves_reasoning`.
- #709 passes `reasoning = NULL`, so the key is reasoning-free (`<think></think>`) — it matches a client that **strips** reasoning, which is exactly #691 (`chat_history_uses_tool_context()` keeps the `<think>` tag but renders empty reasoning when the client echoes none).

Opposite client modes — and on current `main` the `has_tools` + strip-reasoning case is handled by neither: `should_remember_thinking_checkpoint()` still bails on `r->has_tools`, and `should_remember_tool_context_checkpoint()` here requires `prompt_preserves_reasoning`. So this PR's tool-context path and #709 are complementary, not redundant: this one owns the preserve case (plus Anthropic / waypoints / disconnect), #709 owns the strip case (#691).

The clean unification is to pick the reasoning by the flag rather than hardcoding it — roughly `build_tool_checkpoint_suffix(..., r->prompt_preserves_reasoning ? parsed_reasoning : NULL, ...)` at both the tool-call and tool-context sites, and dropping the `prompt_preserves_reasoning` precondition from the tool-context gate so the stripped case is remembered too. That covers both modes in one path and lets #709 fold in as the `NULL` branch. Happy to do that reconciliation on the #709 side once directions settle.

(Heads-up: #714 also changes `thinking_live_remember`'s signature, so this is a three-way touch on that function plus the post-turn checkpoint block in `generate_job_inner`.)

fzeng656:
I tested PR #789 with OpenClaw on macOS/Metal and found a few additional KV reuse failure modes around tool turns and auxiliary requests. I put the follow-up changes here:

https://github.com/fzeng656/ds4/commit/86187c8b3094c295752d2171f2e1a365aa8cd30d

Repro/setup:
- Apple M5 Max, 128 GB unified memory, Metal backend
- DeepSeek V4 Flash 0731
- OpenClaw tool workflow
- --batched-session 2
- long context around 80K–90K tokens

What I observed:
1. OpenClaw inserts short auxiliary/status/runtime requests between main-agent turns. With the existing slot scoring, a tiny accidental common prefix, often BOS only, could select the long resident slot and evict valuable KV state.
2. Main-agent replay can retain a very large exact common prefix while rewriting only the tail. I saw cases such as live=79183 prompt=79378 common=75643, but the server still fell back to a full prefill.
3. Tool checkpoint canonicalization may return REBUILD_NEEDED when the rewritten tail falls outside the safely rewritable raw-KV window.
4. Under disk-cache pressure, storing the newest live checkpoint could evict the older continued checkpoint that is actually the best rebuild anchor.
5. OpenClaw also injects a transient trailing runtime-context user message that is removed on replay, so the live transcript and the next canonical replay can diverge exactly at that role boundary.

The follow-up patch does four main things:
- preserves resident-slot affinity by preferring the shortest slot when the common prefix is trivial, while keeping longest-common-prefix selection for meaningful affinity;
- recognizes/removes the transient OpenClaw runtime-context tail when building the future tool checkpoint;
- when a large live common prefix cannot be rewritten in place, preserves existing disk checkpoints and reloads the best byte-prefix checkpoint before replaying the tail;
- keeps the immediately previous continued checkpoint as a rebuild anchor under eviction pressure.

Observed result after the changes:
- before: 80K–90K token mismatch could trigger a full prefill taking roughly 200–250 s;
- after: requests around 90K consistently reload an 81,920-token disk checkpoint and replay only ~5K–9K tokens, typically around 17–30 s;
- tool checkpoint rebuilds similarly drop to a few thousand replay tokens instead of rebuilding 80K+ from zero.

Example after the fix:
kv cache hit text tokens=81920
chat ctx=81920..90273:8353 TOOLS prompt start
...
tool checkpoint canonicalization needs 5472 tokens rebuild ... cached=81920 source=disk
...
tool checkpoint rebuild done ... source=disk cached=81920 replay=5472 ... 17.036s

Validation:
- make clean && make -j
- full make test passes on my machine, including long-context, tool-call-quality, server tests, Metal kernel tests, layer-pack, mgpu placement, and GPU CLI tests.

I based this branch on #789 head e6c7c43. If useful, feel free to cherry-pick or adapt only the pieces that fit the intended design.

Issue #793, opened 2026-08-13
speed-bench: variant harnesses show slot-order bias — control-vs-control logits mismatch (bit-exact A/B impossible on this config)


daemon2k3:
# Issue: speed-bench variant harnesses show slot-order bias — control-vs-control logits mismatch (bit-exact A/B impossible on this config)

**Component**: `speed-bench/metal_prefill_variant_bench.c` (and `metal_decode_schedule_bench.c`)
**Hardware/config**: Apple M3 Ultra, 512 GB, macOS 26 (Darwin 25.6), DS4 @ 84cc882+, DeepSeek V4 Flash (284B, GGUF q4 imatrix 0731, `--ctx 1048576`, `--kv-cache-boundary-align-tokens 4096`, `--dspark` + MTP support sidecar)
**Reporter**: daemon2k3 (work done with heavy AI assistance, disclosed)

## Summary

On the above config, `metal_prefill_variant_bench` — with the documented ABBA machinery against deterministic run-order — reports *full-vocabulary raw logit mismatches* between counterbalanced slots for **control-vs-control reproductions**, i.e. even when the compared engine paths are identical binaries and identical env knobs. Bit-exact A/B verification of kernel variants is therefore impossible on this config; the only route to classifying numeric changes is the engine-debug dump-diff protocol (below). The observed pattern:

```
metal-prefill-variant-bench: model=ds4flash.gguf prompt=ds4.c prefix=8192
    warmup=32 ctx=8193 repeats=2 candidate_env=DS4_METAL_DISABLE_INDEXER_LLT
run=1 repeat=1 pattern=ABBA slot=1 variant=control tokens=8192
    seconds=13.926444 tokens_per_second=588.2334 exact=yes
metal-prefill-variant-bench: raw logit mismatch at run 2: differing=129280/129280
metal-prefill-variant-bench: first mismatch id=0
    reference=-0x1.54284p+4 (0xc1aa1420) observed=-0x1.58b41cp+4 (0xc1ac5a0e)
```

Separately (with `metal_decode_schedule_bench --include-selection`):

```
metal-decode-schedule-bench: model=ds4flash.gguf prompt=ds4.c prefix=2048
    ctx=4096 warmup=16 measured=512 control=2/32 candidate=2/32
    candidate_env=DS4_METAL_DISABLE_INDEXER_LLT_F3 include_selection=yes
metal-decode-schedule-bench: raw logit mismatch at frontier 0:
[...]

PR #794, opened 2026-08-13
metal: code-layer performance wins (rb16 prefill dispatch, vectorized K/Q staging in the LLT scorer, double-buffered head tiles, simd_max reducers, host fixes, zero-copy payload spans)Code layer wins

daemon2k3:
# PR: Metal code-layer performance wins (rb16 prefill dispatch, vectorized K/Q staging in the LLT scorer, double-buffered head tiles, simd_max reducers, host fixes, zero-copy payload spans)

## Summary

A set of code-layer (non-algorithmic) optimizations over the DS4 Metal backend, measured end-to-end on an Apple **M3 Ultra 512GB + DeepSeek V4 Flash (q4, 153GiB, resident)** against the current `main` + LLT scorer. Every change preserves existing semantics; the two sampling-path items are strictly opt-in via env flags and change behavior only when explicitly requested. **Deterministic per build per lineage; the deep-context dump-diff verification protocol below is included.**

## What is in this PR

| area | change | effect (measured, same-quant isolated A/B pair, alternating trials) |
|---|---|---|
| prefill attention dispatch | dispatch `heads8_rb16` on prefill (was `heads8`) — 16 staged rows per stage kernel with identical per-row online-softmax sequence | part of **+5.2%** cold prefill @117k ctx; warm TTFT −8.0% |
| LLT indexer K/Q staging | float4→half4 vectorized staging (mirrors the NAX kernel) | part of decode +3.9% @117k |
| LLT head-loop double buffer | stage tile t+1 while `simdgroup_multiply_accumulate`-ing tile t | the deep-context decode/TTFT gains above |
| 5 × fp8/hadamard finalize kernels | replace 7-barrier tree max-reduce with `simd_max` + one cross-group exchange (`max` is IEEE-exact — bit-identical) | decode −1–2% |
| Metal decode dispatch env reads | process-lifetime env cache (was ~200–400 linear `environ` scans per token across 43 layers) | host-path hygiene |
| speculative verify scratch | session-owned 2×vocab `spec_row_logits` (was 1.3–2.6 MB `xmalloc`+free per 1–2 committed tokens with DSpark/MTP) | page-fault churn removal |
| tokenizer signature matching | precomputed literal lengths (`sizeof(lit)-1`) instead of `strlen` per scanned byte | 10–50 ms on 100k-token prompt ingest |
| KV disk cache store lookup | mtime gate **off by default** (second-resolution `st_mtime` can skip fresh entries in the same second — deterministic lineage integrity beats ms-scale refresh savings) | correctness-first default |
| checkpoint payload spans | `fwrite`/`fread` directly into shared (unified-memory) Metal tensor backing stores + GPU fence (fwd both ways), removing the 8MB staging loop | warm restore TTFT 30k ctx: 11.4s vs 12.7–19.2s |
| sampling top-p<1 expf loop | **opt-in only** `DS4_SAMPLE_ACCELERATE_EXP=1`: Accelerate `vvexpf` (measured 0.35 ns/op vs scalar `expf` 1.29 ns/op ≈ 2.6–4×) — **not bit-identical to scalar expf, changes the sampling stream by construction — off by default** | micro-benchmarked only |
| Makefile | Darwin `-framework Accelerate` link (for the vvexpf opt-in); `DS4_LTO=thin` opt-in (Apple Silicon release builds) | build hygiene |

## End-to-end measured outcomes

Isolated identical-instance pair on the same q4 model, `--dspark` + `--mtp` production-config, 5+ alternating trials each metric:

```
decode @117k ctx:   32.47 → 35.76 t/s       +3.9%   (medians; ranges non-overlapping)
warm TTFT @117k:    24.62 → 22.65 s         −8.0%   (medians)
cold 117k prefill:  244.5 → 232.5 s         −5.2%
short ctx (8k):     prefill 514.3 → 515.9 t/s, decode 49.79 → 49.69 t/s (parity)
```

## Verification protocol used for this PR's bit-stability claim

The in-tree `speed-bench/*_variant_bench` harnesses show a slot-order bias on this configuration (full-vocab mismatch on control-vs-control) — I filed a separate issue report for it, attached. So the bit-stability was verified via the engine's own debug dump-diff protocol instead:

```bash
# for each variant: dump the layer-43 result_output tensor at pos0, binary-compare
DS4_METAL_GRAPH_DUMP_PREFIX=/tmp/up_a DS4_METAL_GRAPH_DUMP_NAME=result_output \
DS4_METAL_GRAPH_DUMP_POS=0 ./ds4 -m ds4flash.gguf --prompt-file /tmp/prompt.txt
```

Compared across factory `main`, kernel-subset, host-subset, and full branch tips at multiple prompt sizes (96 / 512 / 2048 / 5000-word): `nan=0 inf=0` everywhere, deterministic per build, no all-scan anomalies. Several negative candidates were found on the way (not in this PR): a Metal runtime-arg toggle family that produced broken Metal shader states (archived at `toggle-attempts-ref` in `daemon2k3/ds4` for the record, plus bisect notes), and an f16 indexer-compressed-K cache (`indexer-comp-f16` — nondeterministic at depth, parked).

Honest gaps: precision interplay with the existing NAX/quality paths (M5-gated; untouched by this PR); stream-changing vvexpf is opt-in only; Apple M3 Ultra is the only hardware this was measured on (Kernel-level invariant claims rest on identical math/order, marked as such in each commit message).

## How to review

Branch: `daemon2k3/ds4:code-layer-wins` (tip `594757c`). Recommended reading order: the big host-side commit, then the kernel commit, then the zero-copy/C1 commits. Rollback flags follow the engine's one-env-per-change convention where applicable (`DS4_DISABLE_ZEROCOPY_PAYLOAD_SPANS`; `DS4_SAMPLE_ACCELERATE_EXP` is itself an opt-in).

Open upstream-prep question kept from the session: the slot-order bias in the bench ABBA harness (separate issue with exact repro lines included).

## Evidence

A/B ledger, dump-diff protocol table, C1 (vvexpf) micro-data, the toggle-series bisect kill-chain, and ops notes: `research/m3ultra-dsv4-study/evidence/README.md` (same fork, `research` branch).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,14 @@ UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Darwin)
 NATIVE_CPU_FLAG ?= -mcpu=native
 SAMPLING_TEST :=
+# Optional cross-TU optimization for Apple Silicon release builds:
+#   make DS4_LTO=thin
+DS4_LTO ?=
+ifneq ($(strip $(DS4_LTO)),)
+CFLAGS += -flto=$(DS4_LTO)
+OBJCFLAGS += -flto=$(DS4_LTO)
+LDFLAGS += -flto=$(DS4_LTO)
+endif
 else
 NATIVE_CPU_FLAG ?= -march=native
 SAMPLING_TEST := tests/test_sampling
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ DS4_DSPARK_MODEL ?= $(DS4_TEST_MODEL)
 DS4_DSPARK_SUPPORT ?= gguf/DeepSeek-V4-Flash-DSpark-support-0731.gguf
 
 ifeq ($(UNAME_S),Darwin)
-METAL_LDLIBS := $(LDLIBS) -framework Foundation -framework Metal
+METAL_LDLIBS := $(LDLIBS) -framework Foundation -framework Metal -framework Accelerate
 CORE_OBJS = ds4.o ds4_distributed.o ds4_tp.o ds4_ssd.o ds4_metal.o ds4_layer_pack.o
 CPU_CORE_OBJS = ds4_cpu.o ds4_distributed.o ds4_tp.o ds4_ssd.o ds4_layer_pack.o
 else
--- a/ds4.c
+++ b/ds4.c
@@ -49910,6 +49910,26 @@ static int payload_write_tensor_span(FILE *fp, const ds4_gpu_tensor *tensor,
         payload_set_err(err, errlen, "session tensor is smaller than the payload");
         return 1;
     }
+    /* Zero-copy path on Apple unified memory: write() straight from the shared
+     * Metal buffer (no 8MB staging loop, no second memcpy). */
+    const void *host = ds4_env_cached("DS4_DISABLE_ZEROCOPY_PAYLOAD_SPANS") != NULL
+        ? NULL : ds4_gpu_tensor_const_host_ptr(tensor, offset);
+    if (host) {
+        if (bytes == 0) return 0;
+        const int fd = fileno(fp);
+        const uint8_t *p = (const uint8_t *)host;
+        uint64_t left = bytes;
+        while (left) {
+            const ssize_t w = write(fd, p, (size_t)(left > (uint64_t)SIZE_MAX ? SIZE_MAX : left));
+            if (w <= 0) {
+                payload_set_err(err, errlen, "failed to write session payload");
+                return 1;
+            }
+            p += w;
+            left -= (uint64_t)w;
+        }
+        return 0;
+    }
     uint64_t done = 0;
     while (done < bytes) {
         const size_t n = bytes - done > (uint64_t)cap ? cap : (size_t)(bytes - done);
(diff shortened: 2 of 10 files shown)


adriangalilea:
Data points from Apple M3 Ultra (60-core GPU) 512 GB, DeepSeek V4 Flash 0731 native-MXFP4 GGUF (156 GB), individual changes cherry-picked onto 84cc882:

**The simd_max finalize reducers (dsv4_kv/dsv4_rope/norm hunks) are not bit-exact here.** Full-vocab logits differ from baseline at every frontier (2048/4096/6144/8192, `ds4-bench --dump-frontier-logits-dir` byte-compare), max |Δ| ≈ 1.7 — argmax happened to be stable on my prompt, but that's luck, not a guarantee. IEEE max is order-exact, so the reduction must be over a different *set*: my guess is the two-simdgroup exchange (`m64 = simd_max(scratch[tid])` runs for all 128 threads while only 64 slots were written, and lanes ≥ 64 fold uninitialized scratch into the max), or the hadamard per-simdgroup max not matching the original per-32-value block alignment. Speed was 0.0% on a 2k→16k sweep here, so nothing lost by dropping or fixing it.

**The rb16 prefill dispatch IS bit-exact on MXFP4** (in-tree ABBA harness, 8/8 prefill runs + 273 decode rows, full-vocab) **but flat at short context**: −0.01% at an 8k prefix, −0.4% across a 2k→16k sweep. Consistent with your numbers living at ~117k ctx — below ~16k the attention stage is too small a slice for rb16 to show.

Calibration note relevant to #793: on this machine (MXFP4, ctx 8193) `metal_prefill_variant_bench` control-vs-control is bit-exact, 8/8 runs — the harness itself is sound here.

PR #795, opened 2026-08-13
cuda: persistent expert cache + Q2Q4 mixed-size expert cache optimiza…

ruota:
…tions

- Persistent CUDA SSD streaming expert cache
- Two-class size-split expert cache for mixed Q2/Q4K models (enable Q4K boosted layers 37-42 to use the cache instead of mapped views per token)
- Proportional byte-budget split across expert size classes
- Optional persistent-source page-drop for cold-expert RAM headroom
- CUDA-only, opt-in, no impact on Metal/ROCm/CPU

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -33,6 +33,7 @@
 #include <sys/file.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
+#include <sys/types.h>
 #if defined(__APPLE__)
 #include <sys/sysctl.h>
 #endif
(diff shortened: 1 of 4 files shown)


JordiPosthumus:
A small Linux reclamation finding that may help this work: the `POSIX_MADV_DONTNEED` call inherited by `cuda_model_discard_source_pages_impl()` does not release pages on glibc. This is a pre-existing helper issue, not a finding against the persistent-cache mechanism itself.

I checked current upstream (`b0a147a`) and our GB10/CUDA 13.0 integration. NVCC preprocessing of the latter removes the helper body entirely because `<sys/mman.h>` is absent. Including that header is not sufficient: glibc deliberately implements `POSIX_MADV_DONTNEED` as a successful no-op. [Linux man-pages explanation](https://man7.org/linux/man-pages/man3/posix_madvise.3.html).

A bounded 64 MiB read-only file-mapping probe on GB10/glibc 2.39 confirmed:

- Warm mapping: 65,536 KiB RSS.
- `posix_fadvise(DONTNEED)`, then `posix_madvise(POSIX_MADV_DONTNEED)`: both return success; RSS and resident file pages remain unchanged.
- Linux `MADV_DONTNEED` followed by file advice releases the pages, and rereading the immutable file returns identical bytes. **It is not a safe blanket replacement for private modified/anonymous mappings.**

I also tested the non-destructive `MADV_PAGEOUT` alternative. It preserves copied/refaulted bytes, but made a synthetic 64 MiB staged CUDA copy roughly 60–64 ms versus 3.5–4.5 ms with retention enabled. That is not a measurement of this PR's throughput; it is why I would not substitute it into inference-time staging blindly.

Worth separating the verified no-op from any OOM claim, and checking the compiled helper plus actual residency when evaluating the source-page-drop option.

ruota:
Thanks, this was correct.

I confirmed both issues: `ds4_cuda.cu` lacked a direct `<sys/mman.h>` include on this build path, and on glibc, `posix_madvise(..., POSIX_MADV_DONTNEED)` is a successful no-op.

The update now:

* uses Linux `madvise(..., MADV_DONTNEED)` only for the fd-associated, read-only, whole-file model mapping;
* skips anonymous, synthetic, and mismatched mappings, ensuring that this cannot discard the stack/pinned sources used by the CUDA tests;
* preserves the transient selected-expert upload path without mmap discard;
* performs mapping discard before `POSIX_FADV_DONTNEED` for persistent/cache-backed ranges;
* fixes the support-model case where the main mapping could be checked against the support model's registered size;
* documents the immutable-mapping contract and the KEEP/DROP environment-variable behavior.

I also added a Linux host-only regression test (`python3 tests/test_cuda_source_pages.py`). It exercises the production helper with a real mapping, confirms that its RSS drops from 68 KiB to 0 KiB, verifies byte-identical refaults, and uses caller stubs to cover advice ordering, KEEP/DROP combinations, mismatched fd/map associations, persistent-copy failures, and the support-model-size case.

PR #799, opened 2026-08-13
Metal: per-stream command-queue overlap for batched session decode (~1.7x measured)

sethconvex:
## What

When `ds4_sessions_eval_batch` receives an all-DeepSeek, non-TP decode batch, encode each session's decode pass on **its own Metal command queue** and commit without waiting, then collect all streams. Independent queues let the GPU overlap the latency-bound per-token kernel chains that a single serial queue executes back-to-back.

Mechanics:
- thread-local stream index (`ds4_gpu_set_stream`), lazily created per-stream queues (each gets the model residency set);
- the lazily-grown scratch buffers (`g_*_buffer/_bytes`, transient pool) become per-stream sets so concurrently executing streams never share mutable GPU state — this is the bulk of the diff and is mechanical;
- `ds4_gpu_end_commands_async` / `ds4_gpu_wait_stream`, plus a public `ds4_session_eval_begin/_end` pair for library/CLI callers;
- CPU-side encoding stays single-threaded; only GPU execution overlaps.

Best paired with `DS4_METAL_MODEL_UNTRACKED=1`: hazard tracking on the two shared model buffers otherwise serializes the queues (measured +63% from that flag alone under multi-stream).

## Why

Batch-1 decode on Apple silicon is latency-bound, not bandwidth-bound: an M3 Ultra (2x cores, 1.5x bandwidth of an M5 Max) decodes only ~7% faster. A microbenchmark of dependent skinny-kernel chains shows 4 queues overlap at ~2.3x aggregate on M5 Max. This PR converts that idle capacity into multi-session throughput.

## Measured

M5 Max 128GB, DeepSeek V4 Flash IQ2XXS, inside `ds4-server --batched-session 8` (machine concurrently running an unrelated PyTorch training job, so absolute numbers are depressed; ratios are same-conditions):

| decode path | ms per token per session |
|---|---|
| solo request (count=1) | 116 |
| existing native Metal session batch | ~110/row (count 2-3) |
| **this PR, count=8** | **68.5/row (~1.7x)** |

CLI validation on the same machine: at temperature 0, N concurrent streams produce byte-identical output to solo runs (verified N=4, with and without the untracked flag).

## Scope / fallbacks

- GLM sessions, tensor-parallel, and mirror mode keep the existing single-queue path untouched.
- `DS4_METAL_NO_STREAM_OVERLAP=1` restores current behavior entirely.
- Currently overlap takes precedence over the native fused session batch when both apply; happy to flip that default or gate per-backend if you prefer — under our measurements the native path was ~110ms/row while overlap was ~68ms/row, but that deserves confirmation on an uncontended machine and on CUDA-era hardware it may differ.

## Caveats

- Tested on one machine (M5 Max), under GPU contention; would value numbers from an idle M3 Ultra / M4 Max.
- Waiting only the last committed buffer per queue relies on per-queue FIFO completion; capped at 8 streams.
- In the spirit of the project's README: this was built with strong AI assistance (Claude), with measurements driving every step. Full diagnostic trail (dispatch census, per-stage decode budget, queue-overlap microbenchmarks) available if useful.

https://claude.ai/code/session_0194nbHLo3afzCgw1QW3zpTX

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -62185,8 +62185,12 @@ static int ds4_sessions_eval_batch_metal(
     }
     bool native_shared = false;
     bool native_qkv = false;
+    /* SSD streaming's expert-cache slabs are process-global mutable state,
+     * so concurrently executing streams must not share them: keep streaming
+     * configurations on the single-queue path. */
     const bool stream_overlap =
         overlap_all_deepseek && !e->tp.active && !mirror && count > 1 &&
+        !e->ssd_streaming &&
         getenv("DS4_METAL_NO_STREAM_OVERLAP") == NULL;
     bool ok = true;
     if (stream_overlap) {
--- a/ds4.c
+++ b/ds4.c
@@ -49333,6 +49333,9 @@ struct ds4_session {
     void *cancel_ud;
     uint32_t prefill_cap;
     int ctx_size;
+    int stream_id;
+    int pending_token;
+    bool pending_eval;
     bool checkpoint_valid;
     bool mtp_draft_valid;
     bool greedy_splitkv_anchor_valid;
@@ -61668,6 +61671,80 @@ static int ds4_session_eval_probe_tp(ds4_session *s, int token, bool probe_mtp,
     return rc;
 }
 
+void ds4_session_set_stream(ds4_session *s, int stream_id) {
+    if (s) s->stream_id = stream_id;
+}
+
+/* Async decode pair for multi-stream generation: _begin encodes and commits
+ * one token's decode pass on the session's own GPU stream without waiting;
+ * _end waits for that stream, reads logits, and commits the token to the
+ * session timeline.  Callers interleave other sessions' _begin calls in
+ * between; independent per-stream queues let the GPU overlap them.  Metal
+ * encoding must stay serialized by the caller (one thread, or a lock around
+ * _begin). */
+int ds4_session_eval_begin(ds4_session *s, int token, char *err, size_t errlen) {
+    if (!s) return -1;
+#ifdef DS4_NO_GPU
+    (void)token; snprintf(err, errlen, "GPU support is not compiled in");
+    return -1;
+#else
+    if (s->distributed || ds4_session_is_cpu(s) || !s->checkpoint_valid) {
+        return ds4_session_eval(s, token, err, errlen);   /* sync fallback */
+    }
+    int room = s->ctx_size - s->checkpoint.len;
+    if (room <= 1) { snprintf(err, errlen, "context is full"); return -1; }
+    ds4_engine *e = s->engine;
+    ds4_gpu_set_stream(s->stream_id);
+    if (ds4_gpu_begin_commands() == 0) {
+        snprintf(err, errlen, "%s begin failed", ds4_backend_name(e->backend));
+        return -1;
+    }
+    if (!metal_graph_encode_token_raw_swa(&s->graph, &e->model, &e->weights,
+                                          token, (uint32_t)s->checkpoint.len,
+                                          true, false)) {
+        (void)ds4_gpu_synchronize();
+        snprintf(err, errlen, "%s decode encode failed", ds4_backend_name(e->backend));
+        s->checkpoint_valid = false;
+        return -1;
+    }
+    if (ds4_gpu_end_commands_async() == 0) {
+        snprintf(err, errlen, "%s commit failed", ds4_backend_name(e->backend));
+        s->checkpoint_valid = false;
+        return -1;
+    }
+    s->pending_token = token;
+    s->pending_eval = true;
+    return 0;
+#endif
+}
+
+int ds4_session_eval_end(ds4_session *s, char *err, size_t errlen) {
+    if (!s) return -1;
+#ifdef DS4_NO_GPU
+    snprintf(err, errlen, "GPU support is not compiled in");
+    return -1;
+#else
+    if (!s->pending_eval) return 0;
+    s->pending_eval = false;
+    ds4_engine *e = s->engine;
+    if (ds4_gpu_wait_stream(s->stream_id) == 0) {
+        snprintf(err, errlen, "%s stream wait failed", ds4_backend_name(e->backend));
+        s->checkpoint_valid = false;
+        return -1;
+    }
+    if (ds4_gpu_tensor_read(metal_graph_logits(&s->graph), 0, s->logits,
+                            (uint64_t)DS4_N_VOCAB * sizeof(s->logits[0])) == 0) {
+        snprintf(err, errlen, "%s logits readback failed", ds4_backend_name(e->backend));
+        s->checkpoint_valid = false;
+        return -1;
+    }
+    token_vec_push(&s->checkpoint, s->pending_token);
+    s->mtp_draft_valid = false;
+    s->checkpoint_valid = true;
+    return 0;
+#endif
+}
+
 int ds4_session_eval(ds4_session *s, int token, char *err, size_t errlen) {
     bool probe_mtp = true;
 #ifndef DS4_NO_GPU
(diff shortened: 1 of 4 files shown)


sethconvex:
Test status on M5 Max with the June `chat-v2` Flash q2 GGUF:

- `make test-metal-session-batch` (the dedicated test for the touched path): **PASS** — `sessions=2 steps=6 mixed_suffix=8 exact_logits=1`, batches routed through the new overlap path.
- Full `ds4_test`: 5 assertion failures (ds4_test.c:4957 x2, :4981 x2, :5278 top5_overlap) — **identical failures reproduce on pristine main** (84cc882) with the same GGUF, so they pre-date this PR; presumably test-vector drift against the older model file revision. Happy to re-run against a current GGUF if pointed at one.

sethconvex:
Speed regression check per CONTRIBUTING.md — same machine, back-to-back runs, same model file.

Machine: MacBook Pro M5 Max 128GB, Metal. Model: DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix.gguf. **Caveat: an unrelated PyTorch training job was running throughout (both runs equally), so absolute t/s are depressed and noise is elevated; treat deltas within ~±7% as noise. Re-running on idle hardware would be worthwhile.**

Command:
```
./ds4-bench -m <gguf> --prompt-file speed-bench/promessi_sposi.txt \
  --ctx-start 2048 --ctx-max 8192 --step-incr 2048 --gen-tokens 64 --csv out.csv
```

| ctx | gen_tps main (84cc882) | gen_tps this PR | prefill_tps main | prefill_tps PR |
|---|---|---|---|---|
| 2048 | 22.27 | 22.79 | 374.5 | 305.5 |
| 4096 | 20.68 | 19.31 | 275.8 | 259.5 |
| 6144 | 20.42 | 20.23 | 268.3 | 264.8 |
| 8192 | 18.76 | 20.63 | 250.4 | 272.1 |

Single-stream generation and prefill are within run-to-run noise in both directions (gen mean 20.5 vs 20.7); the per-stream scratch indexing shows no measurable single-stream cost. The multi-session gains reported in the PR description are on top of this unchanged baseline.

Correctness commands run: `make test-metal-session-batch` (PASS, exact_logits=1), full `ds4_test` (5 pre-existing failures, identical on pristine main with this June-era GGUF — details in previous comment).

sethconvex:
Idle-machine follow-up (training job that contended earlier runs has finished; no other load):

**Single-stream `ds4-bench` sweep** (same command as above):

| ctx | gen_tps main | gen_tps this PR |
|---|---|---|
| 2048 | 38.74 | 41.60 |
| 4096 | 34.78 | 36.15 |
| 6144 | 29.58 | 34.28 |
| 8192 | 28.44 | 34.09 |

No single-stream regression (PR side ran first and coolest, so I'd attribute its apparent edge mostly to thermal ordering — the safe claim is parity).

**End-to-end through `ds4-server --batched-session 8` + `DS4_METAL_MODEL_UNTRACKED=1`**, 80-token completions:
- 1 request: 19.3 t/s
- 8 concurrent: **36.1 t/s aggregate (1.87x)** — request wall-time including each request's serialized prefill; the decode-phase overlap is higher (68.5 vs 116 ms/token per the earlier per-batch timings).

sethconvex:
Correction/refinement on the previous comment's server numbers — a background process was still running during that "idle" measurement. On a genuinely idle machine:

- solo request: 28.9 t/s end-to-end; 8 concurrent (80-token completions): 39.6 t/s aggregate = **1.37x**.
- The lower multiplier vs the contended run (1.87x) is Amdahl on the **serialized per-request prefill**, which dominates short completions once decode is fast. The decode-phase overlap itself is stable at ~1.7x (68.5 vs 116 ms/token from `DS4_SERVER_BATCH_LOG` timings), so longer generations approach that end-to-end.
- Follow-up opportunity (out of scope here): letting prefill of a joining session overlap in-flight decode of others via the same per-stream queues would remove that serial section — possibly composing with the existing `--mixed-prefill-quantum` machinery.

magicseth:
@antirez

sethconvex:
Known limitations and design trade-offs of this approach, for review consideration:

**Correctness / safety**
- **SSD streaming excluded** (pushed as a follow-up commit): the streamed-expert cache slabs are process-global mutable state outside the per-stream scratch split, so streaming configurations stay on the single-queue path.
- **Relies on `DS4_METAL_MODEL_UNTRACKED=1` for most of the gain** — without it, hazard tracking on the shared model buffers re-serializes the queues. Untracked is sound while model buffers stay strictly read-only at inference time; anything that ever writes them (imatrix collection, steering experiments) would corrupt silently rather than stall. Worth deciding whether batched mode should set it implicitly or keep it opt-in.
- **Maintenance invariant**: every *future* lazily-grown scratch buffer must follow the per-stream indexing convention, or it silently reintroduces a cross-stream race that single-stream tests will never catch. If this lands, a comment near the scratch declarations (or a debug assertion pattern) may be worth adding.
- With more than 8 sessions mapped onto 8 streams, completion waits rely on per-queue FIFO command-buffer completion (standard Metal behavior, but an assumption).

**Behavior / operational**
- Throughput-for-latency trade: per-request speed drops as concurrency rises (~5–12 t/s per stream at N=8 on M5 Max). No priority/QoS between sessions.
- Sustained peak GPU load: single-stream decode's idle gaps effectively yielded to co-tenant GPU work; filling them means max power draw, thermals, and starvation of anything else using the GPU.
- Per-stream memory: each active stream duplicates the ~30 lazily-grown scratch buffers in addition to its session context buffers; fine at 128 GB with the q2 Flash, tighter on 96 GB machines. No accounting/cap yet.
- Per-request prefill remains serialized through the worker, which caps end-to-end gains for short completions (measured 1.37x at 80 tokens idle vs ~1.7x decode-phase); overlapping a joining session's prefill with in-flight decode is the natural follow-up.

**Evidence limits**
- All measurements are from one machine (M5 Max 128 GB, Metal, q2 Flash). The gate currently prefers stream overlap over the native fused session batch wherever both apply; that default is justified only by this machine's numbers and deserves confirmation elsewhere (and may well be wrong for CUDA-class backends).

sethconvex:
Completing the CONTRIBUTING.md speed-regression record with the default linear sweep — plus a methodology note that matters on shared dev machines.

**1. Full default sweep** (`--ctx-start 2048 --ctx-max 65536 --step-incr 2048 --gen-tokens 128`) was run on both this branch and main (84cc882). The two ~20-minute runs were sequential, and uncontrolled background load shifted between them, producing implausible deltas in both directions (this branch "+140%" at some frontiers, "-96%" at others). I'm not presenting those CSVs as a comparison — sequential long sweeps aren't valid on this machine.

**2. Interleaved A/B instead**: three alternating short sweeps per branch (PR, main, PR, main, ...; fresh build + model load each), per-frontier medians:

| ctx | main gen med (3 runs) | PR gen med (3 runs) | delta |
|---|---|---|---|
| 2048 | 31.38 | 30.30 | -3.4% |
| 4096 | 27.99 | 29.32 | +4.8% |
| 6144 | 27.23 | 28.94 | +6.3% |
| 8192 | 26.82 | 29.36 | +9.5% |

Median-of-medians gen ratio 1.055; prefill median ratios 0.84–1.11. Within-branch run spreads were ±30%, so the honest claim is **parity — no detectable single-stream regression**, consistent with the earlier calm-window back-to-back result (20.5 vs 20.7 t/s means). A quiet machine (which this dev box rarely is) or CI hardware would tighten the bound; the per-stream scratch indexing adds one thread-local load + array index per scratch access at encode time, so a large hidden regression is also not mechanically plausible.

sethconvex:
Final synthesis after a full matrix of conditions on the quiet machine (M5 Max, q2 Flash, 8x300-token completions via `--batched-session 8`):

| serving mode | aggregate t/s |
|---|---|
| sequential/FIFO (requests one at a time) | 37.5 |
| concurrent-8, **this PR's stream overlap** | 32.0 |
| concurrent-8, current native Metal session batch | 6.1 |

Honest framing of what this PR is and is not:

- **It is not an aggregate-throughput win over FIFO on an idle machine.** Current main's fused kernels + intra-token concurrent encoding already keep the GPU busy at batch-1 (~46 t/s solo here); cross-stream overlap can't create capacity that no longer exists, and costs ~15% vs FIFO from contention.
- **It is a 5x improvement to what batched mode actually does on Metal today.** The native session-batch path measured 6.1 t/s aggregate under the same workload — concurrent serving currently costs 6x throughput; with this patch it costs 15%. The point of `--batched-session` is that every client streams immediately instead of queueing behind minutes of other requests; this PR makes that fairness nearly free.
- **It also wins outright when the GPU is shared** (measured 1.7–1.9x aggregate over the same path while an unrelated training job ran) and for short-burst workloads — single-stream decode competes poorly for GPU slices against co-tenant work; 8 streams reclaim them.

Earlier comments' larger multipliers were measured under those conditions (contention / short completions / pre-fusion-era baselines) — this table on a quiet machine with long generations is the conservative bound. If the maintainers prefer, the honest minimal version of this PR is: keep stream overlap as the Metal batched-decode path (replacing the native session batch there), making no claims beyond "batched serving stops being 6x slower than serial."

trueimage:
can you share the test/bench command, I can run on an m3 ultra with the Q4K or MXFP4 gguf

GiorgioOppo:
Hi @sethconvex, thanks for the PR, I implemented your optimization in PR #621

GiorgioOppo:
@sethconvex Don't close the issue for my comment, maybe Q4 will never go into production

sethconvex:
oh ok! Just trying not to add noise!

PR #813, opened 2026-08-16
Two-machine tensor parallelism for the ROCm backend

davidcanar:
> **Status (14 Sep 2026):** this branch is the historical base of a stack and stays at its original trunk point, so GitHub reports conflicts with current `main`; it is not being rebased. The same five commits, merged with trunk (and the GLM 5.3 work on top), are in #1024 (`glm-tp-rocm`), which is mergeable, so please review the ROCm TP transport there. Stack order: #813 → #835 → #1018 → #1024.

## Summary

ds4's two-machine tensor parallelism (`--tensor-parallel` with `--role coordinator|worker`) currently requires the **Metal backend** and Apple's Thunderbolt RDMA stack (`librdma.dylib`). This PR ports it to the **ROCm (HIP) backend**, so a pair of AMD machines can split a DeepSeek model 50/50 and exchange gate partials over any verbs device.

## What changed

- **ds4_server.c**: wire TP CLI parsing/validation, leader bind, and the mirrored worker loop (previously only the `ds4` CLI supported TP).
- **ds4.c**: enable `ds4_engine_tp_bind` on the ROCm build, validate Q4_K routed / Q8_0 shared+attention+output layouts (the ownership-aware kernels cover those), and allocate the transport slab as GPU/CPU-shared memory.
- **rocm/ds4_rocm_tp.cuh** (new, ~835 lines): TP gate machinery for the eager default-stream execution model — arrival/release flag words in the slab, a transport service thread, spin-wait kernels, and big-gate staging through pinned host-mapped memory. Plus the TP compute kernels: k-sliced Q8 matvec, attention output projection for the rank's head groups, fused HC-expand combine (mirrors the CUDA backend's `hc_expand_kernel` with `has_add=1`), and prefill row-range attention variants.
- **rocm/ds4_rocm_moe.cuh / _launch.cuh**: expert ownership (`tp_rank`/`tp_world`/rebased weight pointers) across the Q4_K kernels and sort filtering, so each rank computes only its resident expert half; the shared-expert partial is folded into the down projection (`add_in`).
- **ds4_tp.c**: Linux verbs support via dlopen (macOS keeps `librdma.dylib`, Linux uses `libibverbs`), a **dedicated bulk QP** for prefill row swaps with per-round recv barriers, the decode lookahead window armed at bring-up, and real error-code logging (the inline verbs API returns codes without setting errno).

## Design notes

- The transport slab uses `ds4_gpu_tensor_alloc_shared()` (new backend hook): Metal aliases its shared buffers; ROCm uses `hipHostMalloc(hipHostMallocMapped)`, coherent on APUs with `HSA_XNACK=1`.
- Gate release signaling differs from Metal by necessity (no shared events): the service thread publishes 64-bit release words (tagged row/batch/big sequences) in the slab's in-flags region, and stream-ordered spin kernels park the GPU until the exchange lands.
- The two-QP layout exists because several Thunderbolt verbs providers RNR-error the QP permanently on a send that races the peer's `post_recv`; bulk traffic gets its own QP with receives posted before a data-socket barrier, so no send can ever arrive without a posted recv.

## Testing

- Two Strix Halo boxes (gfx1151, ROCm 7.14), thunderbolt-ibverbs `usb4_rdma0` device, `--transport rdma` and TCP fallback both exercised.
- DeepSeek-V4-Flash Q4KExperts/Q8Attn/Q8Shared/Q8Out: each rank loads ~81 GiB of the 153 GiB model (half the routed experts).
- ~14 tok/s decode over RDMA; `--debug-hash 8` lockstep checks pass; long prefills (445 tokens, attention row split), concurrent sessions, and multi-turn sessions all stable.
- Model scope: DeepSeek models with Q4_K routed experts and Q8_0 shared/attention/output tensors (validated at bind time); GLM is rejected on ROCm for now.


MerlijnWajer:
I would like to test this PR, would you mind providing some instructions on how to test? I have four strix halo machines with thunderbolt <-> connectx mellanox RDMA capable network cards. I was also wondering what you used for the RDMA thunderbolt ibverb device.

davidcanar:
> I would like to test this PR, would you mind providing some instructions on how to test? I have four strix halo machines with thunderbolt <-> connectx mellanox RDMA capable network cards. I was also wondering what you used for the RDMA thunderbolt ibverb device.

I used a direct USB4 cable and the https://github.com/hellas-ai/thunderbolt-ibverbs/ driver

this is the command I run on the coordinator:

ds4-server \
  --warm-weights \
  -t 32 \
  --ctx 393216 --prefill-chunk 4096 \
  -m /home/davidccanar/models/DeepSeek-V4-Flash-Q4KExperts-F16HC-F16Compressor-F16Indexer-Q8Attn-Q8Shared-Q8Out-chat-v2-imatrix-0731.gguf \
  --role coordinator --tensor-parallel \
  --transport rdma \
  --listen 10.0.2.1 8088 \
  --kv-disk-dir ~/ds4/server-kv --kv-disk-space-mb 265536 \
  --host 0.0.0.0 --port 1234 --cors

And the worker:

ds4-server \
  --warm-weights \
  -t 32 \
  --role worker --tensor-parallel --coordinator 10.0.2.1 8088 \
  --ctx 393216 --prefill-chunk 4096 \
  -m /home/davidcanar/models/DeepSeek-V4-Flash-Q4KExperts-F16HC-F16Compressor-F16Indexer-Q8Attn-Q8Shared-Q8Out-chat-v2-imatrix-0731.gguf \
  --transport rdma \
  --kv-disk-dir ~/ds4/server-kv --kv-disk-space-mb 265536 \
  --host 0.0.0.0 --port 1234 --cors

You could also specify which rdma device to use with the parameters:

--rdma-device usb4_rdma0 --rdma-gid-index N

angeleitec-lang:
Tested this on real hardware: two AMD Strix Halo boxes (Ryzen AI Max+ 395, 128GB unified memory each) connected via **Mellanox ConnectX-3 Pro** running in **native InfiniBand mode** (not RoCE), ROCm 7.14, gfx1151, DeepSeek-V4-Flash-0731 (Q4KExperts/F16HC mixed quant, 153GB).

Built from this branch (`rocm-tensor-parallel`) with `make strix-halo CC="cc -fPIC"` (needed `-fPIC` added to `CC`, otherwise the final link fails with `relocation R_X86_64_32 cannot be used against local symbol; recompile with -fPIC` — the C objects are compiled without `-fPIC` by default but get linked via `hipcc`'s PIE-enabled driver).

`--tensor-parallel --transport rdma` starts both worker and coordinator fine, each loads its 50/50 shard correctly, but the RDMA connection setup fails:

```
ds4-server: tp rdma: no IPv4-mapped GID on the active port; give the Thunderbolt interface its own IPv4 (e.g. macOS 'sudo ifconfig en1 inet 10.99.0.2/30 alias', Linux 'sudo ip addr add 10.99.0.2/30 dev en1') on both machines
```

Checked the GID table on the IB port (`/sys/class/infiniband/<dev>/ports/1/gids/*` and `gid_attrs/types/*`): only a single native GUID-based link-local GID is present (`fe80::.../IB/RoCE v1`), no RoCEv2-style IPv4-mapped GID entry — which makes sense, since the fabric is native InfiniBand (LID-addressed), not RoCE/Ethernet. The `192.168.100.x` addresses are assigned via IPoIB on top, but that doesn't populate an IP-mapped GID for `rdma_cm` the way it would on a RoCE NIC or the Thunderbolt bridge.

So it looks like the RDMA connection setup currently assumes RoCEv2-style IP-to-GID resolution (via `rdma_cm`), which works for RoCE/Ethernet cards and presumably the Thunderbolt-ibverbs bridge, but doesn't cover native IB fabrics with LID-based addressing and no IP-mapped GID. Worth noting since the PR description mentions ConnectX Mellanox cards as supported hardware — that's presumably true for ConnectX cards configured in Ethernet/RoCE port mode, but not for ones running native IB (like ours, and probably any inherited-from-an-InfiniBand-switch setup).

Pipeline mode (`--role coordinator/worker --layers`) works great on the same hardware over plain TCP, for reference — this is specifically about the RDMA path in tensor-parallel mode.

Happy to test a fix if there's interest in native-IB support (LID/GID resolution instead of IP-mapped GID lookup), or if there's a config knob I'm missing to point it at the right GID index.

angeleitec-lang:
Follow-up to my earlier comment: found and fixed the native-IB GID issue, `--tensor-parallel --transport rdma` now works end-to-end on our two-Strix-Halo Mellanox setup.

Root cause: the RTR address-handle setup always builds a GRH with the peer's IPv4-mapped GID, and `tp_rdma_open` hard-fails if no such GID exists. Both of those are specifically Thunderbolt/RoCE-emulation requirements — real InfiniBand (LID-routed, no GRH) doesn't need or populate an IPv4-mapped GID at all. The peer's LID and `link_layer` are already exchanged in the info handshake, so the fix just uses them instead of failing:

```diff
@@ tp_rdma_open (GID selection) @@
+    } else if (r->port.link_layer == IBV_LINK_LAYER_INFINIBAND) {
+        /* Native InfiniBand (LID-routed) fabric: no GRH/GID needed at all,
+         * addressing goes by port LID directly. Real Mellanox HCAs (unlike
+         * the Thunderbolt/RoCE-style emulation this file was written
+         * against) don't populate an IPv4-mapped GID and don't need one. */
+        r->gid_index = 0;
+        memset(&r->gid, 0, sizeof(r->gid));
     } else {
         for (int i = 0; i < r->port.gid_tbl_len; i++) {
             ... (existing IPv4-mapped GID search, now only reached for non-IB link_layer)

@@ tp_rdma_register_and_exchange (RTR address handle) @@
         a.ah_attr.dlid = (uint16_t)r->peer.lid;
         a.ah_attr.port_num = 1;
-        a.ah_attr.is_global = 1;
-        memcpy(a.ah_attr.grh.dgid.raw, r->peer.gid, 16);
-        a.ah_attr.grh.sgid_index = (uint8_t)r->gid_index;
-        a.ah_attr.grh.hop_limit = 1;
+        if (r->port.link_layer == IBV_LINK_LAYER_INFINIBAND &&
+            r->peer.link_layer == IBV_LINK_LAYER_INFINIBAND) {
+            /* Native IB: route by LID, no GRH needed. */
+            a.ah_attr.is_global = 0;
+        } else {
+            a.ah_attr.is_global = 1;
+            memcpy(a.ah_attr.grh.dgid.raw, r->peer.gid, 16);
+            a.ah_attr.grh.sgid_index = (uint8_t)r->gid_index;
+            a.ah_attr.grh.hop_limit = 1;
+        }
```

Kept the existing GRH/IPv4-GID path untouched for everything else (Thunderbolt, RoCE) — this only takes the LID branch when both sides report `IBV_LINK_LAYER_INFINIBAND`.

Result on our two Strix Halo boxes (Mellanox ConnectX-3 Pro, native IB, DeepSeek-V4-Flash-0731 153GB, 50/50 split, ctx=262144):

```
ds4-tp: rdma device ibp196s0 (port state 4)
ds4-tp: worker connected, transport=rdma
tensor parallelism bound: rank 0, 50/50 expert split, rdma transport
...
chat ctx=6..56:50 gen=50 THINKING decoding chunk=17.54 t/s avg=17.54 t/s
```

~17.5 tok/s decode over real RDMA, correct output. Happy to open a PR with this against your branch if useful, or you can cherry-pick the diff above directly — whichever's easier.

PR #815, opened 2026-08-16
server: drop a partial trailing UTF-8 sequence from finished text

adriangalilea:
When a generation stops inside a multibyte token (`max_tokens`, a stop sequence), the last bytes of the generated text are an incomplete UTF-8 sequence. Streaming already holds such a tail back between chunks (`utf8_stream_safe_len`), but the final flush and the non-stream body emitted it as-is, so the JSON response carried invalid UTF-8.

Repro on `84cc882` (Metal, DeepSeek V4 Flash MXFP4 0731, `--batched-session 4`): a Spanish group-chat prompt whose greedy first token is 👋, `"max_tokens": 1`, non-stream. The body contains `"content":"\xf0\x9f\x91"` (three of the four bytes) and Python's `json.loads` fails with `UnicodeDecodeError: invalid continuation byte`; any strict client does the same. With this change the same request returns `"content":""` and `max_tokens: 2` returns `"👋"`, as before.

The change applies the same boundary once to the finished text, before any response is built (stream final flush or non-stream body). Only the partial sequence is dropped; complete sequences and ASCII are untouched, and the streamed prefix is unaffected since it never crossed that boundary. Unit test `test_generation_tail_drops_partial_utf8` added; `./ds4_test --server` passes (macOS, Metal build). The live checkpoint keeps the full token; a client replaying the returned text will not extend the session in this corner case, which it could not before either (the text was undecodable).

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -11991,6 +11991,15 @@ decode_again:
                             &last_decode_log_completion);
     }
 
+    /* A generation cut mid-codepoint (max_tokens, a stop sequence) must not
+     * leak the partial sequence: no client can decode it and the JSON body is
+     * required to be UTF-8. Streaming already holds such a tail back between
+     * chunks; the final flush and the non-stream body have to drop it. */
+    if (text.ptr) {
+        text.len = utf8_stream_safe_len(text.ptr, 0, text.len, false);
+        text.ptr[text.len] = '\0';
+    }
+
     if (j->req.stream && !structured_stream && text.len > plain_stream_pos) {
         char *tail = xstrndup(text.ptr + plain_stream_pos, text.len - plain_stream_pos);
         if (!sse_chunk(j->fd, &j->req, id, tail, NULL)) {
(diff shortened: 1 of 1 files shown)


tim-chiron:
Still reproduces on current `main` (`0aaea5a`): Metal, M3 Ultra, DeepSeek-V4.1-Flash Q4 (`DeepSeek-V4.1-Flash-Q4.gguf`), `--batched-session 4`, non-streaming.

```sh
curl -s http://127.0.0.1:8090/v1/chat/completions -H 'Content-Type: application/json' \
  -d '{"model":"deepseek-v4.1-flash","messages":[{"role":"user","content":"ping"}],"max_tokens":4,"temperature":0,"chat_template_kwargs":{"enable_thinking":false}}'
```

The body contains `"content":"Pong! \xf0\x9f"`, the first two bytes of a four-byte emoji. With `finish_reason: "length"`, the response is invalid UTF-8, and Python's `json.loads` on the raw bytes raises `UnicodeDecodeError`.

We hit it in practice through a client that sends a 4-token reachability ping.

PR #817, opened 2026-08-16
ROCm: env opt-in to keep the Q8->F16 weight cache in multi-model (MTP) setups

jyatesdotdev:
## What this does

Fixes a silent prefill regression in multi-model ROCm setups, behind an opt-in environment variable.

The chain of events being fixed:

- Loading a second model map — a multi-token-prediction or DSpark support model — trips the runtime's multi-model guard, which disables the optional Q8→F16 expanded-weight cache for the entire process.
- With the cache disabled, the attention-output matrix multiply runs on the direct Q8 path instead of the hipBLAS F16 path. At prefill shapes the direct path is 5.2 times slower: 348.7 ms against 66.7 ms per layer per 4096-token chunk in the original measurement.
- Setting `DS4_ROCM_Q8_F16_CACHE_MULTI_MODEL=1` lets the cache build alongside the second model map. The existing per-allocation budget check (a 512 MiB reserve on GPUs with 112 GiB or more) still guards against overcommit, and an allocation failure falls back to the Q8 path with a log line.
- Without the environment variable set, behavior is completely unchanged.

Measured effect with a support model loaded: a 10,500-token prefill went from 74.2 s to 61.5 s — 21% faster. Decode is unchanged.

The pull request also includes the two diagnostics that found the problem: `DS4_ROCM_BATCH_LAYER_TIMING` (per-layer GPU timing with command-buffer boundaries) and `DS4_DIST_SYNC_TIMING` (a readback timer for distributed synchronization).

## Correctness

On Radeon 8060S (gfx1151, TheRock toolchain), rebased on current main:

- End to end on two machines with greedy decoding and the released Q4_K-experts model file:
  - with the variable unset, output is byte-identical to unmodified main — the default path is untouched;
  - with the variable set, runs complete with correct output. In single-model runs the setting changes nothing, since the guard only trips when a second model is loaded.
- Numerics: with the variable set in a multi-model setup, the expanded-weight cache runs in F16 instead of falling back to Q8. This is the more precise path and is deterministic across runs, but it is not bit-identical to the Q8 fallback. Anyone who needs to reproduce the old fallback exactly leaves the variable unset.

## How to test

`make strix-halo`, then any ROCm run with a multi-token-prediction support model loaded. Compare prefill rates with and without `DS4_ROCM_Q8_F16_CACHE_MULTI_MODEL=1`.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -59664,13 +59664,31 @@ int ds4_session_eval_layer_slice(ds4_session *s,
         }
     } else {
         if (ok) ok = ds4_gpu_begin_commands() != 0;
+        static int batch_layer_timing = -1;
+        if (batch_layer_timing < 0)
+            batch_layer_timing = getenv("DS4_ROCM_BATCH_LAYER_TIMING") != NULL ? 1 : 0;
         for (uint32_t il = layer_start; ok && il <= layer_end; il++) {
+            double lt0 = 0.0;
+            if (batch_layer_timing) {
+                if (ok) ok = ds4_gpu_end_commands() != 0;
+                if (ok) ok = ds4_gpu_synchronize() != 0;
+                lt0 = now_sec();
+                if (ok) ok = ds4_gpu_begin_commands() != 0;
+            }
             ok = metal_graph_encode_layer_batch(g,
                                                 &e->model,
                                                 &e->weights.layer[il],
                                                 il,
                                                 pos0,
                                                 n_tokens);
+            if (batch_layer_timing) {
+                if (ok) ok = ds4_gpu_end_commands() != 0;
+                if (ok) ok = ds4_gpu_synchronize() != 0;
+                if (ok) ok = ds4_gpu_begin_commands() != 0;
+                fprintf(stderr,
+                        "ds4: batch layer timing pos=%u n=%u il=%u elapsed=%.1f ms\n",
+                        pos0, n_tokens, il, (now_sec() - lt0) * 1000.0);
+            }
         }
     }
     if (ok && output_logits) {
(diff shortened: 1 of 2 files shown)


jessa0:
it's not clear to me why the new behaviour is hidden behind an env var here.. it's claimed that the new behaviour is more precise and faster.

jyatesdotdev:
Since I only have the gfx1151,I was making it opt in for others. I can always remove that. I might have got ahead of myself publishing these.

jyatesdotdev:
@jessa0 btw thanks for taking a look at this, appreciate it (and thanks for testing that other PR!). It seems like the gating of cache-off-under-multi-model behavior is consistent with upstream, and according to my agent this adds VRAM pressure that might not work as well on other architectures (even though I got +21% on gfx1151). I'm going to double check the numbers when I get home later, and I'm good with flipping it to be default-on with the env acting as opt out instead if that makes sense.

jessa0:
> @jessa0 btw thanks for taking a look at this, appreciate it (and thanks for testing that other PR!)

I'm happy to have any perf boost I can get, so thank you!

> this adds VRAM pressure that might not work as well on other architectures

Ah, I see. So many tunables. Not sure about defaults then! Seems like the project as a whole needs to look at how to make the plethora of tunables manageable then. Might be worth looking at what llama.cpp does for these kinds of things.. I guess the Linux kernel approach for example is to have "middle of the road" defaults, and then auto-tune at runtime, and have a maze full of variables to manually tune in /sys.

jyatesdotdev:
Closing this pull request to cut down the number of open submissions from me. The Q8 to F16 expanded-weight cache change is being folded into a single performance branch rebased on current main, together with my other ROCm work, and will return as one combined submission with fresh measurements against the current tree.

PR #818, opened 2026-08-16
server: live-KV rewind-reuse, prompt-window timing, first-chunk prefill log fix

jyatesdotdev:
## What this does

Two server-side prefill improvements and a logging fix. No kernel or numeric path changes.

1. **Reuse the live key-value cache on rewind.** When an incoming prompt is a strict prefix of the tokens already in the live session, the server rewinds to one token before the prompt end and re-evaluates that single token to restore the logits, instead of re-running the whole prefill. When a prompt diverges from the live session after a long shared prefix (at least `DS4_KV_REWIND_MIN_TOKENS` tokens, default 256), the server rewinds to the shared prefix and prefills only the suffix. The discarded tail of the live session is snapshotted first, under the normal eviction policy. In the production logs that motivated this, 84 of 86 cache misses were exact repeats of an earlier prompt, each paying a full re-prefill. The decision logic lives in one function, `kv_rewind_reuse_target()`, with unit tests in the server suite. Setting `DS4_KV_REWIND_REUSE=0` disables the feature.

2. **Prefill phase timing.** With `DS4_PREFILL_TIMING` set, the server logs the cold, sync, and tail phases of each prompt window, separating fixed per-prompt cost from cost proportional to token count.

3. **First-chunk logging fix.** The first progress callback zeroed its measurement interval, so the first prefill chunk always logged a rate of 0.00 tokens/s. The bug is most visible in distributed mode, where the first callback covers an entire pipeline chunk, and it exists in main today; a reviewer flagged it on this pull request. The first interval is now measured from the start of the prefill.

The GLM-specific mandatory rewind that main already contains is preserved exactly as it is; the new generalized path only engages for non-GLM sessions.

A fourth change — a smaller first chunk for pipelined distributed prefill — was in the original version of this pull request and has been removed. Measurement prompted by review showed that the smaller first chunk shortens the pipeline fill, but the extra chunk boundary cancels the gain. With 4096-token chunks on a prompt of about 12,000 tokens, fill latency dropped from 30.8 s to 19.2 s while the total moved only from 60.9 s to 60.6 s; with 1024-token chunks the totals were 49.5 s against 49.7 s.

## Correctness

Rebased on current main (`84cc882`), Radeon 8060S (gfx1151), TheRock toolchain:

- Two porting notes. First, main's GLM rewind is kept verbatim and the new path is restricted to non-GLM sessions. Second, two string comparisons against `cache_source` literals became booleans captured at assignment time; the semantics are identical, because later assignments to `cache_source` only happen on paths where no cache was hit, and the new form also keeps GCC's `-Wstring-compare` warning quiet.
- `./ds4_test --server` passes, including the new `kv_rewind_reuse_target()` decision-table tests.
- End to end on one machine, with the released Q4_K-experts model streamed from SSD: repeating an identical greedy request logs `live kv cache rewind-hit live=154 prompt=114 common=114 reused=113`, the prompt phase drops from 55.8 s to 0.63 s, and the response is byte-identical to the cold run.
- After removing the first-chunk change, a two-machine distributed run confirmed the uniform chunk schedule behaves as before (24.7 tokens/s prefill, 13.8 tokens/s decode).

## Numerics

A generation after a rewind is a valid greedy continuation — the token before the prompt end is re-evaluated to restore the logits — but it is not guaranteed to be bit-identical to a full re-prefill in the rare case where two logits tie exactly. `DS4_KV_REWIND_REUSE=0` restores the previous behavior.

### DIFF
--- a/ds4_distributed.c
+++ b/ds4_distributed.c
@@ -3287,7 +3287,7 @@ static void dist_prefill_reader_emit_progress(
 
     while (*reported < completed) {
         (*reported)++;
-        uint32_t rel = (*reported) * reader->chunk_cap;
+        uint32_t rel = *reported * reader->chunk_cap;
         if (rel > reader->total_tokens) rel = reader->total_tokens;
         uint32_t current = reader->progress_base + rel;
         if (current > reader->progress_total) current = reader->progress_total;
@@ -3593,7 +3593,7 @@ static int dist_coordinator_prefill_prompt_pipelined(
     uint64_t chunk_prefix_hash = dist_token_hash_prefix(prompt->v, span_start);
     for (uint32_t i = 0, hash_pos = span_start; i < chunk_count; i++) {
         const uint32_t remaining = span_end - hash_pos;
-        const uint32_t chunk = remaining < chunk_cap ? remaining : chunk_cap;
+        uint32_t chunk = remaining < chunk_cap ? remaining : chunk_cap;
         chunk_prefix_hash = dist_token_hash_update_span(chunk_prefix_hash,
                                                         prompt->v + hash_pos,
                                                         chunk);
@@ -3650,7 +3650,7 @@ static int dist_coordinator_prefill_prompt_pipelined(
             break;
         }
         const uint32_t remaining = span_end - pos;
-        const uint32_t chunk = remaining < chunk_cap ? remaining : chunk_cap;
+        uint32_t chunk = remaining < chunk_cap ? remaining : chunk_cap;
         const uint64_t hidden_bytes64 = (uint64_t)chunk * hc_values * sizeof(float);
         const uint32_t hidden_bytes = (uint32_t)hidden_bytes64;
         ds4_dist_prefill_send_slot *slot =
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -10644,9 +10644,13 @@ static void server_progress_cb(void *ud, const char *event, int current, int tot
     if (display_current > display_total) display_current = display_total;
     double pct = display_total > 0 ? 100.0 * (double)display_current / (double)display_total : 100.0;
     double avg_tps = elapsed > 0.0 ? (double)display_current / elapsed : 0.0;
-    int interval_tokens = p->seen ? current - p->last_current : 0;
+    /* First callback: the interval is everything since prefill start, not
+     * zero — otherwise the first chunk always logs a bogus 0.00 t/s (most
+     * visible in distributed mode, where the first callback covers a whole
+     * pipeline chunk). */
+    int interval_tokens = p->seen ? current - p->last_current : display_current;
     if (interval_tokens < 0) interval_tokens = 0;
-    double interval_s = p->seen ? now - p->last_t : 0.0;
+    double interval_s = p->seen ? now - p->last_t : elapsed;
     double chunk_tps = interval_s > 0.0 ? (double)interval_tokens / interval_s : 0.0;
     p->last_current = current;
     p->last_t = now;
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -11172,6 +11172,30 @@ static uint64_t server_next_sequence(server *s) {
     return seq;
 }
 
+/* Rewind-reuse decision for the live KV cache.  Returns the session position
+ * to rewind to, or -1 when the request must take the cold/disk path.
+ *
+ * The classic continuation case (live state is a prefix of the prompt) is
+ * handled by the caller and never reaches here.  Two miss shapes are
+ * recoverable:
+ *   - prompt is a strict prefix of live (exact repeat, regenerate, branch
+ *     from live): rewind one token before the prompt frontier so the sync
+ *     re-evals a single token, restoring frontier logits and leaving the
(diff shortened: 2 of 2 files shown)


jessa0:
hmm what exactly is the goal of the first-chunk pipeline ramp? it seems like you'd still end up waiting the same amount of time for the first and second chunk combined as it would without this change. i tested it, and it seems like this is the case (though the logs are confusing to look at, since logging the first chunk in distributed mode has been broken for a while):

```
Aug 16 20:28:36 localhost ds4:  ds4: distributed coordinator: pipelined prefill 4 chunks of up to 4096 tokens through 1 worker, first hop <redacted>, send depth 2, flow window 3
Aug 16 20:28:45 localhost ds4:  ds4: distributed telemetry: request=392 hop=0 layers=20:42 route=0 pos=17462 tokens=1024 eval=4800.765ms downstream_wait=0.000ms forward_send=0.000ms input=64.00MiB output=0.00MiB
Aug 16 20:28:54 localhost ds4:  0816 20:28:54 ds4-server: chat ctx=17462..27803:10341 TOOLS prefill chunk 1024/10341 (9.9%) chunk=0.00 t/s avg=58.85 t/s 17.400s
Aug 16 20:29:09 localhost ds4:  ds4: distributed telemetry: request=393 hop=0 layers=20:42 route=0 pos=18486 tokens=4096 eval=15391.491ms downstream_wait=0.000ms forward_send=0.000ms input=256.02MiB output=0.00MiB
Aug 16 20:29:12 localhost ds4:  0816 20:29:12 ds4-server: chat ctx=17462..27803:10341 TOOLS prefill chunk 5120/10341 (49.5%) chunk=227.12 t/s avg=144.49 t/s 35.434s
Aug 16 20:29:25 localhost ds4:  ds4: distributed telemetry: request=394 hop=0 layers=20:42 route=0 pos=22582 tokens=4096 eval=15801.417ms downstream_wait=0.000ms forward_send=0.000ms input=256.02MiB output=0.00MiB
Aug 16 20:29:25 localhost ds4:  0816 20:29:25 ds4-server: chat ctx=17462..27803:10341 TOOLS prefill chunk 9216/10341 (89.1%) chunk=306.07 t/s avg=188.79 t/s 48.817s
Aug 16 20:29:30 localhost ds4:  ds4: distributed telemetry: request=395 hop=0 layers=20:42 route=0 pos=26678 tokens=1125 eval=5221.465ms downstream_wait=0.000ms forward_send=0.000ms input=70.32MiB output=0.49MiB
Aug 16 20:29:30 localhost ds4:  0816 20:29:30 ds4-server: chat ctx=17462..27803:10341 TOOLS prefill chunk 10341/10341 (100.0%) chunk=214.94 t/s avg=191.32 t/s 54.051s
Aug 16 20:29:30 localhost ds4:  ds4: distributed coordinator: pipelined prefill done tokens=10341 chunks=4 total=54.051s 191.32 t/s local=35.434s send=0.533s 1211.68 MiB/s
Aug 16 20:29:30 localhost ds4:  0816 20:29:30 ds4-server: chat ctx=17462..27803:10341 TOOLS prompt done 54.084s
```

there's 5 seconds from 28:36 to 28:41 where the second machine is waiting (28:45 minus the eval time of 4.8s), and then 10 seconds from 28:45 to 28:54 (29:09 minus eval time of 15.4s) where it's waiting again. from then on, its fully utilized. but the total wait there is 15s, which is the same as the processing time of a single 4k chunk.

jyatesdotdev:
I'll pull some more testing data and post a benchmark tomorrow. Honestly speaking, this just a subset of things that _seemed_ to improve performance, and I pulled it out of a larger set of changes. My initial PR was massive but I was able to get 200+ prefill and 15ish decode, but might need to do some more work to actually extract the useful bits.

jyatesdotdev:
Confirmed on our pair — you're right. Isolated A/B, unique ~11.9k prompts, two-node gfx1151:

| config | fill (first chunk logged) | total prompt |
|---|---:|---:|
| 4096 steady, no ramp | 30.8 s | 60.9 s |
| 4096 + 1024 ramp | 19.2 s | 60.6 s |
| 1024 steady, no ramp | 7.6 s | 49.5 s |
| 1024 + 256 ramp | — | 49.7 s |

The ramp cuts the fill by 11.6 s and the extra chunk boundary eats almost all of it — total within noise in both regimes. It's out of the PR; what remains is the rewind-reuse and the timing instrumentation (which is what produced the phase numbers above).

Also fixed the first-chunk logging you flagged — the first progress callback hard-zeroed the interval (`interval_tokens`/`interval_s` gated on `p->seen` in the prefill progress logger), so the first chunk always printed `0.00 t/s`. It now measures the first interval from prefill start; that's the second commit.

Side note from the same runs: the 1024-chunk/window-3 config beats 4096 by ~23% end-to-end on this prompt (60.9 → 49.5 s) — that was the actual prefill lever in our tuning, not the ramp.

jyatesdotdev:
Closing this pull request to cut down the number of open submissions from me. The live key-value rewind reuse change is being folded into a single performance branch rebased on current main, together with my other work, and will return as one combined submission with fresh measurements against the current tree.

PR #819, opened 2026-08-16
ROCm: port decode-island graph capture to HIP (staged, default-off)

jyatesdotdev:
## What this does

Ports decode-island graph capture to HIP for the ROCm backend. The feature is disabled by default.

A decode island is a group of kernels in the decode layer whose launch geometry and buffer bindings do not depend on the token position. Such a group can be captured once as a HIP graph and replayed on later tokens, removing the per-launch CPU overhead. The capture interface (`ds4_gpu_decode_graph_*`) already exists in the tree and is implemented for CUDA; on ROCm it was stubbed out. This pull request implements it:

- Islands are captured once per buffer-alias key and replayed afterwards. A replayed graph re-runs the recorded kernels exactly, so replay output is byte-identical to the eager execution it recorded. Any capture failure retires the entry, and the island runs eagerly from then on — also byte-identical.
- Every kernel launch in the ROCm backend now targets `ds4_rocm_stream()`, which is the capture stream while capture is armed and the default stream otherwise. Behavior without capture is unchanged.
- Enable with `DS4_ROCM_DECODE_GRAPHS=span` (islands two and up) or `DS4_ROCM_DECODE_GRAPHS=all` (also the eval islands). Single-GPU processes only.

Why it ships disabled: on gfx1151 today, decode time is dominated by other stages, and capture measures as neutral — 13.42 tokens/s eager against 13.37 with graphs enabled. The implementation is included anyway because the per-island launch chains become worth capturing once they are batched across rows (the multi-row decode work, tracked separately), and because the port is complete, tested, and inert when off.

One integration point is deliberately absent: capture of speculative-span stages belongs with the multi-row decode work and will arrive with it. This branch does not modify `ds4.c`.

## Correctness

Rebased on current main (`84cc882`), Radeon 8060S (gfx1151), TheRock toolchain:

- Launch-site audit: every kernel launch in the ROCm backend goes through `ds4_rocm_stream()`, except the three launches that intentionally use the dedicated shared-expert stream, which are unchanged.
- End to end on two machines with greedy decoding and the released Q4_K-experts model file:
  - with the variable unset: output byte-identical to unmodified main;
  - with `DS4_ROCM_DECODE_GRAPHS=all`: output byte-identical to eager execution;
  - throughput within noise of eager, as stated above.

## How to test

`make strix-halo`, then any ROCm decode run with and without `DS4_ROCM_DECODE_GRAPHS=all`. Outputs must match exactly.

### DIFF
--- a/ds4_rocm_compat.cu
+++ b/ds4_rocm_compat.cu
@@ -16,12 +16,244 @@ static int rocm_tier_valid(int tier) {
     return tier == 0 && g_n_gpus == 1;
 }
 
-/* Decode-island graph capture is CUDA-only for now; ROCm decodes eagerly. */
-extern "C" int ds4_gpu_decode_graphs_supported(void) { return 0; }
-extern "C" int ds4_gpu_decode_graph_begin(const ds4_decode_graph_key *key) { (void)key; return -1; }
-extern "C" int ds4_gpu_decode_graph_end(const ds4_decode_graph_key *key) { (void)key; return -1; }
-extern "C" void ds4_gpu_decode_graph_abort(const ds4_decode_graph_key *key) { (void)key; }
-extern "C" void ds4_gpu_decode_graphs_invalidate(void) {}
+/* ------------------------------------------------------------------------
+ * Decode-island HIP graph capture.  Port of the CUDA implementation in
+ * ds4_cuda.cu (design from the Entrpi/ds4 batched-serving fork): the
+ * position-independent islands of a decode layer are captured once per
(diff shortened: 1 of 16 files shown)


jyatesdotdev:
Withdrawing this pull request. On gfx1151 the capture path measures as neutral (13.42 tokens/s eager against 13.37 with graphs enabled), so today it is inert machinery behind a default-off switch. Its stated value depended on batched multi-row launch chains arriving in later work; that work is still a design decision away, and if it lands, this port can be revived then. Until that point this is review surface without a performance case. The remaining pull requests in the series are unaffected.

PR #820, opened 2026-08-17
distributed: descriptor-framed NHI transport, ROCm event timing, regression gates

jyatesdotdev:
## What this does

Extends the distributed layer with a second bulk transport, GPU event timing, and the tests that qualify both.

- **Protocol v3 links.** Link negotiation is versioned and fails closed: peers agree on either descriptor-framed TCP or NHI (the USB4/Thunderbolt native host interface), with generation-scoped bulk descriptors, an ACK/READY activation barrier, and single-link lifecycle checks. `--dist-transport auto` selects descriptor-framed TCP. `--dist-transport nhi --dist-nhi-device PATH` selects the NHI backend. `--dist-transport tcp` keeps the legacy protocol.
- **NHI transport.** Bulk payloads move over the Thunderbolt link through a zero-copy stream device with staged CPU copies. A separate mapped-lease mode (`DS4_DIST_NHI_MAPPED=1` on both peers) hands contiguous payloads directly to the registered driver slot, with a HIP fence on every ownership transfer.
- **Multi-token-prediction support** in the distributed transport and session layer.
- **ROCm event timing** (`ds4_rocm_timing`): low-overhead GPU event timing with a host-testable fake backend, plus prequantized decode controls.
- **Tests:** `tests/test_dist_v3` (87 checks), `tests/test_rocm_timing_fake` (183 checks), `tests/test_gpu_timing_stubs` (15 checks), `tests/test_transport`, and two opt-in live tests that need two hosts (`make test-nhi-live`, `make test-nhi-live-rocm`).
- **Documentation:** `RDMA.md` describes the transport design and the hardware limits behind it; README and STRIXHALO are updated.

## Performance

Measured on two Radeon 8060S machines over Thunderbolt, with a pipelined prefill of about 7,500 tokens and 32-bit activations. Full tables and method are in the benchmark comment below.

| configuration | bulk send throughput |
|---|---:|
| legacy TCP | 1062 MiB/s |
| descriptor-framed TCP (default with `auto`) | 1191 MiB/s |
| NHI, CPU-copy mode | 1192 MiB/s |
| NHI, mapped-lease mode | 131 MiB/s |

Descriptor framing is 12% faster than the legacy protocol. The NHI CPU-copy path matches descriptor-framed TCP exactly, which is why it remains a lab option behind an explicit device flag, as `RDMA.md` states. The mapped-lease mode is nine times slower because of the per-transfer fences; it exists for hardware qualification and defaults off. At these payload sizes, transport is about half a percent of total prefill time — the value of the faster framing is in per-token transfers and larger activation shapes, not in the prefill total.

## Correctness

Rebased on current main (`84cc882`):

- Warning-free `make strix-halo`. Built with `-fPIC` in `CFLAGS`; main does not yet carry the `ROCM_HOST_CFLAGS` fix, which ships separately in #656. The build rule already references `ROCM_HOST_CFLAGS`, which stays empty until that lands.
- `make test-dist-v3 test-gpu-timing test-rocm-timing`: 87 of 87, 183 of 183, and 15 of 15 checks pass.
- End to end on two machines with greedy decoding and the released Q4_K-experts model file, coordinator running layers 0–21 and worker running 22–output: output byte-identical to unmodified main's distributed run at the same settings, throughput at parity (24.7 tokens/s prefill, 13.4 tokens/s decode).

## Scope

Left out to keep this reviewable: the multi-row decode kernels (#821), batched speculative verification, decode staggering, HIP graph capture (#819), and the MXFP4 kernels (#656). One note for whoever merges both this and #656: a follow-up adds `n_expert` and `atomic_out` parameters to the MXFP4 down-projection kernel for batched verification; that change is deliberately absent here because main has no MXFP4 code.

## How to test

`make strix-halo && make test-dist-v3 test-gpu-timing test-rocm-timing`. The live NHI tests need two hosts; see `RDMA.md`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,8 +28,17 @@
 /tests/test_mxfp4_cuda
 /tests/test_mxfp4_dot
 /tests/test_mxfp4_metal
+/tests/test_mxfp4_rocm
 /tests/test_q4k_dot
+/tests/test_rocm_qkv_fusion
+/tests/test_rocm_timing_fake
+/tests/test_gpu_timing_stubs
 /tests/test_sampling
+/tests/test_transport
+/tests/test_dist_v3
+/tests/test_distributed_transport
+/tests/test_nhi_live
+/tests/test_nhi_live_rocm
 *.o
 *.dSYM/
 __pycache__/
@@ -37,3 +46,4 @@ __pycache__/
 /misc/
 .*.swp
 .DS_Store
+/.idea/
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,20 @@
 CC ?= cc
+CXX ?= c++
 UNAME_S := $(shell uname -s)
 
 ifeq ($(UNAME_S),Darwin)
 NATIVE_CPU_FLAG ?= -mcpu=native
 SAMPLING_TEST :=
+METAL_SYNTH_TEST := tests/test_mxfp4_metal
 else
 NATIVE_CPU_FLAG ?= -march=native
 SAMPLING_TEST := tests/test_sampling
+METAL_SYNTH_TEST :=
 endif
 
 DEBUG_FLAGS ?= -g
 CFLAGS ?= -O3 -ffast-math $(DEBUG_FLAGS) $(NATIVE_CPU_FLAG) -Wall -Wextra -std=c99
+CXXFLAGS ?= -O2 $(DEBUG_FLAGS) $(NATIVE_CPU_FLAG) -Wall -Wextra -std=c++11
 OBJCFLAGS ?= -O3 -ffast-math $(DEBUG_FLAGS) $(NATIVE_CPU_FLAG) -Wall -Wextra -fobjc-arc
 QUALITY_CFLAGS ?= -O3 $(DEBUG_FLAGS) $(NATIVE_CPU_FLAG) -Wall -Wextra -std=c11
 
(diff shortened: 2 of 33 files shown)


jyatesdotdev:
Benchmark data for the transport layer, measured on the two-node gfx1151 pair (Thunderbolt link, f32 activations, Q4_K-experts Flash GGUF, coordinator layers 0:21 / worker 22:output, `--debug` telemetry, ~7.5k-token pipelined prefill, 2×4096-token chunks):

| transport config | total | prefill t/s | coord local | bulk send | send throughput |
|---|---:|---:|---:|---:|---:|
| legacy v2 TCP (`--dist-transport tcp`) | 83.69 s | 89.44 | 55.56 s | 0.441 s | **1061.71 MiB/s** |
| v3 TCP (`--dist-transport auto`, production) | 83.50 s | 89.64 | 55.74 s | 0.393 s | **1191.44 MiB/s** |
| v3 NHI cpu-copy (`--dist-transport nhi --dist-nhi-device …`) | 83.62 s | 89.52 | 55.65 s | 0.392 s | **1192.03 MiB/s** |
| v3 NHI mapped leases (`DS4_DIST_NHI_MAPPED=1` both peers) | 83.64–83.88 s | 89.2–89.5 | 55.6–55.8 s | 3.57–3.60 s | **130–131 MiB/s** |

Read-out:

- Protocol v3's descriptor framing is **+12% bulk throughput** over legacy v2 TCP (1191 vs 1062 MiB/s) on identical work.
- NHI cpu-copy negotiates (`transport=nhi-cpu-copy` confirmed in both logs) and lands at exact parity with v3 TCP (1192 vs 1191 MiB/s) — consistent with `RDMA.md`'s status note: the NHI backend is correct and qualified, but does not beat descriptor-framed TCP on full-model throughput. It stays lab-only (`--dist-nhi-device`).
- NHI mapped leases are **9× slower** (131 MiB/s; reproduced across two runs) — the per-transfer HIP synchronization fences dominate. That's why mapped mode is explicit qualification only and `DS4_DIST_NHI_MAPPED` defaults off.
- Transport is ~0.5% of end-to-end prefill at these sizes (0.39 s of 83.5 s); layer eval dominates. The v3 gains matter for activation-heavy shapes and latency of small per-token transfers, not for the bulk prefill total on this pair.

Correctness held in every config: identical requests completed, outputs coherent; regression gates (`make test-dist-v3 test-gpu-timing test-rocm-timing`) pass 87/87, 183, 15 as reported in the PR body.

jyatesdotdev:
Withdrawing this pull request. The transport measurements (posted above) made the case against it: on the target hardware the NHI CPU-copy path is exactly at parity with descriptor-framed TCP (1192 vs 1191 MiB/s) and the mapped-lease mode is nine times slower (131 MiB/s), while bulk transport is about half a percent of prefill time — so none of this moves end-to-end numbers on Thunderbolt-connected gfx1151 pairs. What remains is review surface without a performance case.

The qualification results and the hardware analysis stay available in `RDMA.md` in this branch for anyone who wants to revisit the transport on different hardware. The remaining pull requests in this series are unaffected.

PR #821, opened 2026-08-17
ROCm: exact-rows k-row kernel tier for batched decode (kernels + exactness gate)

jyatesdotdev:
## What this does

Adds a family of ROCm Q8_0 kernels that evaluate several rows in one call — two to five rows — while keeping each row's reduction in exactly the same order as the existing one-row kernel. Every row of a batched call is therefore bit-identical to a one-row call on the same data. That property is what makes these kernels usable for exact speculative verification, where verified logits must match what single-token decode would have produced. It is the same invariant every other decode kernel in this backend already holds.

- `rocm/ds4_rocm_q8.cuh`: the k-row kernels — two to five rows per block, one warp per row, one-row reduction order.
- `rocm/ds4_rocm_matmul.cuh`, `rocm/ds4_rocm_shared_expert.cuh`: k-row entry points.
- `ds4_rocm_compat.cu`: C-callable wrappers.
- `tests/test_q8_krow_rocm.c`: exactness test. Every row of every batched call is compared byte for byte with the one-row reference, across several shapes.
- `tests/bench_q8_krow_rocm.c`: throughput benchmark comparing one batched call against the same rows evaluated sequentially.
- `make test-q8-krow-rocm` runs the exactness test.

## Performance

On Radeon 8060S (gfx1151), best of three runs of 200 iterations each; every cell re-verified bit-identical during the run. One batched call against k sequential one-row calls, on 4096×4096 Q8_0 weights:

| rows | batched | sequential | speedup |
|---:|---:|---:|---:|
| 2 | 0.032 ms | 0.089 ms | 2.8× |
| 3 | 0.029 ms | 0.135 ms | 4.7× |
| 4 | 0.032 ms | 0.179 ms | 5.5× |
| 5 | 0.036 ms | 0.222 ms | 6.3× |
| 6 | 0.228 ms | 0.267 ms | 1.2× |

The other measured shapes (2048×4096 and 4096×14336) show the same pattern: two to six times faster at two to five rows. At six rows and beyond, the dispatch falls back to per-row evaluation — exactness is preserved and the speedup drops to about one — so callers should size batches to five rows or fewer. Speculative spans of two to three rows sit in the best region. Full tables are in the benchmark comment below.

## Scope

This pull request contains the kernel tier and its tests. Nothing calls the new kernels yet, and no existing path changes.

The host-side driver that uses them — batched evaluation of speculative spans — is not included, for a concrete reason. After this work was forked, main independently reworked the decode-phase evaluator for its own speculative decoding (the `quad_store` logic in `metal_graph_encode_decode_layer_phase`). The driver was built against the pre-rework evaluator, and the two designs solve the same problem with incompatible control flow. Porting the driver means choosing one of the two designs, and that should be a maintainer decision, not a side effect of this pull request. The driver runs in production on our branch, where spans four rows deep verify with no loss of acceptance; it will follow as its own pull request once the evaluator direction is agreed.

## Correctness

Rebased on current main (`84cc882`), TheRock toolchain:

- `make test-q8-krow-rocm` passes; all shapes bit-exact.
- Additive only: no existing launch site or host path changes.

## How to test

`make strix-halo && make test-q8-krow-rocm`. The benchmark builds the same way as the test and runs standalone.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ DS4_LINK_LIBS ?= $(CUDA_LDLIBS)
 METAL_LDLIBS := $(LDLIBS)
 endif
 
-.PHONY: all help clean test test-metal-session-batch test-mxfp4-cuda test-cuda-session-batch test-cuda-mixed-batch dspark-acceptance dspark-verify-depth mtp-verify-depth cpu cuda cuda-spark cuda-generic cuda-regression strix-halo rocm
+.PHONY: all help clean test test-metal-session-batch test-mxfp4-cuda test-cuda-session-batch test-cuda-mixed-batch dspark-acceptance dspark-verify-depth mtp-verify-depth cpu cuda cuda-spark cuda-generic cuda-regression strix-halo rocm test-q8-krow-rocm
 
 ifeq ($(UNAME_S),Darwin)
 .PHONY: metal-decode-schedule-bench metal-prefill-variant-bench check-mxfp4-half-lut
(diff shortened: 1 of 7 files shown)


jyatesdotdev:
Benchmark data for the kernel tier (gfx1151, TheRock, best-of-3 × 200 iters; new `tests/bench_q8_krow_rocm.c`, added to this PR). Comparison: one batched k-row call (`ds4_gpu_matmul_q8_0_decode_rows_exact_tensor`) vs k sequential one-row calls (`ds4_gpu_matmul_q8_0_tensor`, n_tok=1) — i.e., what span evaluation costs without the tier. Every cell re-verified bit-identical between the two paths.

**4096×4096 q8_0:**

| k | k-row batched | k× one-row | speedup |
|---:|---:|---:|---:|
| 2 | 0.0316 ms | 0.0893 ms | **2.83×** |
| 3 | 0.0289 ms | 0.1346 ms | **4.66×** |
| 4 | 0.0324 ms | 0.1787 ms | **5.51×** |
| 5 | 0.0355 ms | 0.2223 ms | **6.25×** |
| 6 | 0.2276 ms | 0.2668 ms | 1.17× |

**2048×4096:** k=2 3.31×, k=3 5.16×, k=4 6.40×, k=5 2.61×, k=6 1.54× — all bit-exact.

**4096×14336 (wide FFN):** k=2 2.08×, k=3 3.33×, k=4 4.24×, k=5 5.26×, k=6 1.03× — all bit-exact.

Read-out:

- The kernel tiers (k=2..5) give **2.1–6.4×** over sequential one-row evals while staying bit-identical — that's the exact-rows contract the speculative verifier needs, at batched-span cost instead of per-row cost.
- k=6 shows the deliberate fallback boundary: beyond the kernel tiers the dispatch reverts to the per-row path (exactness preserved, speedup ~1×). Span drivers should size to ≤5 rows; MTP draft-2 spans (2–3 rows) sit in the best region.
- Absolute times are microbenchmarks of the matvec alone; end-to-end decode impact rides with the host-side span driver (see the PR's scope note).

PR #827, opened 2026-08-18
server: idle prefill quantum follows the engine's prefill cap

adriangalilea:
With no generation active, `server_session_sync` chops every prefill into 2048-token quanta (`server_prefill_quantum_for`), while the engine's own prefill cap is 4096 for Flash and 8192 for PRO (`ds4_prefill_cap_for_prompt`). The quantum exists so the model executor is released between chunks and a decode can enter; halving the engine's cap doesn't improve that bound, it just doubles the fixed per-chunk cost — each chunk pays the full MoE expert scan regardless of size. This makes the idle quantum follow the engine's resolved cap (new accessor `ds4_engine_prefill_quantum`, which resolves the auto value; `ds4_engine_prefill_chunk` returns 0 when no explicit `--prefill-chunk` was given), flooring at the historical 2048 when no engine is attached (unit-test structs). `--mixed-prefill-quantum` while decoding is untouched.

Measured on Apple M3 Ultra (60-core GPU) 512 GB, DeepSeek V4 Flash 0731 native-MXFP4 GGUF (156 GB), base 84cc882.

Engine side — a 16,384-token cold prefill in one `ds4_session_sync`, by chunk size (`./ds4-bench --prompt-file speed-bench/promessi_sposi.txt --ctx-start 16384 --ctx-max 16384 --gen-tokens 0 --prefill-chunk N`):

| chunk | prefill t/s |
|---|---|
| 1024 | 482 |
| 2048 | 574 |
| 4096 | 590 |
| 8192 | 596 |

Served path — two `ds4-server --ctx 1048576 --batched-session 4` instances with fresh `--kv-disk-dir`, same request suite (a ~7.9k-token chat prompt cold, then extensions), same boot, back to back:

| | 2048 quantum (main) | engine cap (this PR) |
|---|---|---|
| cold prefill | 443 t/s / 18.2 s | 461 t/s / 17.7 s |
| warm chat extension | 9.94 s | 9.38 s |
| decode | 43.4 t/s | 43.3 t/s |

The served path also fragments: a session restored from a disk anchor at an unaligned position steps in `base + k*quantum` chunks, and the partial chunks (526 and 1249 tokens in a live log) run at 280-385 t/s against ~560 for full ones; larger quanta halve the fragment count.

Correctness: full-vocab logits at frontiers 2048/4096/6144/8192 are byte-identical to main at ds4-bench's default chunk (`--dump-frontier-logits-dir` byte-compare). Note for downstream users: greedy outputs of *long served prompts* can shift once, since the served chunk size is part of the numeric layout — same class as changing `--prefill-chunk`. `./ds4_test --server` green (the quantum unit test keeps its engine==NULL floor assertions); built with zero warnings. Running in production on the machine above since 2026-08-18.


adriangalilea:
Rebased onto current main (`a04f46f`). The only conflict was `server_prefill_quantum_for` gaining the GLM 5.3 floor in the generation-active branch; that branch is kept verbatim, this PR still only changes the idle branch (engine prefill cap instead of the hard-coded 2048). Builds clean.

PR #828, opened 2026-08-18
metal: non-blocking GPU stage timestamps behind DS4_METAL_GPU_STAGE_TIMESTAMPS

adriangalilea:
The stage profilers (`DS4_METAL_LAYER_STAGE_PROFILE`, `DS4_METAL_DECODE_STAGE_PROFILE` and the flash/MoE siblings) attribute time by ending the command buffer and waiting on the CPU at every stage boundary — fine for coarse questions, but each stage pays a CPU round trip, so the printed numbers don't add up to the unprofiled total and can't build a stage budget.

`DS4_METAL_GPU_STAGE_TIMESTAMPS=1` reuses the same stage boundaries but only *commits* the batch in flight and tags it (`ds4_gpu_stage_flush`); nothing waits. Once the layer/token completes (the existing `ds4_gpu_end_commands` already waits for every pending buffer), the report reads each tagged buffer's `GPUStartTime`/`GPUEndTime` and prints per (part, stage) GPU-busy ms, plus a total line with busy, span (first start → last end) and the gap between them — the scheduling overhead the run actually paid. `DS4_METAL_GPU_STAGE_TIMESTAMPS_DETAIL=1` adds one line per buffer. Diagnostic switch, off by default, validates the one release path; other backends flush and record nothing.

    ds4: gpu stage times what=prefill pos=0 tokens=2048 part=ffn stage=routed_moe buffers=43 gpu_ms=1375.896
    ds4: gpu stage times what=prefill pos=0 tokens=2048 total gpu_busy_ms=... span_ms=... gap_ms=... buffers=473 incomplete=0

Measured on Apple M3 Ultra (60-core GPU) 512 GB, DeepSeek V4 Flash 0731 native-MXFP4 GGUF, base 84cc882 — 16k cold prefill at `--prefill-chunk 2048`, first chunk, ms summed over the 43 layers, against the blocking profiler:

| stage | blocking profiler | GPU timestamps |
|---|---|---|
| ffn/routed_moe | 1393 | 1376 |
| attn/output_proj | 536 | 525 |
| attn/q_path | 352 | 335 |
| attn/attention | 299 | 281 |

i.e. the blocking numbers carry ~5-7% of sync overhead; the timestamp numbers are GPU busy time on the GPU clock and sum to the chunk total. This is what let me localize the long-context prefill decay to the attention stage alone (299 → 1098 ms/chunk from pos 0 to pos 14k while every other stage stays flat).

Exactness: full-vocab logits at frontiers 2048/4096/6144/8192 byte-identical to a build without the change (`ds4-bench --dump-frontier-logits-dir` byte-compare) — and the switch is off by default anyway. Built with zero warnings; model-free `make test` targets pass.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -27972,6 +27972,27 @@ static bool metal_graph_stage_profile_enabled_for_layer(
     return metal_graph_profile_layer_value_match(layer_env, il);
 }
 
+/* DS4_METAL_GPU_STAGE_TIMESTAMPS: the same stage boundaries as the layer/decode
+ * stage profilers, but each boundary only commits the batch and tags it (see
+ * ds4_gpu_stage_flush); nothing waits.  Reported per prefill chunk / decode
+ * token by ds4_gpu_stage_report. */
+static bool metal_graph_gpu_stage_timestamps(void) {
+    static int enabled = -1;
+    if (enabled < 0) {
+        const char *v = getenv("DS4_METAL_GPU_STAGE_TIMESTAMPS");
+        enabled = v != NULL && strcmp(v, "0") != 0;
+    }
+    return enabled != 0;
+}
+
+static bool metal_graph_gpu_stage_timestamps_layer(uint32_t il) {
+    return metal_graph_gpu_stage_timestamps() &&
+           metal_graph_stage_profile_enabled_for_layer(
+                "DS4_METAL_GPU_STAGE_TIMESTAMPS",
+                "DS4_METAL_GPU_STAGE_TIMESTAMPS_LAYER",
+                il);
+}
+
 static bool metal_graph_layer_stage_profile_enabled(uint32_t il) {
     return metal_graph_stage_profile_enabled_for_layer(
                 "DS4_ROCM_LAYER_STAGE_PROFILE",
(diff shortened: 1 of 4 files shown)


adriangalilea:
Extended to the DeepSeek V4.1 decode graph (`ds41_graph_step`), which is a separate graph from `metal_graph_*` and so printed nothing before. Same non-blocking mechanism; the tags are `part=v41` with `stage=` pre_attn / attn_project / attn_kv / attn_select / attn_gather / attn_heads / attn_out / post_attn / moe_route / moe_shared / moe_routed / after_moe / logits, reported per token. Off, the command stream is unchanged.

What it showed on a Mac Studio M3 Ultra with the Q4 model resident (2k prompt, mean of 40 decode tokens, GPU busy ms per token): moe_routed 8.6, attn_out 4.8, attn_project 4.3, moe_shared 3.1, pre_attn 2.8, moe_route 2.7, post_attn 2.1, attn_heads 2.1, attn_select 1.0, logits 0.9, attn_kv 0.7, after_moe 0.6, attn_gather 0.5 — 36.9 ms busy in a 37.7 ms span. The same instrument with the per-layer drains still in place reported 37.3 ms busy in a 46.9 ms span, i.e. 9.6 ms of scheduling gap per token, which is what led to the queue change in the V4.1 decode PR. Branch rebased on current main.

PR #831, opened 2026-08-18
metal: register-blocked, K-register-resident indexer prefill scorer (tiled4+tiled5)

adriangalilea:
Stacks on #830 (its commit is the base of this branch); independent of every other open PR.

The tiled2 scorer (#830) feeds each 8x8 simdgroup MMA with two fresh threadgroup loads — one Q tile, one K tile, zero register reuse. On an M3 Ultra it runs at ~14 TFLOPS while this engine's register-blocked dense GEMMs reach 19-23; the gap is register blocking, and the CUDA backend's own history agrees (08fecd9 "cuda: register-block the prefill indexer scorer" preceded the SM121 mxf4 work). Two steps, one reuse ladder:

- **tiled4** — the token tile doubles to TM=16: each simdgroup keeps two accumulators and shares every staged K tile between them. Threadgroup loads per MMA drop from 2.0 to 1.5.
- **tiled5** — the staged K tile is invariant across the head loop, yet tiled2/tiled4 reload it from threadgroup memory for every head. Each simdgroup loads its sixteen 8x8 K tiles into registers once, before the head loop; per head only the two Q tiles are loaded. Loads per MMA reach 1.0. The ~16 live matrix registers per simdgroup cost nothing measurable — Apple9 allocates registers as a cache.

Staging, the two-barrier-per-head structure and the per-pair reduction are exactly tiled2's — sixteen 8x8 MACs per head in db order, relu(dot)*w in float, heads ascending — so outputs are bit-identical. One non-obvious constraint keeps them so: the output cells per thread must be spelled out as scalar accumulators. A loop over an accumulator array lets the compiler contract each cell's relu(dot)*w*scale term differently from tiled2's scalar form under fast math (~1 ULP per head, compounding over the 64-head chain into logit drift; unroll/vectorize pragmas and explicit fma() do not pin it — the hand-unrolled scalar shape does).

tiled5 is the default; `DS4_METAL_DISABLE_INDEXER_SCORES_TILED5` restores tiled2, `DS4_METAL_INDEXER_SCORES_TILED4` selects the intermediate for A/B. `tests/test_indexer_tiled4.c` is a standalone no-model harness: it runs the scorer twice (env off/on) on engine-shaped random data up to a 64k frontier chunk (edge tiles, causal cut, resumed-anchor shapes) and byte-compares.

Measured, M3 Ultra 512 GB / DeepSeek V4 Flash 0731 MXFP4 (156 GB GGUF), macOS 26.2:

- `metal_prefill_variant_bench --candidate-env DS4_METAL_DISABLE_INDEXER_SCORES_TILED5 --prefix-tokens 8192 --repeats 2`: bit-exact 8/8 runs.
- Full-vocab frontier logits across the whole sweep below: byte-identical at every step through 64k.
- Decode untouched (the tiled path only runs at n_tokens >= 32); decode bench bit-exact.
- Full CONTRIBUTING sweep (2k -> 64k step 2k, gen 128, ABBA order, medians of 4 runs), tiled5 vs tiled2:

| ctx | tiled2 t/s | tiled5 t/s | Δ |
|---|---|---|---|
| 2048 | 665.8 | 666.9 | +0.17% |
| 8192 | 574.9 | 585.0 | +1.71% |
| 16384 | 544.9 | 563.5 | +3.31% |
| 32768 | 490.6 | 523.8 | +6.34% |
| 49152 | 446.3 | 489.1 | +8.76% |
| 65536 | 410.7 | 460.8 | +10.89% |

Monotone in context; gen deltas within noise (|Δ| <= 0.44%, sign-mixed). tiled4 alone measured +5.38% @64k on the same protocol — the 2.0 -> 1.5 -> 1.0 loads-per-MMA ladder maps monotonically onto throughput. Running in production on this box (goldens unchanged).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ endif
 .PHONY: all help clean test test-metal-session-batch test-mxfp4-cuda test-cuda-session-batch test-cuda-mixed-batch dspark-acceptance dspark-verify-depth mtp-verify-depth cpu cuda cuda-spark cuda-generic cuda-regression strix-halo rocm
 
 ifeq ($(UNAME_S),Darwin)
-.PHONY: metal-decode-schedule-bench metal-prefill-variant-bench check-mxfp4-half-lut
+.PHONY: metal-decode-schedule-bench metal-prefill-variant-bench check-mxfp4-half-lut test-indexer-scorers
 
 all: ds4 ds4-server ds4-bench ds4-eval ds4-agent
 
@@ -76,6 +76,7 @@ help:
 	@echo "  make test         Build and run tests"
 	@echo "  make metal-decode-schedule-bench  Build the balanced Metal decode schedule benchmark"
 	@echo "  make metal-prefill-variant-bench  Build the balanced Metal prefill variant benchmark"
+	@echo "  make test-indexer-scorers  Compare all pre-M5 indexer scorers byte-for-byte"
 	@echo "  make check-mxfp4-half-lut  Verify the checked-in MXFP4 half LUT matches the generator"
 	@echo "  make test-mxfp4-metal  Check the MXFP4 half LUT, then run Metal MXFP4 exactness tests"
 	@echo "  make dspark-verify-depth  Run DSpark speculative verification smoke if support GGUF is present"
@@ -124,6 +125,15 @@ speed-bench/metal_prefill_variant_bench: speed-bench/metal_prefill_variant_bench
 	$(CC) $(CFLAGS) -o $@ $^ $(METAL_LDLIBS)
 
 metal-prefill-variant-bench: speed-bench/metal_prefill_variant_bench
+
+tests/test_indexer_scorers.o: tests/test_indexer_tiled4.c ds4_gpu.h
+	$(CC) $(CFLAGS) -I. -c -o $@ $<
+
+tests/test_indexer_scorers: tests/test_indexer_scorers.o $(CORE_OBJS)
+	$(CC) $(CFLAGS) -o $@ $^ $(METAL_LDLIBS)
+
+test-indexer-scorers: tests/test_indexer_scorers
+	./tests/test_indexer_scorers
 
 tests/test_mxfp4_metal.o: tests/test_mxfp4_metal.c ds4_gpu.h
 	$(CC) $(CFLAGS) -I. -c -o $@ $<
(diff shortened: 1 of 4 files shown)


JordiPosthumus:
Independent M3 Ultra validation of the exact current heads (`#830` at
`55b7103`, `#831` at `d9bece7`): this looks like a real long-context win.

Test machine/model: Apple M3 Ultra 512 GiB, pre-M5 Metal path,
DeepSeek-V4-Flash-Vision-Exp MXFP4, 262,144 context, 4,096 prefill cap. I used
the balanced `metal_prefill_variant_bench` with fresh sessions and ABBA/BAAB
ordering; it aborts on any byte difference in the final 129,280-float vocab
vector.

Results (optimized throughput divided by rollback throughput):

- #830 + #831 vs legacy: +0.67% at 8K (585.20 vs 581.31 t/s), +2.56% at 32K
  (544.81 vs 531.18), and +6.32% at 64K (524.25 vs 493.11).
- #830 tiled2 vs legacy at 32K: +1.23% (551.32 vs 544.61).
- #831 tiled5 vs tiled2 at 32K: +2.06% (548.83 vs 537.74).

All 3,102,720 full-model logit floats across those comparisons were
byte-identical. I also compared legacy/tiled2/tiled4/tiled5 directly over
2,132,086 model-free output cells covering the 31/32/33-token gate,
63/64/65-row edges, unaligned resumed positions, causal masking, and 64K-sized
compressed frontiers; all were byte-identical. A cleanup/reinitialize repeat,
the existing two-session real-model oracle (`max_abs=0`), and an
image-conditioned cancel/retry smoke also passed.

Two small polish items surfaced while testing:

1. The new Q16/K16 scratch buffers and capacities are not cleared by
   `ds4_gpu_cleanup()`.
2. `tests/test_indexer_tiled4.c` is not wired into the Makefile, and at the
   final #831 head its one fixed shape compares the default tiled5 path with
   tiled4 rather than checking every new path directly against legacy.

I prepared one clean follow-up commit directly on your #831 head that only
adds that cleanup and expands/wires the regression coverage; it does not alter
the scorer or its dispatch policy:
[`f7ad40c`](https://github.com/JordiPosthumus/ds4/commit/f7ad40c68373654c42b88ec7ee74c1526e185a83).
Happy for you to cherry-pick any or none of it. The raw local evidence and exact
commands are preserved separately; I can post a compact log excerpt if useful.

Thanks for the optimization—our measurements independently reproduce the
shape of your result, including the gain growing with context.

adriangalilea:
Thank you for the independent reproduction, and for the cleanup commit. Both items were real: the Q16/K16 scratch was never released in `ds4_gpu_cleanup()`, and `tests/test_indexer_tiled4.c` was neither wired into the Makefile nor comparing every new path against legacy.

Adopted `f7ad40c` as-is on this branch (your authorship preserved), so the PR head is now `d9bece7` + your commit. Reproduced your suite here on the pre-M5 path (M4 Pro): `make test-indexer-scorers` → `PASS: all indexer scorer regressions`, legacy/tiled2/tiled4/tiled5 identical on every case including the 64k-visible frontier and the activation edge.

Your +6.32% at 64K on the M3 Ultra matches what we measured on ours (+6.7% cold 64k for tiled5 over tiled2), and the gain growing with context is the mechanism: the K tiles stay register-resident across the head loop, so the traffic term the scorer saves scales with the compressed frontier.

PR #832, opened 2026-08-18
metal: canonical top-k order + exact streaming top-512 selector

adriangalilea:
Stacks on #830 → #831 (branched from #831's branch; the last three commits are this PR). Fourth entry in the indexer prefill series — this one takes the top-k stage.

The Metal indexer top-k runs a padded bitonic block sort plus a merge cascade — O(n log n) work that grows with context while only 512 indices survive. The CUDA backend already solved this shape with the streaming selector (45f4ef9): one pass over the score row with a running 512th-best threshold. This PR ports it to Metal in two steps.

**Step 1 — a canonical total order.** The bitonic comparator had no tie break: equal scores keep whatever order the comparison network hands them — deterministic per size, but not a total order, so no streaming or partial selector can ever be compared bit-for-bit against it. The canon comparator breaks ties on the index (ascending); the merge kernel is already canonical (left run first), so the whole pipeline becomes a total order over (score desc, idx asc), matching the CUDA family. Tie order among exactly-equal scores is the only output change.

**Step 2 — the streaming selector.** One threadgroup per token walks the score row once, packing (score, idx) into ordered `ulong` keys and keeping a 2048-slot threadgroup buffer with a running 512th-best threshold. Compaction is atomic-free: `simd_ballot` + per-simdgroup counts, one thread exclusive-scans the sixteen counts — a fully deterministic shape. When the buffer nears capacity it collapses in place with a 2048-slot bitonic sort (once more at the end), replacing the padded sort + merge cascade entirely. Output is bit-identical to the canonical path; guards keep the legacy path for `top_k != 512` or fewer than 32 tokens.

Both are defaults with per-call rollback envs (`DS4_METAL_DISABLE_ARGSORT_CANON`, `DS4_METAL_DISABLE_TOPK_STREAM512`). `tests/test_topk_ab.c` compares both GPU paths against a CPU qsort ground truth across sizes, token counts, visibility patterns, and tie densities — 0 diffs on M4 Pro and M3 Ultra.

Measured on an M3 Ultra 512 GB serving DeepSeek V4 Flash 0731 MXFP4, 4-run ABBA medians, frontier logits identical between the two arms at all 32 context frontiers:

| ctx | without | with | Δ |
|---|---|---|---|
| 16384 | 563.9 | 565.9 | +0.36 % |
| 32768 | 524.1 | 530.3 | +1.17 % |
| 49152 | 489.4 | 500.5 | +2.27 % |
| 65536 | 460.6 | 475.1 | **+3.15 %** |

The gain is monotone with context, as replacing O(n log n) sort work with a single O(n) pass predicts. Generation deltas are noise (|Δ| ≤ 0.65 %, sign-mixed). Running in our production since today with zero output diffs on the serving bench.


adriangalilea:
One more datapoint for the monotonicity claim, measured after opening this PR: cold 128k prefill, single sync (chunk 4096), same box, this branch vs main:

| ctx | main | this PR stack | Δ |
|---|---|---|---|
| 131072 | 390.5 t/s | 484.9 t/s | **+24.2%** |

The gain keeps growing past the sweep's 64k frontier, consistent with the top-k's O(n log n) → O(n) mechanism and the scorer's traffic term both scaling with context. (Stack = #830 + #831 + this; ledger entry in the fork.)

JordiPosthumus:
Tested the exact current head `ddac36d` on an otherwise drained Apple M3 Ultra (512 GiB) with DeepSeek-V4-Flash-Vision-Exp MXFP4. This confirms the stack and also found a focused-test wiring issue.

The final commit made canonical argsort and stream512 default-on with `DS4_METAL_DISABLE_*` rollback variables, but `tests/test_topk_ab.c` still sets the earlier opt-in variables. As a result, its nominal canonical arm can also dispatch stream512, so stream-vs-canonical may pass without switching implementations.

I prepared a test-only correction at [`e06f342`](https://github.com/JordiPosthumus/ds4/commit/e06f342650dcba08d0b28f1acf34fb85f1d64f91), based directly on this PR head. It:

- selects the final paths with `DS4_METAL_DISABLE_ARGSORT_CANON` / `DS4_METAL_DISABLE_TOPK_STREAM512`;
- compares both GPU paths to an independent CPU `(score desc, index asc)` reference;
- covers the 31/32-token gate, minimum and odd row widths, dense finite ties, a 64K-sized compressed frontier, and non-512 fallback;
- poisons outputs, releases allocations, and adds `make test-indexer-topk`.

All seven cases passed: 135,136 selected entries per GPU path matched the CPU reference exactly.

Incremental stream512 A/B (canonical ordering in both arms, ABBA, full 129,280-float final-logit comparison after every run):

| Prefix | stream512 | canonical full argsort | stream512 gain |
| ---: | ---: | ---: | ---: |
| 32K | 540.1569 t/s | 534.0709 t/s | +1.1395% |
| 64K | 508.5590 t/s | 498.3680 t/s | +2.0449% |
| 128K | 438.2469 t/s | 404.0194 t/s | +8.4717% |

All 1,551,360 compared logits were bit-identical. These are incremental #832 numbers on top of #830/#831, so they are intentionally not directly comparable to the full-stack-vs-main figure above.

One quality boundary worth recording: canonical tie ordering does change long-context logits relative to legacy tie ordering, as the commit message allows. The repository's 100-case Vision-Exp scorer at 4K was byte-identical (`avg_nll=0.167972379` both), but that does not reach compressed indexer history. With streaming disabled in both arms, held-out next-token scoring after real compressed frontiers gave:

| Prefix | canonical avg NLL | legacy avg NLL | relative change | greedy matches |
| ---: | ---: | ---: | ---: | ---: |
| 32K | 1.021238518828 | 1.029397156880 | -0.7926% | 367 vs 363 |
| 64K | 1.212609296449 | 1.212505775262 | +0.00854% | 342 vs 340 |

Lower NLL is better; combined NLL favored canonical by 0.3593%. This is two 512-token *I promessi sposi* slices, not a universal quality claim, but I saw no meaningful regression.

Also passed: the #830/#831 scorer regression across 2,134,134 exact cells (including cleanup/reinit), optimized Metal build, and clean CPU portability build. The only compiler diagnostics were existing macOS 27 `didModifyRange:` deprecations.

Happy to open the test correction against your branch if that is easier; otherwise the commit above is ready to cherry-pick. Excellent kernel work—thank you.

adriangalilea:
Good catch on `tests/test_topk_ab.c`: it still selected the paths through the earlier opt-in variables, so after the default flip its "canonical" arm could dispatch stream512 and the A/B could pass without switching implementations. Adopted `e06f342` as-is on this branch (your authorship preserved); the PR head is now `ddac36d` + your commit. Reproduced here on an M4 Pro: `make test-indexer-topk` → all seven cases PASS, canonical/default/CPU reference identical (token gate, minimum row, odd width, dense finite ties, 64K compressed frontier, non-512 fallback).

Your incremental numbers line up with ours: +2.0% at 64K and +8.5% at 128K for stream512 alone, against our +24.2% at 128K for the full #830+#831+#832 stack over main. And thank you for recording the tie-order quality boundary with real compressed frontiers — the NLL data (combined −0.36%, i.e. slightly in favour of canonical) is exactly the evidence the commit message could only assert.

adriangalilea:
Rebased onto current main (`a04f46f`), which brought the V4.1 Flash support into the same argsort kernel. Resolution, in one extra commit on top:

- `canon` is now the fourth template parameter of `kernel_argsort_f32_i32` next to upstream's `causal`/`shuffle`, with the index tie-break applied in both the threadgroup comparator and the new SIMD-shuffle phase. Only the non-causal instance (`kernel_argsort_f32_i32_desc_canon`) is exported.
- The V4.1 indexer selects through the causal argsort/merge kernels (they bound each row's width) and keeps upstream's comparator; canon and stream512 apply to the V4 Flash (non-causal) rows only, where they were measured.

`make test-indexer-topk` on the rebased head: all seven cases PASS against the CPU reference (M4 Pro).

PR #835, opened 2026-08-19
Distributed speculative decoding for the pipeline split (MTP + DSpark)

davidcanar:
## Summary

This adds **target-model-verified speculative decoding to the two-node pipeline split** (`--role coordinator` / `--role worker`). The worker already owns the final layers, the output head, and the DSpark capture layers (40–42), so it is the natural drafter: it proposes draft blocks, the coordinator verifies them in batched spans through the normal route, accepted prefixes commit on both machines without replay, and consecutive full accepts chain span-to-span without intermediate per-token decodes.

Speculation only engages on the greedy path (temperature 0), mirroring single-node DSpark. Drafts are always verified by the target model: a bad draft costs acceptance rate, never correctness. The per-token path is untouched and remains the fallback at every stage (support model absent, `--dspark` off, scheduler backoff, or any span/commit failure).

The five commits are layered so each is independently revertable:

1. **Legacy-MTP speculation over the split** — the foundation. The worker runs the MTP head and returns advisory drafts alongside its logits (`F_OUTPUT_DRAFTS`); `F_SPEC_VERIFY` snapshots the worker-owned compressor/indexer frontiers and returns one logits row per token; `F_SPEC_ROLLBACK` restores frontier and token timeline on a miss.
2. **DSpark drafting over the same cycle** — worker-side propose from slice-eval capture, gated by the DSpark confidence/scheduler machinery. Also fixes a latent bug: zero-draft `LOGITS_DRAFTS` replies left the draft count uninitialized.
3. **Prefix commits** — partial accepts no longer replay the accepted prefix. Verify spans capture per-prefix frontier state; a new `DS4_DIST_WORK_F_SPEC_COMMIT` work item commits the accepted prefix on both machines with a bare ACK, and the coordinator reuses the verify's own logits rows. Falls back to the rollback+re-eval path on any mismatch.
4. **Fused spans** — the speculative base decode is folded into the next verify span. On a full accept the worker proposes continuation drafts from the last verified row and returns them in the same reply, so accepted blocks chain with zero extra round trips.
5. **Full-block spans + small-batch kernel work** — `DS4_SPEC_PREFIX_SLOTS` 4→5 so a span carries a whole 5-draft DSpark block, plus a barrier-free direct Q8_0 batch matmul kernel for 2–8-row speculative batches (bit-identical per-token reduction order) and diagnostic env rollbacks from the tuning pass (defaults unchanged).

## Measured results

Hardware: 2× AMD Strix Halo (gfx1151, ROCm), Thunderbolt-net TCP link (~9 Gbit/s, 0.4 ms RTT), DeepSeek V4 Flash (43 layers, Q4K experts) split `0:22` / `23:output`, greedy requests.

| Workload | Plain pipeline | This PR (`--dspark`) |
|---|---|---|
| code-gen, 400 tokens | 40.3 s (~13.0 t/s decode) | 35.0–36.6 s (~13.4 t/s) |
| prose, 300 tokens | ~12 t/s | ~11.6–12 t/s (parity) |

- Engagement evidence on the code run: 28 fused spans, 18 carrying the full 5-draft block, 16 fully accepted, zero fallbacks or token-hash errors.
- On prose the confidence gate and scheduler back off (mostly 0–2 drafts proposed, below the 3-draft minimum) instead of paying for failing verifies — worst case is parity, not a regression.
- Cost model behind the `DS4_DIST_SPEC_MIN_DRAFT=3` default: a 4–6-row verify span costs ~290 ms through both slices including the wire, vs ~102 ms per-token base decode, so a span needs ≥3 accepted drafts to win.
- Greedy output was verified byte-identical to the per-token pipeline path.

## Wire protocol changes

- New work-item flag `DS4_DIST_WORK_F_SPEC_COMMIT` (added to the valid mask): token-span payload, no hidden state, ACK reply, no eval. Frame-level hidden-size validation is gated for this kind.
- Verify replies may carry trailing advisory drafts (`LOGITS_NROWS` + drafts) so a full accept feeds the next span directly.
- Coordinator and worker must run the same build; an older worker rejects the new flag as an invalid item (fails the request, no protocol desync).

## How to test

Worker first, then coordinator. `--mtp <support gguf> --dspark` goes on **both** sides (the worker drafts, the coordinator schedules):

```
# worker (owns layers 23..output + the output head)
./ds4-server --role worker --layers 23:output --coordinator <coord-ip> 8088 \
  -m DeepSeek-V4-Flash-....gguf \
  --mtp DeepSeek-V4-Flash-DSpark-support-0731.gguf --dspark \
  --ctx 131072 -t 32 --warm-weights

# coordinator (owns layers 0..22 + serving)
./ds4-server --role coordinator --layers 0:22 --listen <coord-ip> 8088 \
  -m DeepSeek-V4-Flash-....gguf \
  --mtp DeepSeek-V4-Flash-DSpark-support-0731.gguf --dspark \
  --ctx 131072 -t 32 --warm-weights --host 0.0.0.0 --port 1234
```

Requests must be greedy or speculation skips silently — send `"temperature": 0` explicitly (many clients default to 0.7–1.0).

```
curl http://localhost:1234/v1/chat/completions -H "Content-Type: application/json" -d '{
  "messages":[{"role":"user","content":"Write a C linked list with insert and delete."}],
  "max_tokens":400, "temperature":0}'
```

Observability (all optional):

| Env | Effect |
|---|---|
| `DS4_DSPARK_STATS=1` | one-line propose/accept summary per session |
| `DS4_MTP_SPEC_LOG=1` (coordinator) | per-step gate decision (received temperature, run/skip) and per-cycle `spec fused drafted=N accepted=M` / verify timings |
| `DS4_DSPARK_SPEC_LOG=1` (worker) | per-cycle drafter detail |
| `DS4_DIST_SPEC_MIN_DRAFT=N` | span minimum (default 3) |
| `DS4_MTP_SPEC_DISABLE=1` | kill switch, keeps the support model loaded |

What good looks like in the coordinator log: `mtp spec gate ... temp=0.000 ... -> run`, then `spec fused drafted=5 accepted=5` chains on code-shaped output, `dist spec prefix commit accepted=N` on partial accepts, and zero `span failed` / hash-mismatch lines. Correctness check: the same greedy prompt with and without `--dspark` must produce identical text.

## Limitations and non-goals

- **Greedy only** (same as single-node DSpark): temperature > 0 falls back to per-token decoding.
- **Two-node pipeline only.** Tensor-parallel DSpark was measured and intentionally not pursued: TP's per-layer gate exchanges put the 6-row verify floor at 210–250 ms against plain TP's ~68 ms/token, which is net-negative at any acceptance rate on this hardware.
- Tested on ROCm/gfx1151 only. The distributed paths use backend-neutral GPU APIs, but Metal is untested.
- The rocm diagnostic knobs (`DS4_ROCM_BATCH_DIRECT`, `DS4_ROCM_ATTN_OUTPUT_CUBLAS`, `DS4_ROCM_SHARED_DOWN_CUBLAS`, `DS4_ROCM_Q4K_SORTED_MIN`) are explicit-rollback style, defaults unchanged.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -50973,9 +50973,11 @@ int ds4_engine_mtp_draft_tokens(ds4_engine *e) {
     }
     if (ds4_engine_has_mtp(e)) return e->mtp_draft_tokens;
 #ifndef DS4_NO_GPU
+    /* DSpark drafts flow through the distributed route too: the worker owns
+     * the capture target layers in the pipeline split and runs the propose
+     * there, so the coordinator sizes its cycle from the same block. */
     if (e &&
         e->backend != DS4_BACKEND_CPU &&
-        e->distributed.role == DS4_DISTRIBUTED_NONE &&
         e->support_kind == DS4_SUPPORT_DSPARK &&
         e->dspark &&
         e->dspark_weights.block_size > 1) {
@@ -51178,6 +51180,65 @@ int ds4_session_dist_mtp_draft(ds4_session *s, int token, uint32_t pos,
     return 0;
 }
 
+static bool ds4_session_prepare_dspark_draft(ds4_session *s,
+                                             int token,
+                                             uint32_t pos);
+
+int ds4_session_dist_dspark_draft(ds4_session *s, int token, uint32_t pos,
+                                  int *drafts, int max_drafts, int *n_drafts,
+                                  char *err, size_t errlen) {
+    if (n_drafts) *n_drafts = 0;
+    if (!s || !drafts || max_drafts <= 0 || !s->engine) {
+        if (errlen) snprintf(err, errlen, "invalid distributed DSpark draft request");
+        return 1;
+    }
+    ds4_engine *e = s->engine;
+    if (e->support_kind != DS4_SUPPORT_DSPARK || !e->dspark) {
+        if (errlen) snprintf(err, errlen, "DSpark runtime drafting is not enabled");
+        return 1;
+    }
+    /* Same contract as the local speculative cycle: prepare fills the
+     * session draft block from the captured target hiddens, or leaves it
+     * invalid when the scheduler or the confidence gate skips this cycle
+     * (that is a normal outcome, not an error). */
+    (void)ds4_session_prepare_dspark_draft(s, token, pos);
+    if (!s->dspark_draft_valid || s->dspark_draft_len == 0) {
+        ds4_session_dspark_scheduler_note(s, 0, true, 0.0);
+        return 0;
+    }
+    int n = (int)s->dspark_draft_len;
+    if (n > max_drafts) n = max_drafts;
+    for (int i = 0; i < n; i++) drafts[i] = s->dspark_draft_tokens[i];
+    /* The coordinator consumes the block this cycle either way. */
+    s->dspark_draft_valid = false;
+    s->dspark_draft_len = 0;
+    *n_drafts = n;
+    return 0;
+}
+
+int ds4_session_dist_support_draft(ds4_session *s, int token, uint32_t pos,
+                                   int *drafts, int max_drafts, int *n_drafts,
+                                   char *err, size_t errlen) {
+    if (n_drafts) *n_drafts = 0;
+    if (!s || !s->engine) {
+        if (errlen) snprintf(err, errlen, "missing distributed draft session");
+        return 1;
+    }
+    switch (s->engine->support_kind) {
+    case DS4_SUPPORT_DSPARK:
+        return ds4_session_dist_dspark_draft(s, token, pos, drafts,
+                                             max_drafts, n_drafts,
+                                             err, errlen);
+    case DS4_SUPPORT_MTP_LEGACY:
+        return ds4_session_dist_mtp_draft(s, token, pos, drafts,
+                                          max_drafts, n_drafts,
+                                          err, errlen);
+    default:
+        if (errlen) snprintf(err, errlen, "no support model for drafting");
+        return 1;
+    }
+}
+
 bool ds4_session_dist_frontier_snapshot(ds4_session *s,
                                         ds4_dist_mtp_frontier *f) {
     if (!f) return false;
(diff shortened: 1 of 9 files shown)


jyatesdotdev:
Ran this on a second hardware setup: two AMD Strix Halo machines (Radeon 8060S, gfx1151, ROCm) over Thunderbolt, layers split 0:21 and 22:output — same class of machine as yours, but with the MXFP4-experts build of DeepSeek V4 Flash instead of Q4_K, plus the DSpark support model, `--mtp` and `--dspark` on both sides, temperature 0, your linked-list prompt, 400 tokens.

**The mechanism itself works exactly as described.** Speculation engaged on every request, fused spans chained, and prefix commits behaved:

```
ds4: dist mtp spec fused drafted=5 accepted=5 cont=1 span=349.3ms
ds4: dist spec prefix commit accepted=4
```

Fourteen fused spans on the code prompt, several full five-draft blocks fully accepted, no span failures and no hash errors. Your orchestration holds up on different hardware.

**Two findings from the MXFP4 side, both format-dependent rather than flaws in this pull request:**

1. **Performance.** A six-row verify span costs 341–350 ms on MXFP4 experts against your 290 ms on Q4_K, while the per-token baseline here is 69 ms (14.5 tokens per second). Even with excellent acceptance the end-to-end result was slower with speculation than without: 14.5 tokens per second base, 13.4 with speculation enabled.

2. **Correctness.** The speculative output differed from the per-token output for the same prompt (first divergence twenty characters in), and two consecutive speculative runs of the same prompt also differed from each other.

The likely root cause is in the MXFP4 expert kernels, not in your code: batches of two to four rows run on direct kernels that preserve the single-row accumulation order, but larger spans — and a full five-draft block plus its base token is six rows — fall onto the prefill tile path, which accumulates in a different order and is expensive at small sizes. Q4_K testing cannot see this because its small-batch path preserves order (your direct kernel plus the sorted-pairs work). In other words: the span path silently assumes that every batched kernel a span touches is bit-identical, per row, to the single-row kernel. That invariant might be worth stating explicitly in the code or the pull request, so each backend and quantization format can be audited against it.

We run order-exact batched expert kernels for exactly these span sizes on our production branch (the same family as the k-row tier in #821). Next step on our side is to put those under this driver and re-run the comparison — if spans become exact and cheap on MXFP4, that lands as a follow-up to this pull request with measurements. Happy to share raw logs or run other configurations if useful.

PR #837, opened 2026-08-20
feat(qwen): Qwen3.8 27B family + DFlash2 speculative (squashed, qwen-only)

audreyt:
**Status: ready for review.** Head updated `cca10ae` -> `bd9fda2` with four verification commits on top of the original payload. Original description preserved below.

Post-push verification found and fixed four regressions on this branch. Each fix is a separate commit, each gated on byte-identical greedy A/B vs the known-good tree (`b94c2ee` lineage) and a real Metal pass on `tests/test_qwen38_bench.sh`:

1. **Restored the two kernel review fixes below**, which had regressed relative to `qwen38`: Q4_64A float4 chunk addressing in `glm_q4_64a_dequant4` (`(idx4>>1)*2` -> `idx4*2`), and the `kernel_qwen_gqa_attn_decode` K/V race. The race fix removes the ordering dependency entirely - current-position reads come from `k_new`/`v_new`, not the cache - which is option 2 in Flor1an-B's comment below. Their analysis was correct: `threadgroup_barrier(mem_device)` orders lanes within one threadgroup only, and `h = tgpig` makes each head its own threadgroup, so the barrier could never close this race.
2. **Restored six lost fields of `ds4_qwen_gdn_core_args`.** Without them `metal/qwen_gdn.metal` failed to compile, taking down the whole Metal program - every model silently fell back to CPU at the previous tip.
3. **Root-caused the degenerate-output regression** (`375703f`): the MLX GDN core kernel port dropped per-head q/k l2-normalization, cross-channel output RMS norm, and ssn_norm weighting before z-gating (buffer bound but never read); its `(32,1,1)`-per-dv_idx layout cannot compute the cross-channel RMS norm at all. All 48 GDN layers ran through it on every forward. Scalar kernel restored everywhere, unfaithful mlx kernel deleted; plain-CLI greedy decode is again byte-identical to the known-good tree on both Q4_64A and Q4_K_M models.
4. **Fixed the server/session Metal route for the hybrid model** (`bd9fda2`): session-side prefill/decode used a full-attention-only forward that zero-filled every GDN (`qwen_linear_attn`) layer and never advanced its recurrent state, so every `--metal` server completion degraded to symbol soup - on every prior build of this branch (pre-existing; the CLI path used the correct hybrid forward, which is why all CLI A/Bs kept passing). All four session call sites now route through the hybrid forward and the dead forward is deleted. Verification: `/v1/completions` and `/v1/chat/completions` produce coherent output; raw CLI A/B is three-way byte-identical (known-good baseline == upstream port == this commit); CPU and Metal outputs are byte-identical.

Verification evidence per commit: decoded-text cmp clean vs baseline on both the bench driver and the plain CLI path, plus full `tests/test_qwen38_bench.sh` pass with Metal genuinely running (no CPU-fallback warning).

---

feat(qwen): Qwen3.8 27B family + DFlash2 speculative (squashed, qwen-only)

Single squashed commit of qwen38 qwen payload only, onto antirez/main@84cc882.

This keeps ONLY qwen-specific changes (30 files, +10852/-120) — origin/main
sync (39 files, idle-timeout, DSpark, PR #489/#755, dir-steering,
speed-bench, etc.) is NOT included and should land via its own PRs.

Qwen 3.8 family (27B hybrid 48×GDN + 16×full, 5120 hidden, 64 layers)
- New GGUF types Q4_64A (36) / Q2_64A (37) affine g64 + Q8_0 token_embd.
  Q2_64A 7.1G fits 8GB Neo vs Q4 13-14G; CPU 0.47 vs 0.36, Metal 7.5 vs 7.3 t/s.
- CPU: Q4/Q2 matvec/embed, pooled Metal GQA expand+fill, causal GQA
  attention with per-engine KV cache, GDN (QWEN_GDN_*) + SwiGLU FFN.
- Metal: dense.metal Q2_64A kernels, get_rows Q4/Q2, moe.metal GDN,
  dflash2.metal, qwen_gdn.metal, repeat.metal; hybrid
  qwen_hybrid_metal_forward_tokens pooled batch dispatch.
- NVFP4 type 40 (ollama 27b-mlx) 8-wide float4 qmv, 29 t/s greedy;
  tiled conv V, -exp(A_log), e/h_proj split; converter
  gguf-tools/ollama_nvfp4_to_gguf.py.

MTP / speculative (greedy-identical vs llama.cpp)
- Official MTP F16 sidecar (DS4_QWEN_MTP_HEAD, 849MB) + ofou q2-q4
  compact head with 98304+control map and rerank table; top-32
  shortlist + F16 rerank, packed affine-2 g64.
- Persistent 4096-slot MTP KV with history prefill walk (0-based,
  rewrite from target hiddens on accept), one dispatch/layer verify
  (row-batched split/rope/decode), single eval/round.
- Adaptive K: 2 default, 5 when draft-2 EMA>0.85; longcopy 43->61.7
  t/s on M5 Max (generation clock excludes KV warmup), capitals 54.8.
- Optional 2-tap conv / target_norm, fused NVFP4 SwiGLU, skip
  redundant primary repair & skip KV repair on full accept.

DFlash2 (block diffusion)
- ds4_dflash2.inc (880 lines) + arch dflash detection, CLI
  --dflash/--dflash2/--draft-model/--spec-type draft-dflash/
  --dflash-n-max 7; 4×32 tokens in one hybrid Metal batch
  (DFlash2 1.1G Q4_K_M draft, block 8).

Server / concurrency
- Enable speculative in batched_mode (temp<=0 && (mtp||dflash));
  per-engine qwen state (qwen_cpu_kv/gdn in ds4_engine, removes
  g_qwen_global_mu); batched pool sized from
  placement_session_count_hint; greedy temp 0 default for Qwen
  (fixes 1+1= -> 2).

Tooling / docs / tests
- gguf-tools/{gguf_requant_q4_64a,mtp_head_to_gguf,
  ofou_mtp_head_to_gguf,ollama_nvfp4_to_gguf,quants}
- download_model.sh qwen-dflash/qwen3.8/qwen-q4/q8 aliases (ggml-org
  + z-lab, links ds4flash.gguf+qwen38.gguf); README DFlash2 section;
  speed-bench/bench_qwen38.sh + tests/test_qwen38_bench.sh harness;
  metal/dflash2.metal+qwen_gdn.metal.

Bench M5 Max (Qwen3.8-27B NVFP4/Q2_64A 7.1G, greedy-identical)
  AR 29.4 t/s -> K=2 50, longcopy K=5 61.7, capitals 54.8, DFlash 4×32.

Base: antirez/main 84cc882
Head: qwen38-squashed bd9fda2 (review fixes restored, Metal compile fixed, GDN scalar kernel restored, server session route fixed)
Co-authored-by: audreyt <au@civic.ai>


Flor1an-B:
Re: the `kernel_qwen_gqa_attn_decode` race flagged by the automated review — the
`threadgroup_barrier(mem_flags::mem_device)` added in the latest commits doesn't
close it.

`threadgroup_barrier` only synchronizes threads *within the same threadgroup*. But
`h = tgpig` (`threadgroup_position_in_grid`) means each head is dispatched as its
**own threadgroup**. The barrier ensures head 0's own SIMD lanes agree with each
other before proceeding, but does nothing to stop head 1/2/3...'s threadgroups —
which can start executing at any time relative to head 0's — from reading
`k_cache`/`v_cache` before head 0's write has landed. There's no cross-threadgroup
ordering guarantee in a single dispatch on Metal; `mem_flags::mem_device` widens
what the barrier covers (device memory, not just threadgroup memory) but doesn't
change its scope from "this threadgroup" to "this dispatch."

Two ways to actually fix it, roughly in order of how invasive they are:
- Split into two dispatches: a small kernel (or a `dispatchThreads` call gated to
  `h==0`) that writes `k_cache`/`v_cache` for the current position, encoded and
  committed *before* the attention-decode dispatch that reads it. A command-buffer
  boundary (or at least an explicit `MTLFence`/`MTLEvent` between the two
  encodes) gives the ordering that's missing today.
- Or have every head redundantly write its own `kv_h` slice (the PR's own commit
  message figured this was fine since values are identical across heads sharing
  a KV group) — cheap, no dispatch restructuring, and removes the ordering
  dependency entirely instead of trying to synchronize it.

Happy to help verify against a repro if useful — this is exactly the kind of bug
that won't show up in a quick smoke test (non-deterministic, depends on GPU
scheduling) but can silently corrupt decode under real load.

audreyt:
Closing in favour of #844, which contains this branch's content as its first
commit (`5f10ce4`) plus the Ornith 1.5 hybrid runtime and classic DFlash draft
on top. Keeping both open just splits review across two copies of the same
diff.

Nothing from the review here is dropped. Where each item landed:

| Reported here | State in #844 |
|---|---|
| `kernel_qwen_gqa_attn_decode` writes the cache only from `h == 0` while every head reads it | Fixed. The loop now reads the current position straight from `k_new`/`v_new`, the cache write is narrowed to `h % group == 0` writing its own KV slice, and the `mem_device` barrier is gone. Folded into commit 1, so no commit in #844 carries the race. |
| `glm_q4_64a_dequant4` mapped `idx4` 0 and 1 to the same 2-byte chunk | Fixed: the offset is `idx4 * 2u`. |
| `ds4_cli.c` forcing `dflash_draft_n_max = 7` | The CLI no longer forces it — it just parses the flag. The real bound was the verify batch width, which is now named `DS4_DFLASH2_MAX_DRAFT` and applied once when the draft loads. |
| `ds4_server.c` doing the same | Same fix, same place. |
| `tests/test_qwen38_bench.sh` reading `gguf-tools/quants.h` by relative path | Fixed: the harness passes `$ROOT/gguf-tools/quants.h`. The check was also rewritten to assert the enum bound by name (`COUNT == max_id + 1`) instead of a hardcoded literal. |

@Flor1an-B — your cross-threadgroup analysis was right, including the part I had
wrong about `mem_device` widening coverage rather than scope. The fix follows
your second suggestion: remove the ordering dependency rather than try to
synchronize it. You are credited in commit `5f10ce4`. Detail and an invitation
to look at the rest of the new Metal are in #844.

Issue #791, opened 2026-08-13
Cannot create a session on a discrete NVIDIA GPU: spurious cudaMalloc OOM (four causes, two with patches)


daimonionnn:
### Summary

On a discrete NVIDIA GPU (RTX PRO 6000 Blackwell, 96 GiB, driver 595.84), ds4
builds cleanly but **cannot create a session at all** — every session
`cudaMalloc` fails with a spurious out-of-memory, including a 230 MiB context
buffer for a 2048-token session while `nvidia-smi` reports ~18 GiB free.

Four causes, each hiding the next. The first two look portable and I have
patches for them; the last two are design questions for a discrete card rather
than something I would presume to fix.

With all four worked around locally, the same machine runs the antirez IQ2XXS
quant at **2161 tok/s prefill / 72.6 tok/s generation** at 2k, and ~2100 tok/s
prefill flat out to 65k — so the engine itself is in good shape here once it can
allocate.

---

### 1. `cudaHostRegisterReadOnly` is not supported by this driver

`cudaDevAttrHostRegisterReadOnlySupported` reads **0** on driver 595.84.
`ds4_gpu_register_model_map_no_copy()` passes
`cudaHostRegisterMapped | cudaHostRegisterReadOnly` unconditionally, so it fails
with *"operation not supported"* and there is no fallback:

```
ds4: CUDA (no-copy) host registration skipped: operation not supported
```

A retry without the flag would be enough — the mapping is not written through.

### 2. The Linux mapping cannot be pinned at all (`r--s`)

Retrying without `ReadOnly` then fails with *"invalid argument"*, because on
Linux the model is still mapped through the Metal branch in `ds4.c`:

```
7fa...-7fb... r--s 00000000 103:02 ...  DeepSeek-V4-Flash-IQ2XXS-...gguf
```

`MAP_SHARED` + `PROT_READ`. I probed the driver directly to find the rule:

| mapping | `cudaHostRegister` |
|---|---|
| anonymous `malloc`, 64 MiB | ok |
| file-backed `MAP_PRIVATE`, `PROT_READ\|PROT_WRITE` | ok |
| file-backed over an `O_DIRECT` fd | ok |
| anonymous `mmap`, 8 GiB | ok |
| file-backed `PROT_READ` + `ReadOnly` flag | **operation not supported** |

So `PROT_READ` + `ReadOnly` is the only rejected combination. Mapping
[...]

PR #872, opened 2026-08-27
CUDA: reject prefill mixed shapes wider than the shared scores window (#803)

### LINKED ISSUE #803: CUDA: prefill mixed scalar fallback has a fixed `scores[512]` and no `n_score` cap (ROCm / decode already check)
## Summary

`attention_prefill_mixed_kernel` stores every attention score in `__shared__ float scores[512]`. `n_score = raw_count + visible_comp` is not capped, and the last hop of `attention_prefill_mixed_launch` does not call `cuda_attention_score_buffer_fits` or refuse the launch.

The same kernel on ROCm uses `DS4_ROCM_ATTENTION_PREFILL_MIXED_SCORE_CAP` (2048), returns in-kernel if `n_score` is above the cap, and the host refuses the launch with a log line. CUDA decode mixed attention uses `DS4_CUDA_ATTENTION_SCORE_CAP` (8192) plus `cuda_attention_score_buffer_fits` and an online fallback. CUDA prefill mixed last hop has neither.

```c
__shared__ float scores[512];
uint32_t n_score = raw_count + visible_comp;
/* writes scores[r] and scores[raw_count + c] with no cap */
```

`attention_prefill_mixed_launch` tries token-tile, then `attention_static_mixed_heads8_online_kernel` (`head_dim==512`, not `--quality`, `n_tokens>=128`), then cublas (`g_cublas_ready && head_dim==512 && n_tokens>1`), then the 512-slot kernel. `DS4_CUDA_NO_WINDOW_ATTENTION` and `DS4_CUDA_NO_CUBLAS_ATTENTION` skip the two fast paths and land on that last hop.

## Environment

- HEAD `84cc882`
- 2× RTX 2080 Ti (sm_75), Ubuntu, nvcc / compute-sanitizer 12.0, nsys 12.0
- `make cuda-generic`

## Repro

Zero-prefix Flash prefill on a ratio-4 layer with a first chunk of about 1600 tokens reaches `ds4_gpu_attention_prefill_static_mixed_heads_tensor` with `n_comp=400`, `window=128`, `head_dim=512`. Last token `n_score = 128+400 = 528 > 512`. With both fast paths disabled the launch is the 512-slot kernel:

```sh
DS4_CUDA_NO_WINDOW_ATTENTION=1 DS4_CUDA_NO_CUBLAS_ATTENTION=1 \
  ./ds4 --cuda -m ds4flash.gguf -p "<prompt of ~1600 tokens>"
```

Same numbers through the executor API the engine uses (`n_tokens=1600, n_comp=400, window=128, ratio=4, head_dim=512`, both env vars set):

- nsys: 100% of GPU time is `attention_prefill_mixed_kernel` (grid `1600 x 1`, block 256).
- `n_score=528` writes 16 floats past `scores[512]` into the same block's `__shared__ float partial[256]`. `cudaDeviceSynchronize` succeeds. compute-sanitizer memcheck treats the whole static shared region as one object and stays silent.
- To make the write leave the shared region (same kernel, larger `n_score`): `n_tokens=1024, n_comp=400, window=0, ratio=1, n_head=1, head_dim=8` → last token `n_score=1424`. compute-sanitizer `--tool memcheck` on 2026-08-14:

```
prefill mixed trigger: n_tokens=1024 n_comp=400 window=0 ratio=1
  n_head=1 head_dim=8 last n_score=1424 (cap 512)
attention_prefill_static_mixed_heads rc=1
Invalid __shared__ write of size 4 bytes
  at ds4_cuda.cu:7149:attention_prefill_mixed_kernel
  Address 0xc10 is out of bounds
ERROR SUMMARY: 1164 errors
```

`:7149` is `scores[raw_count + c] = s`. Bare CUDA then fails `cudaDeviceSynchronize` with `unspecified launch failure`.

## Suggested fix

Before launching the 512-slot kernel, compute `n_score` the same way the kernel does. If it is above 512, refuse the launch or take `attention_static_mixed_heads8_online_kernel` when `head_dim==512`. Add `if (n_score > cap) return` inside the kernel, matching ROCm. Do the same for `ds4_gpu_attention_prefill_masked_mixed_heads_tensor`; it shares this launch.

tedin7:
Fixes #803.

## What

`attention_prefill_mixed_kernel` (the scalar prefill fallback) stores every attention score in a fixed `__shared__ float scores[512]`, and neither the kernel nor the last hop of `attention_prefill_mixed_launch` caps `n_score = raw_count + visible_comp`. Shapes wider than 512 write past the array — silently while the overflow lands in the adjacent `partial[256]`, and as an invalid shared write beyond that (see the analysis in #803).

Following the fix suggested in the issue, and mirroring what the ROCm side already does (`rocm/ds4_rocm_attention_launch.cuh`, `DS4_ROCM_ATTENTION_PREFILL_MIXED_SCORE_CAP`):

- **Host**: `attention_prefill_mixed_launch` computes the worst-case score count (`max_raw = min(window, n_tokens)` or `n_tokens` when `window == 0`, plus `n_comp`) and refuses the launch with the same log line as ROCm when it exceeds 512. Both `ds4_gpu_attention_prefill_static_mixed_heads_tensor` and `ds4_gpu_attention_prefill_masked_mixed_heads_tensor` go through this launch, so both are covered.
- **Kernel**: `if (raw_count + visible_comp > 512u) return;` before any shared write, as defense in depth, matching the ROCm kernel.

## Regression test

New case in `tests/cuda_long_context_smoke.c` (runs under `make cuda-regression`):

- an overflow shape (`n_tokens=1024, window=0, ratio=4` → 1280 scores) must be refused by the launcher;
- a windowed shape that fits (128 raw + 256 comp = 384) must keep working.

`head_dim != 512` keeps the launcher off the token-tile / window / cublas paths so the test deterministically exercises the scalar fallback, without env vars.

## Validation

Environment: RunPod community pods, RTX 3080 and RTX 3090 (sm_86), driver 580.x, CUDA 12.4 toolkit, `make cuda CUDA_ARCH=sm_86`, upstream `c1d4597`. compute-sanitizer from CUDA 13.0 (`cuda-sanitizer-13-0`) — note for other reproducers: the 12.4 sanitizer fails to start at all on 580+ drivers (`Target application terminated before first instrumented API call`).

| | unpatched | patched |
|---|---|---|
| `compute-sanitizer --tool memcheck` on the new test | 40 × `Invalid __shared__ write` at `ds4_cuda.cu:7149` in `attention_prefill_mixed_kernel` (RTX 3090; 100+ on RTX 3080 with a wider run) | 0 invalid writes |
| new regression case | overflow shape accepted | overflow shape refused with the log line, fitting shape passes, `cuda long-context regression: OK` |
| `make test` | — | passes up to the suites that need the real GGUF (not run: no model weights on the pod) |

Two unrelated sanitizer reports remain before and after the patch: `cudaErrorNotSupported` from `cudaHostRegister` in `cuda_model_range_ptr` (`ds4_cuda.cu:747`), reached from the pre-existing decode test. That is the discrete-GPU host-register behavior discussed in #791, not introduced here.

Not done: a before/after `ds4-bench` run, since the pod had no model weights. The guard adds one comparison per launch on the host path only; the kernel guard is two integer adds and a compare per block before any shared traffic.

## Transparency

A commit referencing this issue (`2a2a887f`, by @marcodelpin) appears in #803's timeline, but the fork it lived in is no longer reachable and no PR ever landed. This implementation was done independently against current main.

### DIFF
--- a/ds4_cuda.cu
+++ b/ds4_cuda.cu
@@ -7122,6 +7122,10 @@ __global__ static void attention_prefill_mixed_kernel(
     uint32_t raw_count = t + 1u - raw_start;
     uint32_t visible_comp = (t + 1u) / ratio;
     if (visible_comp > n_comp) visible_comp = n_comp;
+    /* Defense in depth: the launcher rejects shapes wider than the shared
+     * scores window, but never write past it even if a new call site skips
+     * that check (issue #803). */
+    if (raw_count + visible_comp > 512u) return;
     __shared__ float scores[512];
     __shared__ float partial[256];
     __shared__ float max_s;
@@ -18095,6 +18099,21 @@ static int attention_prefill_mixed_launch(
                                                                         n_head,
                                                                         head_dim);
         return cuda_ok(cudaGetLastError(), "attention mixed unpack launch");
+    }
+    /* The scalar fallback kernel accumulates raw + compressed scores in a
+     * fixed __shared__ float scores[512]; refuse shapes whose widest block
+     * would write past it, same bound and log as the ROCm launch (#803). */
+    const uint32_t max_raw = (window != 0u && window < n_tokens) ? window : n_tokens;
+    if ((uint64_t)max_raw + n_comp > 512u) {
+        fprintf(stderr,
+                "ds4: CUDA attention mixed scalar fallback unsupported for %llu scores "
+                "(cap=%u, tokens=%u, comp=%u, window=%u)\n",
+                (unsigned long long)((uint64_t)max_raw + n_comp),
+                512u,
+                n_tokens,
+                n_comp,
+                window);
+        return 0;
     }
     dim3 grid(n_tokens, n_head, 1);
     attention_prefill_mixed_kernel<<<grid, 256>>>((float *)heads->ptr,
--- a/tests/cuda_long_context_smoke.c
+++ b/tests/cuda_long_context_smoke.c
@@ -156,10 +156,77 @@ static int check_decode_attention_overflow_path(void) {
     return rc;
 }
 
+/* Issue #803: the scalar prefill mixed fallback keeps scores in a fixed
+ * __shared__ float[512]. Shapes whose widest block needs more entries must be
+ * rejected by the launcher (so the caller falls back) instead of letting the
+ * kernel write past the array. */
+static int check_prefill_mixed_scores_window(void) {
+    /* head_dim != 512 keeps the launcher off the window/cublas paths and on
+     * the scalar fallback that owns the shared scores window. */
+    const uint32_t n_head = 2;
+    const uint32_t head_dim = 64;
+    const uint32_t ratio = 4;
+    const uint32_t max_tokens = 1024;
+    const uint64_t q_count = (uint64_t)max_tokens * n_head * head_dim;
+    const uint64_t raw_count = (uint64_t)max_tokens * head_dim;
+    const uint32_t max_comp = max_tokens / ratio;
+    const uint64_t comp_count = (uint64_t)max_comp * head_dim;
+    const uint64_t page = 4096;
+
+    float *q_host = (float *)calloc((size_t)q_count, sizeof(float));
+    float *raw_host = (float *)calloc((size_t)raw_count, sizeof(float));
+    float *comp_host = (float *)calloc((size_t)comp_count, sizeof(float));
+    float *model_host = (float *)calloc((size_t)page / sizeof(float), sizeof(float));
+    if (!q_host || !raw_host || !comp_host || !model_host) return 1;
+
+    ds4_gpu_tensor *heads = ds4_gpu_tensor_alloc(q_count * sizeof(float));
+    ds4_gpu_tensor *q = ds4_gpu_tensor_alloc(q_count * sizeof(float));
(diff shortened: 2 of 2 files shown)


tedin7:
Follow-up on the missing speed numbers: I attempted the CONTRIBUTING.md before/after `ds4-bench` run on the same machine used for the sanitizer validation (RTX 3090 24 GB, `ds4f-q2`, commit `c1d4597`).

Not achievable on this class of hardware:

- without `--ssd-streaming` the 81 GB model cannot fit a 24 GB card (`CUDA model arena alloc failed ... out of memory` during model prepare);
- with `--ssd-streaming`, `ds4-bench` crashes at the first 2048-token prefill with an illegal memory access in the layer-major streaming path — **identically on unpatched main and on this PR's build** (3/3 runs each). Filed separately with the exact repro as #881; it is unrelated to this change, which is host-side shape validation only (two integer comparisons per launch on the scalar-fallback path, no kernel or memory-traffic changes).

So: correctness evidence for this PR is the compute-sanitizer before/after and the new regression case; speed numbers need a machine where either the model fits in VRAM or #881 is resolved. Happy to rerun the sweep and post the CSVs if either becomes available.

PR #874, opened 2026-08-27
Tune the Metal backend for M3 Ultra

trueimage:
# Tune the Metal backend for M3 Ultra

Eleven commits. **+2.38% decode, prefill flat** on an M3 Ultra, measured with
interleaved A/B sweeps across 32 context lengths. Every change is
bit-exactness-verified against the shader it replaces.

**Roughly half of this PR's value is negative results.** Three changes were
implemented, measured, and reverted. A fourth -- raising the prefill chunk --
was implemented, measured as a *regression* in the standard benchmark
workload, and replaced by a fix for the crash it exposed. A systematic retune
found nothing to retune. Metal 4 turned out to work here and be slower. All of
it is written up under *What did not work*, so nobody re-derives it.

There is also a methodology finding worth more than the speedup: on this
machine a single unalternated benchmark sweep per arm produces errors of
5-6%, larger than everything being measured. See *How these numbers were
measured*.

> **Rebased onto current `main` (110afdd) and re-measured from scratch.**
> The headline number above is from the rebased branch against the current
> `main`, not carried over from the original campaign. What was and was not
> re-verified on this base is listed under *What was re-checked after the
> rebase*.

## Machine and model

| | |
|---|---|
| Device | Apple M3 Ultra, 80 GPU cores, 512 GB, macOS 26.5.2 |
| `MTLDevice.architecture.name` | `applegpu_g15d` (generation 15, tier d) |
| Backend | Metal |
| Model | DeepSeek-V4-Flash, **MXFP4** experts / F16 HC+compressor+indexer / Q8 attn+shared+out |
| Residency | 145.26 GiB resident, no SSD streaming |

## End-to-end result

Two blocks of four interleaved sweeps, arms in **M-B-B-M** and then **B-M-M-B**
order with a rebuild between each. Pooling the two blocks is what makes this
trustworthy: each arm occupies run positions {1,4} in one block and {2,3} in
the other, so any position-dependent effect -- including a non-linear one that
a single ABBA block cannot cancel -- drops out.

```sh
./ds4-bench -m ds4flash.gguf \
```

| ctx | main prefill | branch prefill | Δ | main decode | branch decode | Δ |
|---:|---:|---:|---:|---:|---:|---:|
| 2,048 | 639.28 | 639.57 | +0.04% | 42.80 | 43.88 | +2.54% |
| 8,192 | 567.65 | 568.04 | +0.07% | 38.35 | 39.27 | +2.39% |
| 16,384 | 539.11 | 539.21 | +0.02% | 37.52 | 38.39 | +2.31% |
| 32,768 | 487.97 | 487.57 | -0.08% | 36.01 | 36.86 | +2.35% |
| 65,536 | 411.13 | 410.61 | -0.13% | 33.49 | 34.17 | +2.02% |
| **mean (32 ctx)** | | | **-0.04%** | | | **+2.38%** |

Decode range across the 32 contexts: +2.02% to +2.65%, positive at every one.
Prefill is flat by design -- every win here is in the FP8 KV path, which is
decode-side work.

Per-block agreement, as a check that the pooling is not hiding a split:

| block | order | prefill | decode |
|---|---|---:|---:|
| `dsA` | M-B-B-M | -0.04% | +2.32% |
| `dsB` | B-M-M-B | -0.04% | +2.43% |
| **pooled** | position-balanced | **-0.04%** | **+2.38%** |

## GLM 5.3 Flash: a control, and it behaves like one

GLM 5.3 Flash Q4_K (177.8 GiB, also fully resident) was measured the same way,
one M-B-B-M block.

| ctx | main prefill | branch prefill | main decode | branch decode |
|---:|---:|---:|---:|---:|
| 2,048 | 429.77 | 429.53 | 21.06 | 21.04 |
| 8,192 | 391.63 | 391.90 | 20.64 | 20.66 |
| 16,384 | 388.09 | 388.25 | 20.56 | 20.54 |
| 32,768 | 380.06 | 380.66 | 20.43 | 20.41 |
| 65,536 | 362.99 | 366.68 | 20.18 | 20.21 |
| **mean (32 ctx)** | | **+0.58%** | | **+0.02%** |

**Decode: +0.02%** -- a null. The per-context deltas run -0.17% to +0.29% and only 16 of 32 are positive, which is a coin flip.

**Prefill: +0.58% as pooled, but that number is an artefact and should be read as a null too.** The whole difference comes from one sweep: the first main sweep runs ~0% off the second across the early frontiers and about +3.8% slow across the last six. That is a transient in the tail of one sweep, not a cold cache -- a cold cache would show up at the *start*. Excluding that sweep gives prefill +0.12% and decode -0.02%. Post-hoc exclusion is not evidence, so the honest statement is that this single block does not resolve GLM prefill; it is consistent with null, and there is no mechanism in this branch that could move GLM prefill in the first place.

This is the expected result and it is worth having on the record: every speed
change in this PR is in an FP8 KV / MXFP4 / E4M3 path that GLM 5.3 does not
use. Its `rope_dimension_count` is 0, so it takes a different KV path
entirely. The point of running it is to show the branch does not *regress* a
model it was never aimed at.

Two caveats on this block specifically. It is a single M-B-B-M block, so it is
position-balanced only to first order, unlike the DeepSeek result. And
`ds4-bench` defaults `--ctx-alloc` to `ctx-max + gen-tokens + 1` = 65665, which
is above the `DS4_GLM_METAL_LONG_CONTEXT_THRESHOLD` of 65536, so the whole
sweep runs with the full-attention cap clamped to 4096.

### Speed

| commit | effect |
|---|---|
| replace the E4M3FN binary search with branch-free rounding | **+1.37% decode** |
| reduce the FP8 KV amax with `simd_max` | **+1.27% decode** |
| restrict the matvec reduction tree to simdgroup 0 | **+0.47% decode** |

Combined, measured together against `main`: **+2.38% decode**, prefill flat.

The per-commit attributions above are from the original campaign against the
pre-rebase `main`; only the combined figure has been re-measured on this base.

### Correctness (no speed effect)

- **zero the FP8 amax scratch on partial KV blocks.** The fused
  norm+RoPE+FP8-store kernel dropped the `else` that seeds the reduction
  buffer, so a partial final block folded uninitialised threadgroup memory
  into the amax -- wrong *and* irreproducible between runs. Later subsumed by
  the `simd_max` rewrite; see *Notes on two commits*.
- **bound the IQ2_XXS grid staging to 64 threads.** Five kernels derived a
  staging position from `32*sgitg + tiisg` with no upper bound, correct only
  at `nsg == 2`. Safe today only because the host pins these pipelines to 2.
  Collapsed onto one shared helper so no open-coded copy is left -- that
  duplication is what produced the bug above.
- **assert the FP8 KV prefix is 64-aligned.** Six call sites across three
  backends assume `key_length - rope.dimension_count` is a multiple of 64;
  nothing said so.
- **clamp an out-of-range prefill chunk.** `--prefill-chunk 9216` is a
  documented flag that aborted with `gpu layer 2 attention batch encode
  failed` and no hint at the cause. The raw KV cache is sized
  `align_up(raw_window + prefill_cap, 256)` clamped to 8192 rows, so a larger
  chunk cannot be staged. Now clamps and warns. Inert for both defaults.

### Infrastructure

- **an E4M3FN conversion equivalence oracle.** 65536 inputs covering every
  representable magnitude, every midpoint and one ULP either side, the
  subnormal range, the saturation boundary and a full-range sweep. Confirmed
  to have teeth by mutation: breaking only the ties-to-even rule produces 130
  mismatches.
- **report GPU architecture, generation and core count.** Reporting only.
- **add `DS4_METAL_FORCE_METAL4`.** Off by default; makes the pre-M5 default a
  measured decision rather than an untested one.
- **a `speed-bench` CSV and plot for this hardware.**

# What did not work

Everything in this section was measured against the pre-rebase `main`. None of
it changes any code that ships in this PR, so the rebase does not affect the
conclusions, but the numbers have not been re-taken.

## 1. Raising the prefill chunk looked like a win and was not

This branch originally raised the Flash prefill chunk from 4096 to 8192 on M3
Ultra, behind a device gate, and claimed +1.95% prefill on the strength of a
single-frontier measurement at ctx 8192.

That claim was wrong, and the way it was wrong is the most useful thing in
this PR.

Measured properly -- the same branch against itself, gate on vs gate off, four
interleaved sweeps -- the chunk change **costs** prefill in the standard
benchmark workload:

| ctx | prefill Δ (gate on vs off) |
|---:|---:|
| 2,048 | **-2.11%** |
| 8,192 | -0.62% |
| 16,384 | -0.53% |
| 65,536 | -0.50% |
| **mean (32 ctx)** | **-0.62%** |

The reason is that the benchmark prefills in 2048-token increments, which is
below both the 4096 and the 8192 cap, so the larger chunk is never used --
but it still raises `raw_kv_rows` from 4352 to 8192, allocating roughly double
the raw KV staging workspace that this workload never touches. The cost is
largest at short context where that overhead dominates.

It does help when the cap actually binds, i.e. one large prompt prefilled in a
single call -- but the gain decays as the prompt outgrows the cap:

| single-frontier prefill | cap 4096 | cap 8192 | Δ |
|---:|---:|---:|---:|
| ctx 8,192 | 612.79 | 624.07 | +1.84% |
| ctx 16,384 | 588.84 | 596.08 | +1.23% |
| ctx 32,768 | 553.12 | 556.38 | +0.59% |

So it is workload-dependent: better for one-shot prefill, worse for
incremental prefill, which is also the chat/continuation pattern. It cannot be
made conditional cleanly, because `prefill_cap` is fixed at session creation
from `ctx_size`, before anything knows how the caller will prefill. And it adds
little that is not already reachable: `DS4_METAL_PREFILL_CHUNK=8192` already
exists, so the commit only changed the default.

**Dropped.** What survives is the useful part: the 8192 ceiling is now
documented in the source, and the crash it exposed is fixed (see *What
worked*).

## 2. Three changes were implemented, measured, and reverted

Each was verified bit-identical and benchmarked before being removed. None
regressed anything -- they simply were not improvements, and a tuning PR
should not carry changes that cannot be justified on measurement.

| change | measured | why reverted |
|---|---|---|
| Q8_0 decode weights as 16-bit pairs (7 sites) | +0.19% decode, n=12/arm, ranges overlap | at the noise floor |
| bitonic sort barriers narrowed to simdgroup scope | +0.08% decode, n=6/arm, ranges overlap | top-k is not a meaningful share of decode |
| IQ2_XXS SwiGLU epilogue kept in registers | +0.11% decode on the IQ2XXS model, n=4/arm, ranges overlap | IQ2-only, and noise there too |

Two are worth a specific note against the review that suggested them:

- The Q8_0 load widening was billed as *"the biggest single win"* available.
  It is not; on this hardware it is indistinguishable from noise.
- The IQ2_XXS epilogue fix cannot affect an MXFP4 or Q4_K deployment at all.
  Both dynamic-`float4` sites are in IQ2_XXS kernels; the MXFP4 kernels
  already use a scalar `simd_sum(sumf[row])` per row.

The bitonic change did at least surface a real bug in the advice: guarding on
`j < 32` alone is wrong, because `j == 1` is the last step of each `k` and the
next `k` opens with a cross-simdgroup step. The correct condition is
`j >= 32 || j == 1`.

## 3. There was nothing to retune

The premise going in was that thresholds tuned on a narrower M3 would be badly
wrong on an 80-core part. Tested directly: every one of the 24 `PRE_M5`
features is default-ON here, so each was disabled in turn, interleaved with
baseline so drift cancels.

| disabled | Δ decode |
|---|---|
| `DECODE_PORTS` | **-6.27%** |
| `MXFP4_MOE_DECODE_NSG1` | -2.28% |
| `MXFP4_MOE_DECODE_FIXED_ROUTE_PAIR` | -2.14% |
| `ROUTER_SIMD_FINALIZE` | -2.02% |
| `MXFP4_MOE_DECODE_TG_MULTIPLE` | -1.91% |
| `COMPRESSOR_RATIO4_DECODE_PACK_FUSION` | -1.70% |
| 17 more | -1.2% to -0.2% |
| `COMPRESSOR_EXACT_REDUCTION_FUSION` | +0.05% (noise) |

All 24 earn their keep. Four prefill features additionally cost 1.6-2.9%
prefill when disabled. `DS4_METAL_Q8_MV_NSG` is likewise already optimal at
its default of 4 -- raising it to 8 gains nothing, lowering it to 1 costs
1.18%.

The best MoE decode configuration is **`NSG1`**: *one* simdgroup, worth 2.28%
over the alternative. More GPU width did not want more parallelism per output
row; the winning shape is fewer simdgroups with more rows in flight. That
inverts the intuition the retune was based on.

**No threshold changes are included in this PR, because none were justified.**

Still 24 features on current `main` -- upstream added none in the 97 commits
this branch was rebased over.

**Scope note:** all 24 of those flags are on DeepSeek paths. `ds4_metal.m` has
939 GLM references and *zero* `PRE_M5` gates among them, so "there was nothing
to retune" is a statement about the DeepSeek path only. The GLM path has never
been through this exercise. See *Adjacent work this does not do*.

## 4. Metal 4 works on M3 Ultra and is slower

`supportsFamily:MTLGPUFamilyMetal4` is true here, the compile probe passes,
the pipelines build and run, and output is byte-identical to the default path.
It is simply not faster:

| | prefill | decode |
|---|---|---|
| default (off) | 622.94 | 39.512 |
| `DS4_METAL_FORCE_METAL4=1` | 620.8 | 39.530 |

-0.34% prefill, with the wider spread of the two arms. This matches MLX, which
gates its Neural Accelerator path at `gen >= 17` and would decline this
generation-15 device too.

It also settles a larger question without opening it: porting flash-attention
prefill onto the same MPP path is not worth attempting, because the MPP shape
does not win where it is already implemented.

The flag ships off by default. Its value is that the pre-M5 default is now a
measured decision rather than an untested one, and it can be re-measured on
future hardware.

## 5. Two directions ruled out before implementing

- **Dispatch overhead** (`NSString` per pipeline lookup, linear model-view
  scan, command-buffer stalls). Decode measures at **~98% GPU utilisation**
  (`DS4_METAL_GPU_BUSY_PROFILE`: 12468 ms busy over 12.71 s wall on a
  512-token decode run), so there is no CPU stall to reclaim. Hot pipeline
  lookups are also already cached at 93 call sites on current `main` (91 when
  this was first written; upstream added two). I had assumed the opposite
  before measuring and was wrong.
- **The barrier-free matvec shape selected by `out_dim`, and the `yl[32]` MoE
  staging rewrite.** The NSG results above show the current shape is already
  the better one for the Q8 path, and the `yl[32]` staging is IQ2_XXS-only.

## 6. Why the wins were where they were

Every measurable improvement came from removing a **barrier or a serial
dependency**. Every attempt to save arithmetic or narrow load-issue slots
measured as noise:

| change | Δ decode | verdict |
|---|---|---|
| E4M3 binary search -> branch-free | +1.37% | serial dependency chain removed |
| FP8 amax tree -> `simd_max` | +1.27% | 8 barriers per block -> 2 |
| matvec reduction -> simdgroup 0 | +0.47% | redundant reductions removed |
| Q8_0 byte loads -> 16-bit pairs | +0.19% | noise -- reverted |
| bitonic barriers narrowed | +0.08% | noise -- reverted |
| IQ2 epilogue spill removed | +0.11% | noise -- reverted |

That is consistent with the utilisation number: at 98% GPU-busy while moving
roughly a fifth of peak memory bandwidth on a bandwidth-bound operation, the
machine is busy waiting, not busy computing.

---

## How these numbers were measured

Worth stating separately, because it invalidated an earlier version of this
PR's headline claim and would invalidate anyone else's.

**A single unalternated sweep per arm is not reliable on this machine.** A full
sweep takes several minutes, so running `main` then the branch back to back
gives the two arms different page-cache warmth on a 145 GiB model, different
thermal state, and any background load in between. Measured that way, this
branch appeared to gain **+5-6% prefill and +6-7% decode** -- a flat offset
across all 32 contexts, which looks exactly like a real systematic improvement.

None of the prefill gain survived alternation, and the decode gain was roughly
halved.

Every number in this PR therefore comes from **interleaved** runs, and the
post-rebase measurement goes further: two blocks, **M-B-B-M** and **B-M-M-B**,
with a rebuild between every sweep and an untimed warmup sweep whenever the
model changes. Pooled across the two blocks each arm occupies run positions
{1,4} once and {2,3} once, which cancels position effects that a single ABBA
block cannot -- a drift curve that rises and then falls would otherwise flatter
whichever arm sits in the middle.

Within-arm run-to-run drift under that scheme was 0.32% on average (1.47% worst case), small enough to
resolve the ~2% effect here.

**Concurrent disk I/O invalidates a run completely.** One measurement round was
thrown away entirely because a large file copy was running against the same
disk. It produced a ~2% *uniform level shift* per sweep -- per-context shape
reproduced tightly (sd 0.26-0.30%) while the whole sweep moved -- which is
indistinguishable from a real effect if you only look at one block, and which
happened to inflate whichever arm ran in the middle. If you re-measure this PR,
check the machine is otherwise idle, and treat a per-sweep level shift larger
than a few tenths of a percent as a sign that it was not.

If you re-measure this PR, please alternate. A single before/after pair will
give you a number several times larger than the real one, in either direction.

## What was re-checked after the rebase

Re-verified on current `main`:

| | |
|---|---|
| End-to-end speed | re-measured, two position-balanced blocks |
| GLM 5.3 Flash Q4_K | newly measured as a control |
| Full test suite | 21 suites green (was 13 before upstream added GLM 5.3 and vision suites) |
| 24 `PRE_M5` features | still exactly 24; upstream added none |
| 64-alignment assert | safe for upstream's new `glm5-next` shape |
| Prefill-chunk clamp | inert for all five shipped defaults |
| Conflicts with #621 | re-measured against its current head |
| Cached pipeline call sites | 91 -> 93 |

Carried over from the original campaign and **not** re-measured: the
per-commit speed attributions, everything under *What did not work*, and the
98% GPU-utilisation figure. None of them gate any code in this PR.

## No new device gate

An earlier version of this branch added `ds4_gpu_device_is_m3_ultra()` to gate
the prefill-chunk change, which would have introduced a third
device-identification style to a file that already has two. That commit is
gone, and so is the helper.

What remains is reporting-only: `ds4_gpu_device_architecture()`,
`ds4_gpu_device_generation()` and `ds4_gpu_device_core_count()`. Nothing
dispatches on them. They exist because no existing gate can distinguish an
80-core M3 Ultra from a 10-core M3 -- `ds4_gpu_device_name_contains("M3")` is
equally true of both -- so any future width-dependent tuning has an input to
work from. `DS4_METAL_GPU_ARCH` and `DS4_METAL_GPU_CORES` override them so such
paths can be exercised on other hardware.

Given the retune found nothing to change (see above), landing them as
measurement rather than as a gate is the honest scope.

## Compatibility with #621

Re-verified against #621 at `75503b9`, its current head.

Note that #621 is **not** rebased onto current `main` -- it branches from
`3da2da9` -- so the two branches no longer share a merge base, which
they did when this section was first written.

**Three conflicts, all mechanical**, unchanged in shape from the original
analysis: two in `metal/moe.metal`, one in `ds4_metal.m`. `Makefile`, `ds4.c`
and `tests/ds4_test.c` merge clean.

**`metal/moe.metal`, conflict 1** -- #621 appends a 259-line
`ds4_metal_iq2xxs_half_values` table immediately after `#define iq2xxs_grid`;
this branch appends the staging helper at the same anchor. Both self-contained.
Keep both.

**`metal/moe.metal`, conflict 2** -- #621's `c230e3f` and this branch's IQ2
staging commit **fix the same bug the same way**, arrived at independently.
Their inline grid-stride block and this branch's helper call collide at one
kernel. Take the helper call: identical semantics, and it removes the
duplication rather than leaving two forms of the same fix in one file.

**`ds4_metal.m`** -- #621 adds `ds4_gpu_test_iq2_addr_mid_only_oracle` and this
branch adds `ds4_gpu_test_e4m3fn_dequant` at the same anchor. This one looks
like adjacency but is not: both functions end with the identical
`}` / `return 1;` / `}` tail, so git treats that tail as shared context and the
conflict cuts mid-function. Concatenating both sides nests one function inside
the other. #621's function needs its closing brace, `return 1;` and brace added
explicitly before the second begins.

### On the overlapping fix

#621 got to the IQ2 staging bug independently and their grid-stride form is
better than what this branch originally had, which bounded staging to a fixed
64 threads. That is correct for nsg >= 2 but under-stages at nsg == 1, leaving
the upper half of both tables unwritten. This branch now uses their shape,
credited in the commit message, applied through the shared helper so it covers
all sites rather than two.

### A note on the failure mode this class of merge has

`metal/*.metal` are concatenated and compiled **at runtime**, not at build
time. A shader signature mismatch introduced by a merge therefore passes
`make` cleanly and fails only when the Metal library loads -- CI that builds
but does not run the binary would go green on a broken tree. An earlier version
of this branch changed `helper_mv_reduce_and_write`'s signature and #621 adds a
call site to it, which produced exactly that. That signature change has since
been dropped (it bought nothing measurable), so the helper now matches `main`
and no call site is touched.

### One pre-existing failure in #621, unrelated to this branch

On a clean checkout of #621 at `bfd3554` with no merge involved,
`./ds4_test --metal-kernels` was red:

    ds4-test: output HC weights4 total mismatch=1/16 max_ulp=2
    tests/ds4_test.c:6658: assertion failed: stats.mismatch_count == 0

Identical signature before and after merging this branch, so it was theirs, not
merge damage. Flagging it so it is not mistaken for one later. This has **not**
been re-checked against #621's current head.

## Backend impact

`CONTRIBUTING.md` asks that PRs touching more than one backend be checked on
each. Nine of the eleven commits are Metal shaders, `ds4_metal.m`, tests or
benchmark data, and cannot affect another backend. Two touch shared code in
`ds4.c`, and both are no-ops on current behaviour:

- **64-alignment assert.** Runs on every backend. A no-op on all five shipped
  variants -- their non-RoPE prefix is 448 or 512 -- and only fires if a future
  variant breaks an invariant that six call sites across Metal, CUDA and ROCm
  already depend on silently. Upstream's new `glm5-next` shape is the case this
  was written for: it has `rope_dimension_count = 0`, giving a 512 prefix, so
  it passes. Confirmed against the shipped GLM-5.3-Flash GGUF metadata as well
  as the shape table.
- **prefill chunk clamp.** Only reachable when a caller requests a chunk above
  8192, which currently aborts the run. All shipped defaults are at or below
  the limit -- 4096, PRO's 8192, CUDA tensor-parallel's 2048, and upstream's
  new `DS4_GLM53_PREFILL_CHUNK_TOKENS` of 2048 -- so no configuration changes
  value.

No default prefill chunk is changed by this PR.

**I could not test CUDA or ROCm: I have no such hardware.** The reasoning above
is from reading the gates, not from running them. `make cpu` builds clean. A
maintainer with a CUDA machine should confirm `make cuda-regression` before
merging, though neither commit changes any value on that path.

## Adjacent work this does not do

Reading the GLM paths while re-validating this rebase turned up several
constants that take no input from how much memory the machine has. Recording
them here because they are the same shape as the prefill-chunk finding above,
and because the next person to look should not have to re-derive them. **None
of this is measured, and none of it is changed by this PR.**

- `glm_graph_full_attention_cap` gives the **SSD-streaming** path a full
  attention cap of 8192 and the **fully-resident** path 4096. The
  memory-constrained machine gets the larger window; the resident branch is
  the unexamined one. Above a 65536 context both clamp to 4096, so the
  asymmetry is only reachable below that.
- `DS4_GLM53_PREFILL_CHUNK_TOKENS` is a flat 2048 with no device, memory or
  residency input -- half what the GLM 5.2 path uses and a quarter of the 8192
  structural ceiling documented above.
- `DS4_GLM_METAL_INDEXED_PREFILL_SCORE_SCRATCH_MB` is a fixed 256 MB that
  directly clamps how many rows can be scored per dispatch.
- A trap for anyone testing the first two: `DS4_GLM53_PREFILL_CHUNK_TOKENS`
  (2048) and `DS4_GLM_METAL_FULL_ATTN_LAYER_FLUSH_CONTEXT` (2048) are the same
  number and the comparison is a strict `>`, so the default chunk sits exactly
  on the boundary where per-layer command-buffer flushing switches **off**.
  Raising the chunk to 4096 also switches flushing on across 46 layers. Those
  are two changes, not one, and measuring them together would likely produce
  the same false negative the DeepSeek prefill chunk did.

## Other corrections to the review that prompted this work

Beyond the ones above:

- **"Pre-scale the activation, don't unpack the weight ... the most valuable
  idea in the repo."** Already implemented: `kernel_mul_mv_q4_K_f32_impl`
  masks without shifting and folds the zero point into a `sumy` accumulator
  outside the inner loop.
- **Site counts.** The IQ2_XXS staging is 5 sites, not 6, and one already had
  the guard. The FP8 amax loop is 5 sites, not 3 (`dsv4_rope.metal` has two
  more). The Q8_0 scalar loop is 7 sites, not 5.
- **The 64-alignment assert cannot live in `config_validate_fixed_shape`.**
  `config_validate_model` returns early for `glm-dsa` and never reaches it, so
  the check sits at the shared entry point instead. Still true after upstream's
  restructuring of that function.

## Notes on two commits

**The E4M3 rewrite is simpler than the MLX port that was suggested.** E4M3FN
magnitudes are exactly `(8 + mant) * 2^(e-10)`, and the subnormal binade
continues the same `2^-9` spacing. So the rounded integer `q` is `8 + mant` and
`q * scale` is already the answer -- no code index, no encode step, no table
lookup. Since `code = (e-1)*8 + q`, code and `q` always share parity, so
`rint()` reproduces the ties-to-even-code rule exactly.

It was checked against the previous implementation over **every finite float32
bit pattern -- 4,278,190,080 values, 0 mismatches** -- before the shader was
touched, then confirmed on GPU by the new oracle.

**The `simd_max` rewrite subsumes the partial-block fix.** That bug existed
because a shared staging array's tail lanes went uninitialised on the final
block. Reducing the amax from a register removes the staging array altogether,
so the hazard is structurally impossible rather than guarded against. Both
commits are kept: the first is the minimal fix and explains the defect, the
second is the one that makes it unreachable.

## Verification

Per `CONTRIBUTING.md`, run at every commit:

```sh
make clean && make
make test
./ds4_test --all
./ds4_test --metal-kernels
make cpu        # then: rm -f ds4 ds4-server ds4-bench ds4-eval ds4-agent && make
```

All green on the rebased branch, zero warnings. `./ds4_test --all` is now **21
suites** (it was 13 before upstream added the GLM 5.3 and vision suites); three
skip for optional models this machine does not have
(`dspark-verify-depth`, `mtp-verify-depth`, `glm53-continued-prefill`).

Bit-exactness for each shader change was verified by running one
binary against the previous shader through the `DS4_METAL_*_SOURCE` overrides
(shaders are concatenated from `metal/*.metal` at runtime) and comparing greedy
output byte for byte. The override mechanism itself was confirmed by pointing
it at a deliberately broken file and observing the shader compile fail --
otherwise an A/B could silently compare a shader against itself.

## One thing a reviewer should know

`DS4_PREFILL_MAX_CHUNK` records a structural limit, not a tuned value: 8192 is
the largest chunk the graph can stage, because the raw KV cache is sized
`align_up(raw_window + prefill_cap, 256)` and clamped to 8192 rows. Verified
empirically -- 8192 runs, 8448 does not. Requests above it are now clamped with
a warning rather than aborting mid-encode.

## The remaining headroom is in neither review document

Decode is GPU-saturated (~98% busy) while moving roughly a quarter of the
achievable memory bandwidth on what is fundamentally a bandwidth operation.
That gap is the largest thing left on this hardware, and neither review
document covers it.

What it is *not*, as far as I can tell: a layout problem. Expert weights are
one tensor per layer -- `blk.N.ffn_gate_exps.weight` is `[4096, 2048, 256]` --
so each expert is a contiguous 4.46 MB slab. A token reads 18 such slabs per
layer (6 experts x gate/up/down), each a sequential multi-megabyte stream.
774 sequential reads per token is a well-behaved access pattern, so DRAM page
locality and TLB pressure are unlikely explanations, and dynamic gathering
into a contiguous staging buffer cannot help batch-1 decode at all: it would
read the same bytes, write them, then read them again.

The volume checks out -- 6 x 3 x 4.46 MB x 43 layers = 3.45 GB/token, about
150 GB/s at the measured decode rate -- but I have not measured what this
machine actually achieves for a sequential read, so "a quarter of achievable"
is inferred from spec rather than observed.

Three explanations remain and my data cannot separate them: the bandwidth
ceiling is lower than assumed, the expert access pattern costs more than its
shape suggests, or the MoE decode kernel is not bandwidth-saturating in the
first place. I would lean toward the third -- the winning decode configuration
is `NSG1`, which is a latency-oriented shape -- but that is a hypothesis, not
a finding.

The cheap way to settle it, before anyone builds anything: a microbenchmark
reading the same byte volume three ways -- purely sequential, in the expert
access pattern, and through the real kernel -- which separates ceiling from
pattern from kernel efficiency. If it turns out to be genuinely
bandwidth-bound, no kernel change helps and the lever is amortising the read
across more tokens, which this tree already has in MTP/speculative decode.


trueimage:
## Benchmark: `main` vs this branch

Apple M3 Ultra, 80 GPU cores, 512 GB, macOS 26.5.2, Metal backend.
DeepSeek-V4-Flash MXFP4 experts / F16 HC+compressor+indexer / Q8 attn+shared+out,
145.26 GiB fully resident, no SSD streaming. `main` = `c1d4597`.

```sh
./ds4-bench -m ds4flash.gguf \
```

Arms **interleaved** (M-B-B-M) with a rebuild between each, 2 full sweeps per
arm, means reported. Within-arm run-to-run drift was 0.05-0.34%.

| ctx | main prefill | branch prefill | Δ | main decode | branch decode | Δ |
|---:|---:|---:|---:|---:|---:|---:|
| 2,048 | 639.40 | 639.78 | +0.06% | 43.10 | 44.43 | +3.09% |
| 8,192 | 567.48 | 568.19 | +0.13% | 38.58 | 39.44 | +2.23% |
| 16,384 | 539.04 | 538.88 | −0.03% | 37.64 | 38.52 | +2.34% |
| 32,768 | 487.51 | 487.74 | +0.05% | 36.02 | 36.95 | +2.58% |
| 65,536 | 410.88 | 411.22 | +0.08% | 33.57 | 34.64 | +3.19% |
| **mean (32 contexts)** | | | **+0.06%** | | | **+2.42%** |

Decode gain is positive at every one of the 32 contexts, range +2.09% to
+3.04%. Prefill is flat: every win here is in the FP8 KV path, which is
decode-side work.

### Why alternation matters, and why this PR changed shape

Measured the obvious way — one sweep on `main`, then one on the branch — this
branch appeared to gain **+5-6% prefill and +6-7% decode**, a flat offset
across all 32 contexts that looks exactly like a real improvement.

Almost none of it was real. A sweep takes ~10 minutes, so back-to-back runs give
the two arms different page-cache warmth on a 145 GiB model and different
thermal state. Under alternation the prefill gain vanished entirely and the
decode gain roughly halved.

That mismeasurement had a concrete consequence. This PR originally raised the
prefill chunk from 4096 to 8192 and claimed +1.95% prefill. Measured properly —
the same branch against itself, gate on vs off, four interleaved sweeps — that
change **costs** 0.62% prefill in this workload, because the benchmark prefills
in 2048-token increments that never reach either cap while still allocating
double the raw KV staging rows. It helps only when a large prompt is prefilled
in a single call (+1.84% at ctx 8192, decaying to +0.59% at 32768). It has been
**dropped**; what survives is a fix for the crash it exposed, where
`--prefill-chunk 9216` aborted with `gpu layer 2 attention batch encode failed`
and no indication of the cause.

If you re-measure this PR, please alternate the arms. A single before/after
pair on this machine gives an error several times larger than the effect.

PR #878, opened 2026-08-27
docs: bring-up notes for AMD parts outside the Strix Halo matrix

dongioia:
This one is separate from #875, #876 and #877, and safe to close on its own: it adds a file and changes no code.

STRIXHALO.md covers Ubuntu, GRUB and gfx1151. Bringing the ROCm backend up on a Radeon 780M under CachyOS needed three things that are not in it, and none of the three failures points at its own cause:

1. `hipcc` calls `clang++` without the ROCm include directory, so every HIP source dies on `hip/hip_runtime.h` even though the header is sitting in `/opt/rocm/include`. `--rocm-path=/opt/rocm` fixes it.
2. rocWMMA 2.2.0 enumerates eleven architectures and gfx1103 is not one of them, although it has the same RDNA3 WMMA instructions as gfx1102. Aliasing the two is enough, and the AUR already carries a `rocwmma-gfx1103` package for this.
3. The GTT aperture comes from `LINUX_OPTIONS` in `sdboot-manage.conf` rather than grub, and `amdgpu.gttsize` is in MiB while `ttm.pages_limit` is in 4 KiB pages, so the two have to be worked out separately for the same size.

One thing in there is worth knowing even on a Strix Halo: override `HIPCC`, not `ROCM_CFLAGS`. `ROCM_CFLAGS` carries `--offload-arch=$(ROCM_ARCH)`, so replacing it wholesale makes `ROCM_ARCH` inert and pins the architecture to whatever string you copied. I did exactly that for an evening before noticing.

What I verified on this machine: `make rocm` builds, `make test-mxfp4-rocm` passes, `make cpu` still builds. What I did not: `make test`, which wants a model file. No published GGUF fits 58 GiB resident, so nothing has run end to end here, and the file says so in its first paragraph rather than reading like a support claim.

On provenance: the machine, the failures and the measurements are mine, and I used a coding agent to read the tree and cross-check what the file claims.

If you would rather not carry a file for a part outside your QA matrix, I understand. In that case tell me and I will move it to the ROCm discussion issue instead.

### DIFF
--- a/AMD-NON-STRIX.md
+++ b/AMD-NON-STRIX.md
@@ -1,0 +1,199 @@
+# DS4 on AMD GPUs other than Strix Halo
+
+[STRIXHALO.md](STRIXHALO.md) covers Ubuntu, GRUB and `gfx1151` with 128 GB of
+RAM. This file covers what changes on an Arch-family distribution with
+systemd-boot and a smaller RDNA3 part. The reference machine is a Radeon 780M
+(`gfx1103`, Ryzen 7 8845HS) with 58 GiB of usable system RAM, ROCm 7.2.4 and
+kernel 7.2.0.
+
+What was verified here: `make rocm` builds, `make test-mxfp4-rocm` passes, and
+`make cpu` still builds. What was not: `make test`, which needs a model file.
+No supported model has been run end to end on this machine, because none of the
+published GGUFs fit in 58 GiB resident. The build carries one pre-existing
+warning in `ds4_server.c`, unrelated to anything described here.
+
+## 1. ROCm packages
+
+On Arch and derivatives the SDK is one group, and rocWMMA is a separate
+package:
+
+```sh
+sudo pacman -S --needed rocm-hip-sdk rocwmma rocminfo
+```
+
+Add yourself to the `render` and `video` groups as in STRIXHALO.md.
+
+## 2. hipcc does not find its own headers
+
+On this setup `hipcc` invokes `clang++` without adding the ROCm include
+directory, so any HIP source fails at the first include:
+
+```
+fatal error: 'hip/hip_runtime.h' file not found
+```
+
+`/opt/rocm/include/hip/hip_runtime.h` is present; the driver simply does not
+pass it. `--rocm-path=/opt/rocm` fixes it. Section 4 shows where to put it.
+
+## 3. rocWMMA does not know gfx1103
+
+`rocwmma/internal/config.hpp` enumerates the architectures it supports and
+fails the build on anything else:
+
+```
+config.hpp:79:15: error: static assertion failed: Unsupported architecture
+```
+
+In rocWMMA 2.2.0, shipped with ROCm 7.2.4, that list is `gfx908`, `gfx90a`,
+`gfx942`, `gfx950`, `gfx1100`, `gfx1101`, `gfx1102`, `gfx1150`, `gfx1151`,
+`gfx1200`, `gfx1201`. `gfx1103` is absent although it implements the same
+RDNA3 WMMA instructions as `gfx1102`. Reproduce with:
+
+```sh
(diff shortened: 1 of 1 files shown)


jhohertz:
Just noting you can run it today on a 64GB 780M system on main via ssd-streaming.  You just need to fake the rocm out a bit (gfx1100 vs gfx1103) and tune a couple of values.

Build as:

```
ROCM_ARCH=gfx1100 make rocm -j
```

Run like:

```
HSA_OVERRIDE_GFX_VERSION=11.0.0 DS4_ROCM_STREAM_FREE_RESERVE_GB=2 DS4_ROCM_Q8_F16_CACHE_MB=0  ./ds4 --ssd-streaming -m DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix-0731.gguf -p "Please write me a single-page HTML flappy bird game."
```

You might not see output properly unless you find and change the hardcoded flag for debug output on model loading.

It ain't fast, but it does in fact run.

dongioia:
Thanks, I would not have tried it otherwise. It does run here.

One thing I can add: the `gfx1100` substitution is not needed to get a run. My
build targets `gfx1103` natively and Flash answers prompts with no
`HSA_OVERRIDE_GFX_VERSION` set at all. ROCm 7.2.4 knows the part. The only thing
I had to patch was rocWMMA, whose architecture list skips `gfx1103` even though
it implements the same RDNA3 WMMA instructions as `gfx1102`.

```text
Radeon 780M (gfx1103), Ryzen 7 8845HS, 58 GiB RAM, 4 GiB VRAM, 46 GiB GTT
CachyOS, kernel 7.2.0, ROCm 7.2.4
DeepSeek-V4-Flash-IQ2XXS-...-imatrix-0731.gguf, 80.76 GiB, on NVMe
DS4_ROCM_STREAM_FREE_RESERVE_GB=2 ./ds4 --ssd-streaming -m ./ds4flash.gguf -p ...
```

The startup accounting is what explains the fit. Non-routed weights are the part
that cannot be streamed:

```text
non-routed weights: 8.20 GiB
routed expert size: 6.75 MiB
total expert budget 28.60 GiB = 3.38 GiB prefill headroom
                              + 25.23 GiB dynamic cache (3827 experts)
```

Your substitution is still worth doing, because it is faster. I built the tree
twice, kept the first binary, and alternated them. A single run per build had
told me nothing useful: my first pair suggested a 41% gap that turned out to be
noise. Decode in t/s, with the execution order inside each pair noted:

| | gfx1103, no override | gfx1100 + HSA_OVERRIDE_GFX_VERSION=11.0.0 |
|---|---|---|
| gfx1103 first | 2.19 · 2.50 · 4.48 | 3.00 · 4.18 · (run lost) |
| gfx1100 first | 2.48 · 2.41 · 2.63 | 2.84 · 3.04 · 2.72 |

Yours wins in both orders, around 2.9 against 2.45 t/s. I cannot account for the
two 4.x readings, one per architecture. Prefill sat at 0.33 t/s throughout.

Where I did worse than "it ain't fast" is output quality on long generations.
Short factual prompts come back correct on both builds. Your flappy bird prompt
does not. The `gfx1100` build writes valid HTML, CSS and JavaScript, then repeats
one block of variable declarations until the context ends. The `gfx1103` build
degrades harder, through broken grammar down to single letters. That is one run
per build, so I am not claiming the difference in severity is real, and I have no
way here to tell a 2-bit quantization effect apart from a sampling one.

Does your gfx1100 run actually finish that prompt with a playable game, or did
you stop at "it runs"? If yours completes it, the problem is on my side and I
would like to find it.

I have written this up in AMD-NON-STRIX.md on this PR and credited you for the
idea. Happy to cut it down or drop it if it does not belong there.

PR #882, opened 2026-08-27
kvstore: reject disk KV checkpoints written under different model weights (#805)

### LINKED ISSUE #805: Disk KV cache can be reused across different model weights with the same model_id
I found a potentially unsafe disk KV cache reuse case when switching between different GGUF weights that share the same DS4 model structure.

### Setup

- macOS / Metal
- DeepSeek V4 Flash 0731 official GGUF
- Huihui DeepSeek V4 Flash 0731 variant
- Both launched with the same `--kv-disk-dir`
- Same context / prompt format / quantization-compatible cache settings

### Observed behavior

After running the official model and creating an 81,920-token disk checkpoint, I stopped the server, switched `ds4flash.gguf` to the Huihui model, and restarted DS4 using the same KV cache directory.

The Huihui model immediately accepted the checkpoint created by the official model:

```text
kv cache hit text tokens=81920 ... quant=2 ...
```

So a checkpoint produced using one set of model weights can apparently be restored while another set of weights is loaded, as long as the model structure / model_id and other cache compatibility checks match.

### Concern

KV states depend on the actual model weights. Two GGUFs can have the same architecture, tokenizer and DS4 `model_id` while containing different weights, such as fine-tuned or abliterated variants.

In that case the cache can be structurally compatible while semantically invalid, potentially causing silent output corruption rather than an explicit error.

Cross-quant reuse for different quantizations of the same underlying model makes sense, but it seems useful to distinguish that from reuse across genuinely different model weights.

### Possible fix

Could disk KV metadata include a stable model-weight fingerprint, or another model identity derived from the GGUF, and require it to match on restore?

Ideally the identity would remain compatible across supported quantizations of the same underlying weights if cross-quant reuse is intentional.

### Workaround

For now I am using separate cache directories:

```text
/tmp/ds4-kv-official
/tmp/ds4-kv-huihui
```

I can provide additional trace output or test a proposed fix if useful.

tedin7:
Fixes #805.

## What

`ds4_engine_model_id()` returns the compile-time model variant, so the 1-byte `model_id` in the disk KV cache header identifies the model *shape*, not the weights: two GGUFs with the same structure — the issue's official-vs-Huihui case, or any requant — share it, and a checkpoint written under one set of weights is silently restored under another.

Following the header's own compatibility pattern:

- **`ds4_engine_weights_fp24()`** — a 24-bit FNV-1a fingerprint of the loaded weights: file size, tensor count, and the head of *every* tensor (so no single-tensor edit slips between samples). A few hundred 64-byte `pread`s, once per engine, off the hot path; reads via the descriptor so it also works under SSD streaming where the mapping is partial on purpose.
- Stored in the **free header bytes 21..23** (no version bump, no layout change).
- Every restore path rejects on mismatch with a clear reason (`cached checkpoint was written for different model weights`): `ds4_kvstore_try_load_text`, the agent checkpoint restore, the server tool-map scan, prefix lookup, and the continued-supersede check.
- **Headers written before the fingerprint carry 0 there and stay accepted**, mirroring the `model_id` 0 rule documented in `ds4_kvstore.h` — existing caches keep working, protection applies to checkpoints written from now on.

## Limits (stated, not hidden)

- 24 bits: a 1-in-16M chance that two different weight sets collide, versus the current 100%. If a full-width fingerprint is preferred, it needs a `KV_CACHE_VERSION` bump and a wider field — happy to do that variant instead if you'd rather.
- Legacy (fingerprint-0) checkpoints remain loadable by design; a stricter mode could refuse them, but that would invalidate every existing cache on upgrade.

## Validation (per CONTRIBUTING.md)

- Machine: Linux x86_64, CPU build (`make cpu`, `-DDS4_NO_GPU`), commit `c1d4597`. This change touches no inference path — header bytes, lookup filters and one sampling pass at engine init — so no backend regression surface; `make cpu` builds warning-free.
- New regression tests in the server unit group (`./ds4_test --server`, all green):
  - `test_weights_fp24_of_fd_tells_weights_apart`: two files identical except one tensor byte → different fingerprints; same bytes → same fingerprint (the issue's scenario in miniature);
  - `test_kv_cache_lookup_rejects_wrong_weights`: lookup skips a checkpoint fingerprinted under other weights, still returns its own, still accepts a legacy 0-fingerprint entry.
- Mutation check: breaking the header parse on purpose turns the new test red.
- Not run here: `--logprob-vectors` / `--long-context` (need the model GGUF; no weights on this machine). The change does not touch tokenization, attention or logits.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -36856,6 +36856,8 @@ typedef struct {
 struct ds4_engine {
     ds4_model model;
     ds4_model mtp_model;
+    /* Lazily computed by ds4_engine_weights_fp24(); 0 = not computed yet. */
+    uint32_t weights_fp24;
     ds4_vocab vocab;
     ds4_weights weights;
     ds4_mtp_weights mtp_weights;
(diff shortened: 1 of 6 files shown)


tedin7:
Amended in 95ee00d: the first version compared fingerprints across quantizations too, which would have vetoed the supported cross-quant checkpoint reuse (`reject_different_quant` exists precisely to govern that case). Fingerprints of two quants differ by construction, so the check is now scoped to matching `quant_bits` — where a mismatch really means different weights, which is this issue's case. Across quants the existing policy remains the only gate; a regression case covers it.

Issue #836, opened 2026-08-20
ds4 server memory leak?


ruanshudong:
It looks like the ds4 server has a memory leak? Anyway, after leaving it untouched all night, it got OOM-killed by the system.

Lamerone-net:
I've been using DS4 for a few months on three NVIDIA Spark servers, and I've never had any memory issues, crashes, or anything else.
What are your server's specs?

ruanshudong:
Virtual machine device information: 2*32-core CPU, 400G memory, 8*4090 (24G), kv-cache 320G. I don't know why, but after starting ds4-server, even if no access is made, after a few hours, the server hangs, and I cannot SSH into it. The information I see is that ds4-server was killed due to OOM, and at the same time, sshd was also terminated by the system, making it impossible to SSH login

<img width="1914" height="904" alt="Image" src="https://github.com/user-attachments/assets/f45134f9-4f71-4576-9fb7-2b68e54796fd" />

Lamerone-net:
If I remember correctly, the process terminated is the one using the most memory, but it isn't necessarily the one with the problem.
ds4 is the best candidate to kill to "fix" the OOM problem becasue use a lot of memory.

try this:
1) reboot
2) free -h
3) swapon --show
4) df -h
5) df -i
6) systemctl status ds4-server

To find out which process has used up all the RAM, try
journalctl -b -1 --no-pager | grep -B20 -A30 'Out of memory'

tedin7:
Independent 4-hour measurement: no idle leak on a single-GPU CUDA setup. Details per CONTRIBUTING.md below.

**Machine/backend**: RunPod community pod, RTX 3090 24 GB (sm_86), driver 580.65.06, 251 GB RAM, CUDA 12.4 toolkit, `make cuda CUDA_ARCH=sm_86`, ds4 commit `c1d4597`.

**Model quant**: `ds4f-q2` (DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix-0731, 81 GB).

**Commands**:

```sh
./ds4-server -m ./ds4flash.gguf --cuda --ssd-streaming --ctx 8192 --port 8080
# 3 warmup chat completions, then sampled every 60 s from /proc/<pid>/status,
# /proc/meminfo and nvidia-smi: 3.49 h idle, then 0.48 h with one
# chat completion every 5 min.
```

**Result**: server RSS flat to the byte across the whole run.

[...]

PR #883, opened 2026-08-27
engine: clearer refusal when --ssd-streaming meets a single-GPU --gpu-vram budget (#880)

### LINKED ISSUE #880: CUDA single GPU: --gpu-vram + --ssd-streaming refused as "multi-GPU placement" (misleading message)
## Summary

On a **single** CUDA GPU, passing an explicit `--gpu-vram N` together with `--ssd-streaming` refuses to start with:

```
ds4: --ssd-streaming is not compatible with multi-GPU placement
```

There is no multi-GPU placement — one device, one budget. The message sent us down the wrong path twice before we found the working invocation (drop `--gpu-vram`, streaming manages VRAM itself).

## Environment

- RunPod pod, 1× RTX 3090 24 GB (sm_86), driver 580.65.06, 251 GB RAM
- `make cuda CUDA_ARCH=sm_86`, commit `c1d4597`
- Model: `ds4f-q2` (81 GB)

## Repro

```sh
./ds4-server -m ./ds4flash.gguf --cuda --ssd-streaming --gpu-vram 21 --ctx 8192   # refused
./ds4-server -m ./ds4flash.gguf --cuda --ssd-streaming --ctx 8192                 # works
```

## Where it comes from (reading the source)

With an explicit budget, `engine_classify_multi_tier()` runs the placement: an 81 GB model against a 21 GB budget classifies most layers as CPU-spill entries, the placement becomes tiered, and `e->multi_tier` turns on with a single GPU. The guard at `ds4.c:57748` then fires, but its message names multi-GPU rather than the real condition (explicit budget produced a tiered placement that streaming does not support).

## Suggested fix (either is fine)

- Message: say what happened — e.g. `--ssd-streaming does not support tiered placement; with an explicit --gpu-vram budget smaller than the model, layers spill to CPU. Drop --gpu-vram to let streaming manage VRAM.`
- Or behavior: with `n_gpus == 1` and streaming requested, ignore the explicit budget (with a log line) instead of refusing.

Happy to send a small PR for the message variant if useful.

tedin7:
Fixes #880.

With one CUDA device, an explicit `--gpu-vram` budget smaller than the model classifies the remaining layers as CPU spill, the placement becomes multi-tier, and the streaming guard refuses with *"not compatible with multi-GPU placement"* — on a machine with a single GPU. That message sent us hunting for a phantom second device (repro in #880: two commands, one flag apart).

Message-only change, per the issue's first suggested fix: when `gpu_cfg->n_gpus == 1` the refusal now names the real condition and the way out (drop `--gpu-vram\`, streaming manages VRAM itself). Actual multi-GPU placements keep the existing wording. No behavior change — the guard still refuses exactly the same configurations.

Validation: `make cpu` builds warning-free on commit `c1d4597`; the touched code is a refusal path with no inference surface. Verified the exact refusal text on an RTX 3090 pod while working #836 (that is how the issue was found).

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -57746,8 +57746,20 @@ static int ds4_engine_open_internal(ds4_engine **out,
         return 1;
     }
     if (e->ssd_streaming && e->multi_tier) {
-        fprintf(stderr,
-                "ds4: --ssd-streaming is not compatible with multi-GPU placement\n");
+        /* An explicit --gpu-vram budget smaller than the model classifies
+         * the remaining layers as CPU spill, which makes even a single-GPU
+         * placement multi-tier: on that path the old "multi-GPU" wording
+         * sent people hunting for a second device (issue #880). */
+        if (gpu_cfg && gpu_cfg->n_gpus == 1) {
+            fprintf(stderr,
+                    "ds4: --ssd-streaming does not support tiered placement: "
+                    "the explicit --gpu-vram budget is smaller than the model, "
+                    "so layers would spill to CPU. Drop --gpu-vram and let "
+                    "streaming manage VRAM itself.\n");
+        } else {
+            fprintf(stderr,
+                    "ds4: --ssd-streaming is not compatible with multi-GPU placement\n");
+        }
         ds4_engine_close(e);
         *out = NULL;
         return 1;


OPS-NeoRetro (approved):
APPROVED! 👍

OPS-NeoRetro:
@antirez, please merge this ASAP though, take a look at this sooner or later, just use your own DS4 for a bit, maybe take a look at #621 and merge it if it is great

PR #886, opened 2026-08-28
cli: add --version to every binary

evandhoffman:
There is currently no way to ask a built binary what it is. The five tools report nothing identifying, so a binary that has outlived its checkout — or was copied to another machine — cannot be matched back to a commit.

This matters for anyone keeping benchmark results: a number is only interpretable if the build that produced it can be identified later, and `ds4-server`'s mtime is a poor substitute.

Adds `--version` to `ds4`, `ds4-server`, `ds4-bench`, `ds4-eval` and `ds4-agent`. The string is computed **once** in the Makefile:

```make
DS4_VERSION ?= $(shell git describe --tags --always --dirty 2>/dev/null || echo unknown)
```

and passed to each translation unit that needs it, so every tool from a single build reports the same value:

```
$ ./ds4-server --version
ds4-server c1d4597-dirty
$ ./ds4-agent --version
ds4-agent c1d4597-dirty
```

An `#ifndef DS4_VERSION` fallback keeps the files compiling standalone if the define is absent, and `?=` means a packager can override it (`make DS4_VERSION=1.2.3`).

The flag is handled next to `--help`, before any other parsing, so it works without a model or a valid configuration.

All five binaries build clean with no new warnings on macOS/Metal.

Independent of #885, which touches the same file but a different function.


simoltos:
This was intentional. I don't know if it's the moment for a version

PR #861, opened 2026-08-24
Consolidate ROCm exactness, distributed v3, DSpark speculation, and NHI tensor parallelism for Strix Halo

jyatesdotdev:
## What this does

This is the **single consolidated PR** for the Strix Halo ROCm/distributed work: the exact-kernel and protocol-v3 groundwork, distributed DSpark speculation, the complete **two-node tensor parallelism over Thunderbolt/USB4 NHI** implementation, and the start of **multi-session batched serving** for the layer-split pipeline deployment.

**Rebased onto** antirez/ds4 `main` @ `9139e2a` (Qwen3.8 Flash Next, CUDA SSD, n-gram packaging). Branch head **`5d3801b`** = the 18-commit rebase + one new commit `server: allow configuring exclusive prefill quantum`. Conflicts resolved as unions in `.gitignore`, `Makefile`, `ds4.c`, `ds4.h`, `ds4_kvstore.h`; range-diff 18=18, no extra commits. Validation: Mac Metal clean build 0 warnings + full model-free `make test` set green; gfx1151 Strix Halo pair `make strix-halo -j8` 0 warnings, both nodes byte-identical; `test-mxfp4-rocm`, `test-rocm-qkv-fusion`, `test-q8-krow-rocm`, `test-tp-combine-rocm`, `test-graph-deferred-dump-rocm` pass. ROCm/CUDA/NHI/perf changes since the last banner: new `--prefill-quantum N` flag (exclusive-prefill slice, default 2048) — production now runs `--prefill-quantum 8192` so the `--dist-prefill-window 3` pipeline stays full; measured single-client cold ~16k prefill 188.7 → 238.3 tok/s (+26%) on the pair. Decode unchanged (~15 tok/s; speculative-decode A/B showed MTP/DSpark provide no gain as shipped). Production redeployed (MXFP4, `--batched-session 2`, kv-disk) with the flag. `make test-rocm` still exits nonzero solely on the Linux-generic upstream `tests/test_tp_tcp` failure (abort/hang, not ROCm-specific; reported upstream).

[...]

PR #887, opened 2026-08-28
 ROCm: optimize DeepSeek V4 Flash prefill and DSpark on gfx1151

kyuz0:
This PR adds gfx1151 performance optimizations for DeepSeek V4 Flash IQ2 inference and DSpark, primarily targeting prefill and speculative decoding. The optimized paths are selected automatically; no performance environment variables are required.

## Performance

Measured with ROCm 10.0 and rocBLAS `5.6.0.8d1ae90e` on a Framework Desktop with Ryzen AI Max+ 395, 128 GB unified memory, up to 124 GB iGPU GTT, and `DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix-0731.gguf`.

| Test | Before | After |
|---|---:|---:|
| 4,096-token prefill | 187.26 tok/s | 292.86 tok/s |
| ROCm 10.0 DSpark 4,096-token prefill | 153.52 tok/s | 293.09 tok/s |
| DSpark, 50 generated tokens on a short C coding prompt with high proposal acceptance | 15.23-15.24 tok/s | 24.83-24.86 tok/s |
| DSpark default scheduler, 2,048-token source-code prompt, 128 generated tokens | 15.57 tok/s | 20.58 tok/s |
| Ordinary decode | 16.87 tok/s | 16.87 tok/s |

The source-code workload is a plain-text prompt containing an excerpt of LLVM 22.1.6 source files. LLVM is not built or executed; its source is used only as representative coding context for speculative decoding.

## Main changes

- Routed-MoE prefill uses the four-wave wave32 `x64/y64` IQ2 MMQ tile selected by the production-shape sweep. In the isolated 2,048-token gate/up test it takes 34.435 ms versus 51.586 ms for the `x128/y128` reference, a 33.2% reduction. The launch bounds each expert by `n_tokens` instead of `n_tokens * top_k`, and larger IQ2 batches are processed in bounded 2,048-row tiles.
- Prefill attention uses 32-head wave32 rocWMMA kernels and an 80-key online-attention tile; the indexed path also vectorizes F32-to-F16 KV staging. A separate `4096 x 2048 x 8192` F16 rocWMMA kernel handles the large attention-output-B projection, reducing that live call from 25.89 to 11.88 ms and producing an 8.9% end-to-end gain at warm 4,096-token prefill within the tuning stack.
- Other retained prefill paths use output-shape-specific Q8 row tiles, vectorized Q8 activation staging, native-F16 rocWMMA kernels for the measured M=24, M=64/256/512/1024 and M=8192 projections, and cooperative Q2_K weight staging once per 256-value K slab. These are exact production-shape dispatches; unmatched shapes keep their existing paths.
- Greedy DSpark drafts five tokens from the previous target hidden state, prepends the current target token and verifies all six rows in one target batch. Full and partial accepts commit the captured verifier prefix directly. When five rows are accepted before the sixth rejects, DS4 restores the four-row prefix and evaluates only the fifth accepted row instead of replaying all five.
- The gfx1151 target verifier uses the native tile-4 IQ2 gate/up kernel, one-wave Q8_1 MMVQ for dense projections, eight independent Q8_1 MMVQs for grouped attention-output A, and a fused 128-thread Q2_K down-plus-sum kernel. The dense-Q8, grouped-attention and fused-down specializations are gated to target verification because extending them to the five-row support-model stages was slower. The proposal stage separately moves confidence-zero greedy Markov-plus-argmax to the GPU and skips confidence scoring when the configured threshold is zero.
- rocBLAS solution indices are selected only for known library builds and exact shapes. rocBLAS `5.5.0.cd957402` in ROCm 7.14 uses `-217/-216`; `5.6.0.8d1ae90e` in ROCm 10.0 uses `-50/-49`, plus `-401` for the N=5/N=6 `4096 x N x 8192` attention-B projection. `-401` is 1.30x faster in isolation and cuts verifier time per cycle by about 14%. Unknown builds, unsupported shapes and rejected solutions fall back silently to the default rocBLAS path. Matching performance on another rocBLAS build may require retuning these solution choices.
- The gfx1151 scheduler observes 16 speculative cycles because the measured workload begins with a low-acceptance phase before reaching repeated full accepts. It then bypasses DSpark for the remainder of a low-acceptance request, while profitable requests remain speculative. This policy is automatic and independent of the optional rocBLAS solution override; other backends retain their existing four-cycle behavior.
- Loading the DSpark support model no longer evicts the resident target weight cache, removing the roughly 2x DSpark prefill regression. `ds4-bench` now accepts the DSpark support model and restores local sweep frontiers from session snapshots so prefill intervals are measured without untimed prefix replay.

## Validation

The final ROCm 10.0 build reaches 20.58 tok/s aggregate and 21.53 tok/s steady while generating 128 tokens from that fixed 2,048-token source-code prompt, with 270.98 tok/s prefill in the same process. The verifier-depth test commits 128 speculative tokens at draft depth five with zero argmax gap. Deterministic C, mathematics, Python and prose prompts exercise full, partial and one-row prefix-extension commits with zero verifier errors. All five binaries build and pass smoke tests.

Correctness was checked with `score_official` built from each exact DS4 revision, using the official 100-case Flash manifest at context 4096. Every run completed all 100 cases and 2,313 target tokens with no incomplete or non-finite result.

| Platform and revision | Average NLL | First-token matches | Average greedy prefix |
|---|---:|---:|---:|
| ROCm 10.0, pre-optimization baseline | 0.402107528 | 55/100 | 4.650 |
| ROCm 10.0, this PR | 0.403662338 | 55/100 | 4.690 |
| CUDA 13.0 on NVIDIA DGX Spark, pre-optimization baseline | 0.404811251 | 55/100 | 5.150 |
| CUDA 13.0 on NVIDIA DGX Spark, this PR | 0.404811251 | 55/100 | 5.150 |

The pre-optimization baseline is commit `c1d4597a80e300b803dc642519718f2c999589da`, the exact parent of the first performance commit. On ROCm, average NLL changes by +0.387%, API top-1 agreement changes from 85.863% to 85.949%, and pair-order agreement changes from 98.665% to 98.686%. On CUDA 13.0, the baseline and this PR produce byte-identical scorer TSVs, confirming that the changes do not alter the CUDA arithmetic path.

Terminal Bench Mini is currently running against this PR. Its results will be added when complete; this is the long-horizon check for whether the small ROCm scorer divergence still preserves quality on coding and agentic tasks.

Ordinary SSD streaming remains supported and passes a 512-token tuned-IQ2 prefill run. DSpark cannot currently be combined with SSD streaming.


iammac2:
Hope this is in some way useful:


# PR #887 on ROCm / gfx1151 — measured against its own merge-base

**PR:** [antirez/ds4#887](https://github.com/antirez/ds4/pull/887) —
*ROCm: optimize DeepSeek V4 Flash prefill and DSpark on gfx1151* (@kyuz0)
**Tested:** head `9254c3daf40bc2bcfada6ce4bbd74215cd69ad20`
**Baseline:** merge-base `8db89fe083ae4d17c9a2428ccd29803d3ae8f577`
**Date:** 2026-08-30
**Tester:** @iammac2

**Headline: +45.8% prefill over its own merge-base, decode unchanged, no hangs.**

---

## 1. Method

Both arms built from source at **pinned commits**, in containers derived from @kyuz0's own
`Dockerfile.rocm-10.0` — identical ROCm 10.0 SDK, identical build flags
(`make strix-halo ROCM_ARCH=gfx1151`, `-fPIC`, no other overrides). **Only the git commit
differs**, so the delta is attributable to the PR and nothing else.

| | |
|---|---|
| Machine | Framework Desktop, Ryzen AI Max+ 395, Radeon 8060S (**gfx1151**), 128 GB, Fedora 44 |
| Runtime | ROCm 10.0 (`stable.repo.amd.com` rhel10 packages, `amdrocm-core-devel10.0-gfx1151`) |
| Host | Fedora ROCm 7.1.1, **untouched** — all runs containerised via podman/toolbox |
| Model | `DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix-0731.gguf` (80.76 GiB), resident |
| Sweep | `ds4-bench --ctx-start 2048 --ctx-max 8192 --step-incr 2048 --gen-tokens 128`, `promessi_sposi.txt` |
| Protocol | discarded warmup per arm, 45 s cooldowns, machine otherwise idle |

Both arms compiled with **0 errors**.

## 2. Result — Q8

| ctx | base `8db89fe` prefill | **#887 prefill** | Δ | base decode | #887 decode | Δ |
|---:|---:|---:|---:|---:|---:|---:|
| 2048 | 153.09 | 191.53 | +25.1% | 15.70 | 15.71 | +0.06% |
| 4096 | 201.15 | **293.89** | **+46.1%** | 15.01 | 15.02 | +0.07% |
| 6144 | 197.86 | **288.93** | **+46.0%** | 14.97 | 14.98 | +0.07% |
| 8192 | 196.71 | **285.91** | **+45.3%** | 14.82 | 14.82 | 0.00% |

**Prefill +45.8% (4K–8K mean). Decode +0.05%, i.e. unchanged.**

The three deep frontiers agree to within 0.8 points (46.1 / 46.0 / 45.3), so this is a
stable effect rather than a lucky sample.

Ignore the +25.1% at 2048 — that frontier is warmup-contaminated on both arms (each sits
well below its own 4096 value), a consistent artefact of this sweep on this machine.
**Quote 4K–8K.**

Warmup runs reproduced the measured ones closely (#887 warmup: 289.73 @6144, 286.44 @8192),
so the numbers are repeatable.

## 3. Stability

No `GPU Hang`, no `HW Exception`, no aborts across five runs (two warmups, two measured Q8,
one Q4). 

This does **not** establish that the ROCm path is stable. @kyuz0 has noted a transient
hang under load with certain shapes that is still unfound, and by description it is
intermittent and load-dependent — a clean sweep cannot disprove it.

## 4. Limitation — #887 cannot run AProjQ4 GGUFs

At the pinned head, the AProjQ4 model fails at the first layer on every frontier:

```
ds4: gpu layer 0 raw KV batch store failed
ds4: gpu layer 0 attention batch encode failed
ds4: gpu whole-prefill layer 0 encode failed
ds4-bench: prefill to 2048 failed: rocm prefill failed
```

Model: `...-AProjQ4-SExpQ8-OutQ8-chat-v2-imatrix-0731.gguf` (78.62 GiB, HF `refs/pr/22`).
The branch carries no Q4_K dense attention-projection support, so the whole model is
unusable rather than merely slower. Worth knowing if AProjQ4 matters to the target
audience for this work.

## 5. Scope and caveats

- **ROCm 10 only.** Both arms ran under ROCm 10 because that is what the container
  provides. The +45.8% is a ROCm 10 figure; I have not measured #887 on ROCm 7.1.x, and
  since the PR pins specific rocBLAS solutions the result there could differ.
- **Q8 only** for the speed comparison — AProjQ4 could not run (§4).
- Single run per arm after a discarded warmup, not interleaved A-B-A-B. Sub-1% differences
  would be noise; a 46% effect is far outside that.
- Figures are DS4's own per-frontier `--csv` values, not whole-run averages.

## 6. Reproduction

```sh
# derive from kyuz0's Dockerfile, changing only the pinned commit
sed -e 's|ARG BRANCH=.*|ARG PRSHA=<sha>|' \
    -e 's|git clone -b ${BRANCH} --single-branch ${REPO} .|git clone ${REPO} . \&\& git fetch origin <ref> \&\& git checkout ${PRSHA}|' \
    toolboxes/Dockerfile.rocm-10.0 > Dockerfile.pinned
podman build -f Dockerfile.pinned -t ds4-pinned:rocm-10.0 .
toolbox create --image localhost/ds4-pinned:rocm-10.0 <name>
toolbox run --container <name> bash -lc \
  "ds4-bench -m <model> --prompt-file <prompt> \
     --ctx-start 2048 --ctx-max 8192 --step-incr 2048 --gen-tokens 128 --csv out.csv"
```

`<sha>`/`<ref>` = `9254c3da` / `pull/887/head` for the PR, `8db89fe` / `main` for the base.

kyuz0:
Thank you @iammac2. 

@antirez , Terminal Bench Mini confirmed quality on real coding and agentic workflows: 19/20 (95.0%), same as the full model! This is ready to merge, I just want to point out a few changes I made to help me benchmark this as I wanted, namely adding DSpark support to `ds4-bench` and a mode to `ds4-server` to skip `eos` tokens. These do not affect regular behaviour, they are just additional flags, but they could be dropped.

jyatesdotdev:
@kyuz0 this is great; do you think it's worth continuing to try and optimize for MXFP4? If the quality matches already with the smaller model and the speed is significantly improved, I'm not sure.

kyuz0:
> @kyuz0 this is great; do you think it's worth continuing to try and optimize for MXFP4? If the quality matches already with the smaller model and the speed is significantly improved, I'm not sure.

Quality matches on Terminal Bench Mini, but MXFP4 will be more quality and with upcoming Gorgon Halo this can fit on a single host, so I think it's worth trying to improve. After this PR gets merged, I might give it a go.

jyatesdotdev:
@kyuz0 makes sense! I've been working on https://github.com/antirez/ds4/pull/861, not sure how far I'll get.

PR #892, opened 2026-08-29
GLM-5.3 Flash on M5 Max: bring-up measurements + K-width speculative MTP (DS4_GLM_MTP_WIDTH)

audreyt:
# GLM-5.3 Flash on M5 Max: bring-up measurements + K-width speculative MTP (DS4_GLM_MTP_WIDTH)

Brings upstream `glm-5.3-flash` up on an M5 Max 128 GB and adds a working K-width speculative decode cycle for GLM-5.3's nextn MTP block, measured honestly against the upstream width-2 path.

## Measured baseline (M5 Max, Q2 GGUF, ctx 8192, greedy `--temp 0`)

| mode | prefill | decode |
|---|---|---|
| serial | 76–80 t/s (474 t/s @ 4500-tok prompt) | 33.0 t/s |
| `--mtp` (width 2, upstream) | same | **40.5 t/s** |

MTP acceptance 89.6% over 135 cycles, verify2 median 37.4 ms, head+draft ~6 ms. `make test-glm53-kda` PASS. Greedy goldens byte-identical across serial / `--mtp` / wide-MTP at widths 3, 4, 6.

## What's in the box

**`glm_graph_mtp_step` gains a `raw_hidden` mode.** Chained drafts consume the previous nextn step's own output hidden instead of the target's hc state. Two non-obvious correctness constraints, both load-bearing:

1. The snapshot of `g->next` must be taken *before* the embed stage overwrites it with the next-token embedding — a snapshot in the hnorm stage silently feeds the draft model its own token embedding as "hidden" and produces garbage drafts (symptom: `!defdef`-style doubled tokens).
2. The nextn KV slots at `pos..pos+rows-1` must be populated by the previous cycle's chain steps. Fresh seeding steps are skipped by default (`DS4_GLM_MTP_SEED_STEPS=1` restores) because accepted slots already hold the correct entries, and stale slots after a reject only degrade draft quality — every draft is re-verified, so output correctness never depends on them.

**`ds4_session_glm_spec_cycle_wide`**: verifies `first_token` + K−1 drafts in one layer-major forward (`glm_graph_forward_tokens` / indexed variant), scores every row through the shared head, commits the matching prefix, and on reject restores the KDA snapshot and replays the accepted prefix as a batch. Dispatched only when `DS4_GLM_MTP_WIDTH > 2`; the default path is bit-identical to upstream.

**`DS4_GLM53_PREFILL_CHUNK`** env override for prefill chunk diagnostics (1024/2048/4096 ABBA: 451/437/449 t/s median — noise-level; default unchanged).

## Findings (why the default stays width 2)

- **Decode is dispatch/latency-bound on this model**: a 2-token forward costs 1.23× a one-token forward (37.4 ms vs 30.3 ms). Speculative *width*, not kernel speed, is the lever — consistent with the Qwen3.8 finding that n_tok=2 is near-flat.
- Depth-2 draft acceptance is only ~45% (vs 89.6% at depth 1) even with correct raw chaining, and every reject costs a KDA restore + prefix replay (a second full forward). Per-token cost: W=3 → 30.6 t/s, W=4 → 20.8, W=6 → 16 — all worse than width 2.
- The wide machinery is kept because it flips profitable the moment KDA state supports per-row checkpointing (the reject tax disappears).
- Rejected with evidence: relaxing the `n_expert == 6` gate on `kernel_mul_mm_id_iq2_xxs_pair_swiglu_f16` to GLM's 8/288 fanout (bit-identical output, prefill ABBA median 296 vs 304 t/s — the fused epilogue only pays off at DSV4's 6-of-256).

`next-step.md` has the full measurement log: per-stage decode/prefill profiles, the flush-interval and indexed-decode null results, and MTP cycle math.

## Verification

- `make test-glm53-kda` PASS (GPU unit test for the KDA kernels).
- Greedy goldens: serial == `--mtp` == widths 3/4/6, byte-identical (`cmp`).
- Default width-2 path exercised end-to-end on the 90 GiB Q2 GGUF at ctx 8192 and on a 4500-token prompt.

## Not included

DFlash2 draft support for GLM-5.3: the draft GGUFs exist (qwen3-arch, same tokenizer) but this branch has no dflash2 machinery — it lives in `origin/ornith15`, bound to the Qwen graph. A port is scoped in `next-step.md` as a separate feature branch; the in-GGUF nextn MTP remains the draft mechanism here.


gilbert-barajas:
Filed a fix against this branch for --ssd-streaming: audreyt/ds4#7. The K-width cycle's one-token replay/commit forwards were falling back to full-layer expert mapping, which collapsed width 3 to 0.88 t/s and width 4 to 0.59 on an M5 Max 128GB streaming the Q4_K from NVMe. Routing single-row spans through the selected-address cache kernels takes width 3 to about 7 t/s with byte-identical greedy output. Details and measurements in the PR.

Issue #890, opened 2026-08-28
GLM-5.3-Flash Metal: prefill fails for prompts > 4096 tokens (compact indexed path) — HTTP 500 'metal GLM-5.3 prefill failed at token 0'


aj-img:
## Summary

On the `glm-5.3-flash` branch (a60a2a0), GLM 5.3 Flash on **Metal** fails to prefill any prompt longer than **4096 tokens**. Short prompts work fine. This breaks any long-context use (e.g. 50K-token prompts) on the very hardware the branch advertises (128GB M-series).

## Environment

- Machine: MacBook Pro M5 Max, 128 GB unified memory, macOS 27
- Backend: Metal
- Branch: `glm-5.3-flash` @ `a60a2a0` ("Add GLM 5.3 Flash inference")
- Model: `antirez/glm-5.3-flash-gguf` → `GLM-5.3-Flash-Q2.gguf` (~90 GiB, resident)

## Repro

```sh
./ds4-server -m gguf/GLM-5.3-Flash-Q2.gguf --metal --ctx 1000000 --port 4005
```

Then POST a chat completion with a prompt of N tokens (N varied below):

```sh
curl -s -X POST http://127.0.0.1:4005/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{"model":"deepseek-v4-flash","messages":[{"role":"user","content":"Say OK. <N tokens>"}],"max_tokens":5}'
```

## Results

| Prompt tokens | HTTP | Body |
|---|---|---|
| 3000 | 200 | ok |
| 4096 | 200 | ok |
| **5000** | **500** | `{"error":{"message":"metal GLM-5.3 prefill failed at token 0","type":"invalid_request_error"}}` |
| 8000+ | 500 | same |

Fails in ~2s regardless of size — it's an immediate path failure, not an OOM/timeout.

## Server log (successful boot)

```
ds4: GLM session ctx=1000000 (model max=1048576); full-attention prefill/work cap=4096; compact indexed decode is used beyond the cap
ds4: GLM graph allocating full-attention KV cache: work_ctx=4096 layers=0:44 (45) f16 2.75 GiB
ds4: GLM graph allocating compact DSA cache: rows=1000000 logical_ctx=1000000 kv_layers=45 indexer_layers=11 f16 11.15 GiB
```

The 4096 boundary matches `full-attention prefill/work cap=4096`; the **compact indexed prefill path** (used beyond the cap) is what fails on Metal.

## Expected

[...]

PR #893, opened 2026-08-29
metal: scale GLM 5.3 memory guard with host RAM

### LINKED ISSUE #889: q4 variant of GLM-5.3-Flash does not run on a Mac Studio with 512GB
When executing ds4-server to run the Q4 variant of GLM-5.3-Flash, the following message appears:

ds4: GLM memory guard refused ctx=1000000 compact_cap=1000000 before GLM graph allocation
ds4:   model map: 177.77 GiB
ds4:   graph cache/scratch: 18.14 GiB (full KV 2.75 GiB, compact DSA 11.15 GiB, scratch 4.25 GiB)
ds4:   required model+graph: 195.91 GiB; guard budget: 110.00 GiB (base 512.00 GiB, fraction 0.99, reserve 32.00 GiB, transient 0.00 GiB)
ds4:   GLM-5.3 has a fixed 110 GiB per-rank ceiling; use a smaller --ctx, tensor parallelism, or SSD streaming

The machine is a Mac Studio with 512 GB.  I would guess that there is a hardcoded value suited for a Q2 version.

kullbkri:
## Summary

- keep the fixed 110 GiB GLM-5.3 ceiling for 128 GiB-class hosts and TP ranks
- let 256/512 GiB hosts use the existing fraction, reserve, and wired-limit budgets instead of being unconditionally capped at 110 GiB
- report the fixed-ceiling diagnostic only when that ceiling actually reduced the budget
- document resident Q4 on 256 GiB+ Macs and add it to the release QA checklist

This fixes the Q4 refusal in #889 without weakening the existing protection on 128 GiB systems.

## Validation

Tested on a 256 GiB M3 Ultra with Metal and `GLM-5.3-Flash-Q4_K.gguf` (177.77 GiB), with no `DS4_GLM_MEMORY_GUARD` override.

Clean build:

```text
make clean && make -j8
```

Focused regression coverage passed:

```text
./ds4-eval --self-test-extractors
./ds4_agent_test
./ds4_test --server
./tests/test_layer_pack                 # 97/97
./tests/test_engine_mgpu_placement      # 101/101
./tests/test_gpu_args
./tests/test_gpu_args_cli.sh            # PASS=81 FAIL=0
./tests/test_sampling
```

The Q4 failure mode was exercised directly with the standard public-domain benchmark prompt:

```text
ctx_tokens=30000
prefill_tps=298.12
gen_tokens=128
gen_tps=20.52
gen_steady_tps=20.54
```

At a 32,768-token allocation DwarfStar planned 184.93 GiB total (177.76 GiB model, 3.12 GiB KV, 4.06 GiB buffers). macOS reported about 186 GiB wired, 26% memory-pressure headroom, and no swap growth.

A one-session server also started without an override at the full 1,048,576-token model context. It planned 196.44 GiB total (177.76 GiB model, 14.44 GiB KV, 4.25 GiB buffers), completed an OpenAI-compatible synthetic request, and reused 13 of 14 prompt tokens on the repeat.

Closes #889


eauchs:
This fixes one half of the guard's failure mode. The other half is still there, and it is on the hosts this PR deliberately leaves alone.

On a 128 GiB Mac the fixed 110 GiB ceiling is **above** the device's actual Metal working set, so the guard accepts plans the machine cannot honour. Measured on an M3 Max 128 GB, macOS 26.4, `GLM-5.3-Flash-Q2.gguf`, branch `glm-5.3-flash` @ `a60a2a0` (details in #890):

```
ds4: memory: KV 13.90 GiB + buffers 4.25 GiB + resident model 89.87 GiB = 108.01 GiB planned
```

108.01 < 110, so the guard passes it. But `sysctl iogpu.wired_limit_mb` is 0, so the working set is the macOS default — roughly 107.5 GiB on a 128 GB machine. The plan is already over the ceiling before any transient allocation, and the first prefill chunk pushes it over:

```
ds4: Metal command batch failed: Insufficient Memory (00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory)
ds4: GLM layer-major prefill failed at layer 44 stage 'FFN mHC expand' (pos 0, tokens 2048)
```

Controlled experiment — same binary, same flags, same prompts, plan byte-for-byte identical (`108.01 GiB planned`, `chunk=2048 score_rows=268`), only the sysctl changed:

| prompt | default limit | `iogpu.wired_limit_mb=114688` |
|---|---|---|
| 3000 words | 500 `prefill failed at token 0` | 200 |
| 5000 words | 500 | 200 |
| 8000 words | 500 | 200 |

Coherence at the raised limit, `--ctx 1000000`, 8024 prompt tokens: 8000 words of filler followed by `Ignore the text above. Reply with exactly: BANANA` returns `BANANA` with sane reasoning content.

So #889 and #890 are not two bugs but one seen from both sides: a constant that is too low on a 512 GiB host and too high on a 128 GiB one. Bounding the budget by the device's reported `recommendedMaxWorkingSetSize`, and by `iogpu.wired_limit_mb` when it is set, covers both directions with one rule and needs no per-class constant.

If the fixed ceiling is kept for 128 GiB hosts, `min(110 GiB, actual working set)` would at least stop the guard from accepting an over-budget plan. Naming the ceiling in the failure message would also make the sysctl workaround discoverable — right now the only signal is an OOM at layer 44.

eauchs:
Addendum, replacing an estimate in the comment above with a measurement.

I wrote that the default working set is "roughly 107.5 GiB" on a 128 GB machine. That was inferred from the macOS heuristic, not measured, and I would rather flag it than leave it standing as a number.

What I can measure is more useful anyway: `recommendedMaxWorkingSetSize` already tracks `iogpu.wired_limit_mb`.

```objc
id<MTLDevice> d = MTLCreateSystemDefaultDevice();
printf("%.2f GiB\n", d.recommendedMaxWorkingSetSize / 1073741824.0);
```

```
recommendedMaxWorkingSetSize = 112.00 GiB
sysctl iogpu.wired_limit_mb  = 114688      (= 112 GiB)
```

So the guard does not need to read the sysctl separately — one Metal query returns the effective ceiling with the user's override already applied. That simplifies the rule proposed above to a single source of truth, and it holds on hosts of any size, which is the case this PR is about.

eauchs:
Update to my addendum above: the ceiling figure is no longer an estimate.

I had written "roughly 107.5 GiB" for the stock working set on a 128 GiB host and flagged that it was inferred rather than measured. @evandhoffman has since measured it on a second machine — an M5 Max 128 GB on macOS 26.x, before applying any sysctl (#890, evanwtf/local-llm#30):

```
recommendedMaxWorkingSetSize: 107.52 GiB     # stock
recommendedMaxWorkingSetSize: 112.00 GiB     # with iogpu.wired_limit_mb = 114688
```

Two machines, two chip generations, agreeing to within 0.02 GiB — and the same query reflecting the override when one is set.

So the point this PR leaves open is now measured rather than argued: **the fixed 110 GiB ceiling sits above the real default working set on 128 GiB hosts.** A plan of 108.01 GiB passes the guard and then fails on the device, which is the mirror of the 512 GiB case fixed here. Sizing from `recommendedMaxWorkingSetSize` covers both directions from one source of truth, with no per-class constant and no separate sysctl read.

kullbkri:
Closing as superseded by b0c31af. This PR was prepared against the earlier a60a2a0 GLM branch state; the current glm-5.3-flash branch has removed the unconditional 110 GiB ceiling and now uses host-aware budgeting. The M3 Ultra 256 GiB validation and the subsequent 128 GiB recommendedMaxWorkingSetSize findings remain documented here for reference. The smaller-host safety issue continues in #890.

PR #923, opened 2026-08-31
ROCm: accelerate GLM 5.3 Flash prefill on gfx1151

kyuz0:
## Summary

This PR improves fully resident GLM-5.3-Flash-Q2 prefill and memory use on ROCm/gfx1151:

- 512-token prefill: **83.83 -> 179.70 tok/s** (`+114.4%`)
- 4096-token prefill: **82.24 -> 191.28 tok/s** (`+132.6%`)
- 8192-token prefill: **33.82 -> 111.98 tok/s** (`+231.1%`)
- stranded model-arena capacity: **13.38 -> 0.00 GiB**

The model remains fully resident. SSD streaming is not used.

## Performance

The matched baseline is main at [`3da2da96`](https://github.com/antirez/ds4/commit/3da2da96b250bc8a7564cf3342c34149b6cde429); the PR is rebased onto [`b0982a1b`](https://github.com/antirez/ds4/commit/b0982a1b4ee9d0f157e600bfd102fbeac951a829). Measurements used the same `GLM-5.3-Flash-Q2.gguf`, prompt, context allocation, and `ds4-bench` settings on AMD Radeon 8060S Graphics (`gfx1151`) with 128 GB unified memory and ROCm 10.0 / HIP 7.15.

| Fixed prefill | Matched main | This PR | Change |
|---:|---:|---:|---:|
| 512 tokens | 83.83 tok/s median | 179.70 tok/s median (179.65-179.81) | +114.4% |
| 4096 tokens | 82.24 tok/s | 191.28 tok/s median (191.18-191.37) | +132.6% |
| 8192 tokens | 33.82 tok/s | 111.98 tok/s median (111.89-112.02) | +231.1% |

An 8192-token prefill followed by 64 ordinary decode tokens completes at 112.20 prefill tok/s and 10.83 decode tok/s.

## Memory

| Metric | Matched main | This PR | Change |
|---|---:|---:|---:|
| Model data | 89.87 GiB | 89.87 GiB | unchanged |
| Stranded model-arena capacity | 13.38 GiB | 0.00 GiB | -13.38 GiB |

The fully resident memory plans are 90.72 GiB at 512 tokens, 92.84 GiB at 4096, and 92.89 GiB at 8192.

## Main changes

- GLM Q4_K projections use tiled MMQ for prefill-sized work while decode-sized rows retain the low-overhead direct path; fixed-512 Q4_K projection time falls from 919.1 ms to 110.8 ms.
- Compact causal LoRA attention batches score and value operations through hipBLAS.
- The tuned paired-IQ2 gate/up path admits only GLM-5.3-Flash's exact 288-expert, top-8, 4096x2048x4096 topology while retaining the conservative guard for unknown models.
- Long-context selected attention uses arithmetic-preserving four-head tiles beyond the dense-attention window.
- GLM model ranges use their merged sizes instead of fixed-size arenas, eliminating 13.38 GiB of stranded capacity.
- Partial MMQ tiles zero the invalid suffix once before the K loop and refresh only the valid prefix at each stage, retaining upstream's zero-before-WMMA correction while recovering register pressure, occupancy, and prefill throughput.

## DeepSeek V4 Vision rebase

During development this PR was rebased onto upstream main after [`fc8bf3c3`](https://github.com/antirez/ds4/commit/fc8bf3c39c51892c41f439a4ce17b90643dd1984) added DeepSeek V4 Flash vision, [`e976a543`](https://github.com/antirez/ds4/commit/e976a543314b0c190a6d6ef6c1b278cc68ccae13) added the ROCm Vision-Exp path, and [`f9c459ce`](https://github.com/antirez/ds4/commit/f9c459ce2f662cf2782de77bae1afc738f1f5625) corrected shared ROCm MMQ scratch lifetime and partial-tile reads.

The shared MMQ tail correction initially reduced DeepSeek prefill by 5.9-7.4%. This PR retains its safety properties while restructuring partial-tile loading: HIP scratch remains stream-local and live until safe destruction, invalid columns remain zero before every matrix operation, complete tiles use direct loads, and partial tiles zero their invalid suffix once and refresh only the valid prefix. Matched DeepSeek prefill reaches 234.70/298.89/294.37/290.42 tok/s at the 2K/4K/6K/8K frontiers, improving post-vision main by 5.6-7.0% and finishing within 1% of the pre-vision curve.

Because these kernels are shared by GLM and DeepSeek, validation includes the DeepSeek official scorer, MMQ production shapes, five real Vision-Exp image fixtures, and the repository vision-prompt test with a fully resident DeepSeek model and encoder. This verifies that the GLM performance work composes with the upstream DeepSeek V4 Vision changes rather than testing GLM in isolation.

## Validation

| Check | Result |
|---|---|
| GLM official scorer | 100 cases, 11,559 target tokens, NLL `0.460424520`, 88/100 first-token matches, LCP `6.980`; unchanged |
| DeepSeek official scorer | 100 cases, 2,313 target tokens, NLL `0.401858465`, first-token matches 55/100, LCP `5.100`; unchanged |
| IQ2 production shape | 288 experts, top-8, M=2048, K=4096; token counts 127/128/511/512/513 pass; canaries intact; maximum absolute difference `0.000854` |
| Long context | 8192+64 completes at 112.20 prefill and 10.83 decode tok/s |
| Vision prompt | Fully resident DeepSeek Vision-Exp model and encoder pass image identity reuse, invalidation, changed-image logits, restored-image exactness, and generation with an 81.18 GiB plan |
| ROCm tests | Full `test-rocm`, GLM KDA, IQ2 MMQ, and DeepSeek vision preprocessing suites pass |
| Build | Five production binaries and ROCm test/scorer targets build with zero compiler or linker warnings |
| Source | `git diff --check` passes |

## Backend scope

The GLM attention, MoE admission, and allocation changes are ROCm-specific. The MMQ tail-load restructuring is shared HIP/CUDA code and preserves the existing zero-before-matrix semantics; Metal implementation files are unchanged. The vision-prompt test now defaults to fully resident execution on every backend and enables ROCm SSD streaming only when `DS4_TEST_SSD_STREAMING=1` is explicitly set.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,12 +62,15 @@ ROCM_LDLIBS ?= -lm -pthread -lhipblas -lhipblaslt -lrocblas
 ROCM_MMQ_Y ?= 64
 ROCM_MMQ_FLAGS := $(ROCM_CFLAGS) -std=c++17 -DGGML_USE_HIP -DDS4_HIP_MMQ_Y=$(ROCM_MMQ_Y) $(MMQ_INCLUDES)
 ROCM_MMQ_OBJS := cuda/mmq/ds4_ggml_stubs.rocm.o cuda/mmq/ds4_mmq.rocm.o cuda/mmq/quantize.rocm.o cuda/mmq/mmid.rocm.o cuda/mmq/mmvq.rocm.o cuda/mmq/d2r_stubs.rocm.o
+ROCM_GLM_KERNEL_OBJS := rocm/ds4_rocm_glm_selected_multihead.o
+ROCM_QUALITY_SCORE := gguf-tools/quality-testing/score_official_rocm
+ROCM_QUALITY_OBJS := ds4.o ds4_image.o ds4_distributed.o ds4_tp.o ds4_ssd.o ds4_rocm.o ds4_rocm_compat.o ds4_rocm_unavailable.o ds4_layer_pack.o rax.o ds4_gpu_args.o $(ROCM_GLM_KERNEL_OBJS) $(ROCM_MMQ_OBJS)
 DS4_LINK ?= $(NVCC) $(NVCCFLAGS)
 DS4_LINK_LIBS ?= $(CUDA_LDLIBS)
 METAL_LDLIBS := $(LDLIBS)
 endif
 
-.PHONY: all help clean test test-rocm test-glm53-kda-rocm test-metal-session-batch test-mxfp4-cuda test-mxfp4-rocm test-cuda-session-batch test-cuda-mixed-batch dspark-acceptance dspark-verify-depth mtp-verify-depth cpu cuda cuda-spark cuda-generic cuda-regression strix-halo rocm
+.PHONY: all help clean test test-rocm test-glm53-kda-rocm test-glm53-iq2-mmq-rocm quality-score-rocm test-metal-session-batch test-mxfp4-cuda test-mxfp4-rocm test-cuda-session-batch test-cuda-mixed-batch dspark-acceptance dspark-verify-depth mtp-verify-depth cpu cuda cuda-spark cuda-generic cuda-regression strix-halo rocm
 
 ifeq ($(UNAME_S),Darwin)
 .PHONY: metal-decode-schedule-bench metal-prefill-variant-bench check-mxfp4-half-lut
@@ -163,6 +166,7 @@ help:
 	@echo "  make rocm                Alias for make strix-halo"
 	@echo "  make test-mxfp4-rocm     Build and run the synthetic ROCm MXFP4 MoE test"
 	@echo "  make test-rocm           Core regression suite on ROCm-only hosts"
+	@echo "  make quality-score-rocm  Build the official scorer against the exact ROCm objects"
 	@echo "  make cpu                 Build CPU-only ./ds4, ./ds4-server, ./ds4-bench, ./ds4-eval, and ./ds4-agent"
 	@echo "  make test                Build and run tests"
 	@echo "  make dspark-verify-depth Run DSpark speculative verification smoke if support GGUF is present"
@@ -185,12 +189,24 @@ cuda:
 
 strix-halo:
 	$(MAKE) -B ds4 ds4-server ds4-bench ds4-eval ds4-agent \
-		CORE_OBJS="ds4.o ds4_image.o ds4_distributed.o ds4_tp.o ds4_ssd.o ds4_rocm.o ds4_rocm_compat.o ds4_rocm_unavailable.o ds4_layer_pack.o $(ROCM_MMQ_OBJS)" \
+		CORE_OBJS="ds4.o ds4_image.o ds4_distributed.o ds4_tp.o ds4_ssd.o ds4_rocm.o ds4_rocm_compat.o ds4_rocm_unavailable.o ds4_layer_pack.o $(ROCM_GLM_KERNEL_OBJS) $(ROCM_MMQ_OBJS)" \
 		CFLAGS="$(CFLAGS) $(ROCM_HOST_CFLAGS) -DDS4_ROCM_BUILD" \
 		DS4_LINK="$(HIPCC) $(ROCM_CFLAGS)" \
 		DS4_LINK_LIBS="$(ROCM_LDLIBS)"
 
 rocm: strix-halo
+
+ifneq ($(UNAME_S),Darwin)
+quality-score-rocm:
+	$(MAKE) -B $(ROCM_QUALITY_SCORE) \
+		CFLAGS="$(CFLAGS) $(ROCM_HOST_CFLAGS) -DDS4_ROCM_BUILD"
+
+gguf-tools/quality-testing/score_official.rocm.o: gguf-tools/quality-testing/score_official.c ds4.h ds4_distributed.h ds4_gpu_args.h ds4_ssd.h ds4_tp.h
+	$(CC) $(filter-out -ffast-math,$(CFLAGS)) -I. -c -o $@ $<
+
+$(ROCM_QUALITY_SCORE): gguf-tools/quality-testing/score_official.rocm.o $(ROCM_QUALITY_OBJS)
+	$(HIPCC) $(ROCM_CFLAGS) -o $@ $^ $(ROCM_LDLIBS)
+endif
 
 # Core regression suite for ROCm-only hosts: the CUDA-specific binaries
 # (tests/test_sampling, the CUDA session/mixed-batch oracles) are not part
@@ -200,7 +216,7 @@ test-rocm:
 	$(MAKE) -B ds4_test ds4_agent_test ds4-eval q4k-dot-test mxfp4-dot-test \
 		tests/test_layer_pack tests/test_engine_mgpu_placement tests/test_gpu_args \
 		ds4 ds4-server ds4-bench ds4-agent \
-		CORE_OBJS="ds4.o ds4_image.o ds4_distributed.o ds4_tp.o ds4_ssd.o ds4_rocm.o ds4_rocm_compat.o ds4_rocm_unavailable.o ds4_layer_pack.o $(ROCM_MMQ_OBJS)" \
+		CORE_OBJS="ds4.o ds4_image.o ds4_distributed.o ds4_tp.o ds4_ssd.o ds4_rocm.o ds4_rocm_compat.o ds4_rocm_unavailable.o ds4_layer_pack.o $(ROCM_GLM_KERNEL_OBJS) $(ROCM_MMQ_OBJS)" \
 		CFLAGS="$(CFLAGS) $(ROCM_HOST_CFLAGS) -DDS4_ROCM_BUILD" \
 		DS4_LINK="$(HIPCC) $(ROCM_CFLAGS)" \
 		DS4_LINK_LIBS="$(ROCM_LDLIBS)"
@@ -375,7 +391,7 @@ test-glm53-kda: $(GLM53_KDA_TEST)
 tests/test_glm53_kda_rocm.o: tests/test_glm53_kda.c ds4_gpu.h
 	$(CC) $(filter-out -ffast-math,$(CFLAGS)) $(ROCM_HOST_CFLAGS) -DDS4_ROCM_BUILD -I. -c -o $@ $<
 
-$(GLM53_KDA_ROCM_TEST): tests/test_glm53_kda_rocm.o ds4_rocm.o
+$(GLM53_KDA_ROCM_TEST): tests/test_glm53_kda_rocm.o ds4_rocm.o ds4_image.o $(ROCM_GLM_KERNEL_OBJS) $(ROCM_MMQ_OBJS)
 	$(HIPCC) $(ROCM_CFLAGS) -o $@ $^ $(ROCM_LDLIBS)
 
 test-glm53-kda-rocm: $(GLM53_KDA_ROCM_TEST)
@@ -411,6 +427,9 @@ cuda/mmq/ds4_repack.o: cuda/mmq/ds4_repack.cu cuda/mmq/ds4_repack.h
 ds4_rocm.o: ds4_rocm.cu ds4_gpu.h ds4_glm53_vision_gpu.cuh ds4_deepseek4_vision_gpu.cuh ds4_image.h ds4_iq2_tables_cuda.inc $(ROCM_SRCS)
 	$(HIPCC) $(ROCM_CFLAGS) -c -o $@ ds4_rocm.cu
 
+rocm/ds4_rocm_glm_selected_multihead.o: rocm/ds4_rocm_glm_selected_multihead.cu
+	$(HIPCC) $(ROCM_CFLAGS) -std=c++17 -c -o $@ $<
+
 cuda/mmq/ds4_ggml_stubs.rocm.o: cuda/mmq/ds4_ggml_stubs.cu cuda/mmq/ds4_ggml_stubs.h cuda/mmq/common.cuh cuda/mmq/vendors/hip.h
 	$(HIPCC) $(ROCM_MMQ_FLAGS) -c -o $@ $<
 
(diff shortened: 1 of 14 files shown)


saxs2224:
Hi @kyuz0, thanks for opening this — I ran into exactly the prefill slowness this PR seems to target, wanted to share a data point from a real Strix Halo box.

**Setup**
- Hardware: Ryzen AI Max+ 395 / Radeon 8060S (gfx1151), 128GB unified memory
- OS: Debian trixie, ROCm 7.2.2
- Branch: `main`, checked out at commit `ad4ac91` (last commit before GLM 5.3 vision support was added — I had to pin to this commit because `main` currently fails to build with `--rocm` due to an unrelated `rsqrtf` host/device compile error in `ds4_rocm_deepseek4_vision.cuh`, introduced in `4771329`)
- Model: `GLM-5.3-Flash-Q2.gguf` (glm53-q2 target, ~90GiB)
- Command: `./ds4-server --rocm -m gguf/GLM-5.3-Flash-Q2.gguf --ctx 100000 --kv-disk-dir /var/lib/ds4-kv --kv-disk-space-mb 24576`

**What I observed**
Prefill on a short prompt (~30 tokens) is fast (~40 t/s), and decode is solid and stable (~14.4 t/s across multiple runs, holding steady from ctx=8192 up to ctx=100000). But prefill on a longer prompt via tool-calling (Pi agent, ~2300 new tokens on top of existing context) dropped to:

```
chat ctx=8091..10414:2323 TOOLS prefill chunk 2048/2323 (88.2%) chunk=0.00 t/s avg=21.08 t/s 97.148s
```

~21 t/s average prefill, ~97s for ~2300 tokens — much slower than the short-prompt case and slower than what I'd expect given decode throughput is fine.

Happy to test the branch/patch here if you have something ready, or run `ds4-bench`/collect more structured numbers if useful. Let me know what data would help.

kyuz0:
> Happy to test the branch/patch here if you have something ready, or run `ds4-bench`/collect more structured numbers if useful. Let me know what data would help.

Thank you, tis is already built in my toolbox, and I recommend using it with ROCM 10.0, as I tuned specifically for rocBLAS kernels present in that release. The Ai Toolbox Cockpit is the best way to load the prebuilt container that includes this PR:

https://github.com/kyuz0/ai-toolbox-cockpit

Otherwise the toolbox is built from here: https://github.com/kyuz0/strix-halo-ds4-toolbox.

Let me know how you find it!

saxs2224:
Thanks — dug a bit deeper before reporting back, wanted to share what I found in case it narrows things down.

Same hang, tried disabling specific paths via env vars found in the binary strings:

DS4_ROCM_GLM_Q4_MMQ=0: the "tiled MMQ" log line disappears entirely (so the var does something), but it hangs even earlier now — before any prefill log line at all, still 0% GPU, so whatever fallback it takes also gets stuck.

DS4_GLM53_DISABLE_INDEXED_PREFILL=1: this one changes behavior — no more hang, but instead a repeated explicit failure on every request:

ds4: GLM graph allocating compact DSA cache: rows=100000 ... chunk=0 score_rows=0 score_scratch=0.00 MiB
ds4: GLM full-attention prefill was requested without an expanded KV cache

Looks like disabling indexed prefill routes to a full-attention path, but the expanded KV cache for that path was never allocated (the startup log explicitly says "expanded full-attention KV cache is skipped" when using compact DSA KV only) — so it's a real gap between what the flag disables and what the fallback needs.

Honest question: I'm not fully sure this is hitting the same code path you've validated on your end — is this exact combo (rocm-10.0 image, GLM-5.3-Flash Q2, real multi-thousand-token prefill via tool-calling, not just a short prompt) something you've actually run successfully on your Strix Halo box? Wondering if maybe you tested primarily with a shorter/simpler prompt, or a different quant, and this is a genuine gap in coverage rather than something misconfigured on my end. Happy to test whatever would help pin it down.

kyuz0:
Thanks fro testing! Did you also experience the hang on the main branch? This is still in draft as I haven't had time to properly test it, I started it on Terminal Bench Mini, 5 tasks, and I had to stop to do other things, unfortunately takes 24 hours to go through it and right now I am prioritising other testing.

But, I did observe some failures and looping on terminal bench, so there might very well be an inference bug somewhere, but I did not experience any hangs whatsoever, so not sure what is going on.

saxs2224:
Thanks for the honest update — that actually helps a lot. To answer your question directly: I never got main to compile with --rocm at all (unrelated rsqrtf host/device error in ds4_rocm_deepseek4_vision.cuh, introduced when vision support was added), so all my testing today was exclusively on your rocm-10.0 image, tracking perf/rocm-gfx1151-mmq-kernel-lab. I can't tell you anything about main's behavior.

Here's the full picture from my side, in case any of it narrows down what's different between our setups:

Environment

Host: Debian trixie, ROCm 7.2.2, kernel 6.12.101+deb13
Hardware: Ryzen AI Max+ 395 / Radeon 8060S (gfx1151), 128GB unified memory, GTT raised to ~124GiB via amdgpu.gttsize/ttm.pages_limit boot params
Container runtime: Docker (not Podman/Toolbx) — I don't have Podman installed, so I used plain docker run and separately tried Distrobox
Model: GLM-5.3-Flash-Q2.gguf (~90GiB, the glm53-q2 target from antirez's own download_model.sh)

Container GPU passthrough issue (unrelated to the hang, but worth flagging for the docs)
The cockpit's generated command uses --group-add keep-groups, which is Podman-specific and silently does nothing under Docker. Same problem with named groups (--group-add video --group-add render) under Distrobox — the container's render/video groups have different GIDs (105/39) than the host's (992/44 in my case), so the container never actually gets device access; devices show up owned by nobody:nobody. Had to pass explicit numeric host GIDs (--group-add 992 --group-add 44) with plain docker run to get ROCm to detect the GPU at all. Not sure if this matters for your own testing since you're presumably on Fedora/Toolbx where this may not come up, but could trip up other Docker/Distrobox users trying this image.

The hang itself
Once the GPU is actually visible (ROCm backend initialized on AMD Radeon 8060S Graphics) and the model loads fine (89.87 GiB, ~23s), the first real request hangs. Trigger: an agent (Pi, using tool-calling) sending a prompt of ~2700 tokens as the first turn. Short prompts (~30 tokens) work fine, both prefill and decode, at ~14.4 t/s decode / ~40 t/s prefill — that part is solid and consistent across ctx=8192 up to ctx=262144.

Log stops after exactly one line:

ds4: ROCm GLM Q4_K prefill using tiled MMQ

No error, no further output. rocm-smi shows 0% GPU utilization indefinitely (not slow — genuinely idle, checked repeatedly over minutes). Thread states via /proc/1/task/*/wchan:

TID 7, 8: kfd_wait_on_events
TID 10: drm_syncobj_array_wait_timeout

So a GPU kernel dispatch appears to never signal completion.

Env var experiments (from strings-ing the binary since the container doesn't have gdb/strings and I improvised with tr/grep)

DS4_ROCM_GLM_Q4_MMQ=0: removes the "tiled MMQ" log line entirely, but hangs even earlier — before any prefill log line, still 0% GPU. So whatever fallback path it takes also gets stuck.
DS4_GLM53_DISABLE_INDEXED_PREFILL=1: different behavior — no hang, but an explicit repeated failure instead:
  ds4: GLM graph allocating compact DSA cache: rows=100000 ... chunk=0 score_rows=0 score_scratch=0.00 MiB
  ds4: GLM full-attention prefill was requested without an expanded KV cache

Looks like disabling indexed prefill routes to a full-attention path, but the expanded KV cache for that path isn't allocated (startup log explicitly says "expanded full-attention KV cache is skipped" when using compact DSA KV only). Seems like a real gap between what that flag disables and what the fallback needs, independent of the hang itself.

My best guess, for what it's worth
Given the log names "Q4_K" while I'm loading a Q2 file, and given your own Terminal Bench run saw failures/looping (not hangs) — my guess is there might be two separate issues: an inference-correctness bug you're already seeing on shorter prompts, and a second, prefill-path-specific issue that only shows up on longer prompts/real tool-calling contexts, which is what I'm hitting. Pure speculation on my part though.

Happy to grab a gdb backtrace of the hung threads, run with DS4_ROCM_GRAPH_DUMP_TRACE, try a Q4_K GLM build instead of Q2, or anything else that'd help — no urgency, I know you're mid-benchmark on other things

jyatesdotdev:
I'm going to test this and give it a run on my setup! I haven't played around with GLM yet, should be interesting! I'll post a follow up after I do that.

kyuz0:
@jyatesdotdev - there is an issue with GLM 5.3 inference, I think that will be fixed mainstream, haven't checked if antirez has already fixed it.

jyatesdotdev:
@kyuz0 noted, I'll hold off then. Thanks!

saxs2224:
Tested on AMD Ryzen AI Max+ 395 / Radeon 8060S (gfx1151), 128GB unified memory, GLM-5.3-Flash-Q2.gguf, ROCm 10.0 / HIP 7.15 (same stack as the PR description). Cherry-picked `ba5bed8` + `f24febf` onto current `main` (the PR branch itself is based on an old revision with unrelated noise).

**Fixed prefill** (`ds4-bench --gen-tokens 0`, fresh process per context length, matches your methodology):

| ctx | baseline | with PR | speedup |
|---:|---:|---:|---:|
| 512   | 82.33 | 171.05 | +107.8% |
| 4096  | 81.39 | 187.60 | +130.5% |
| 8192  | 33.84 | 111.61 | +229.9% |
| 16384 | 25.40 | 84.17  | +231.3% |

Matches your numbers within a few percent at every point (yours: 83.83/179.70, 82.24/191.28, 33.82/111.98). Confirms the speedup independently, gain holding around +230% out to 16k.

Correctness: `test-glm53-iq2-mmq-rocm` (the exact tail-tile area the 2nd commit touches) and `test-mxfp4-rocm` both PASS, 0 failures. (The latter needed a manual link — its Makefile rule is missing object files `ds4_rocm.o` now depends on, pre-existing on `main`, unrelated to this PR.)

**Found a real bug though: `glm_selected_multihead_kernel` output is non-deterministic once it fires for a 2nd consecutive chunk** (i.e. context above ~6144 with the default 2048-token chunk). Same binary, same prompt, rerun twice → different logits, diffs up to ~1.5 — not rounding noise, enough to flip argmax in close calls. Single chunk (≤~6144) is always deterministic; 2+ chunks aren't, 100% reproducible here.

Confirmed it's this PR's new path specifically: `DS4_ROCM_GLM_SELECTED_ATTN_MULTIHEAD=0` (fallback) is fully deterministic (~2.5x slower). Ruled out shared-memory overflow (36KB/64KB used) and a missing sync right around the kernel launch (sync immediately before/after doesn't fix it) — not a bug in the kernel's own dispatch.

Fix: a full `ds4_gpu_synchronize()` at the end of `glm_graph_forward_indexed_tokens` (`ds4.c`, before the final `return ok;`) makes it deterministic — verified 0/154892 differing logits at both ctx=8192 and ctx=16384. Cost: within noise (120.91 vs ~121 tok/s @8192, 92.37 vs ~92.3 @16384). Didn't nail the exact race — my guess is the next chunk's indexer scoring starts reusing shared scratch tensors (`batch_indexer_selected`, `layer_indexer_key_cache`) before this chunk's async GPU work finishes — but this barrier fixes it at no measured cost, so it's a reasonable stopgap even if a more targeted fix is possible:

```diff
--- a/ds4.c
+++ b/ds4.c
@@ -51043,6 +51043,15 @@ glm53_indexed_attention_done:
     ds4_gpu_tensor_free(cur_view);
     glm_vision_overlay_free(&vision_overlay);
     ds4_gpu_set_glm_streaming_prefill_full_layer(false);
+    /* Without this barrier, ROCm indexed prefill is non-deterministic
+     * once a second consecutive compact-indexed chunk is processed
+     * (rerunning the exact same input on the same binary twice produces
+     * different logits, diffs up to ~1.5 observed) - reproducible with
+     * DS4_ROCM_GLM_SELECTED_ATTN_MULTIHEAD unset, gone with it set to 0.
+     * Root cause not fully traced; likely the next chunk's indexer
+     * scoring begins reusing shared scratch tensors
+     * (batch_indexer_selected, layer_indexer_key_cache, ...) before this
+     * chunk's async GPU work has completed. This barrier is a
+     * correctness fix, not just a perf regression: measured cost is
+     * within noise (<1%) on GLM-5.3-Flash-Q2 prefill at 8k/16k ctx. */
+    if (ok) ok = ds4_gpu_synchronize() != 0;
     return ok;
 }
```

Worth applying given the speedup, but shouldn't merge without this (or an equivalent) barrier — right now it silently gives wrong/non-reproducible output past ~6k context on ROCm.

Also two unrelated ROCm/CUDA build bugs on plain `main` (hit them cherry-picking, happy to send as a separate PR): `ds4.c` has a few Metal-only calls used without `#if defined(__APPLE__)` guards (`ds4_gpu_dsv4_qkv_norm_defer_kv_next`, `ds4_gpu_add_tensor_tp_flag`, `g_tp_block_ctx`); `rocm/ds4_rocm_deepseek4_vision.cuh:239` calls `rsqrtf()` (device-only on HIP) from host code — fix is `1.0f / sqrtf(x)`.

saxs2224:
Follow-up: found an extra ~8-10% on top of the above, and forgot to mention it in the first comment.

`glm_selected_multihead_kernel`'s score loop reads the f16 KV-LoRA cache one scalar `__half` at a time. Since `kv_lora_dim` is always even here, batching two lanes per step with a `__half2` load halves the memory instructions on that path (same accumulation order per lane, so it's not a reassociation — verified bit-exact via `--dump-frontier-logits-dir`, 0/154892 differing values at both ctx=8192 and ctx=16384 against the unoptimized output):

```diff
--- a/rocm/ds4_rocm_glm_selected_multihead.cu
+++ b/rocm/ds4_rocm_glm_selected_multihead.cu
@@ -58,14 +58,42 @@ __global__ void glm_selected_multihead_kernel(
             const float *low = qk_low +
                 (static_cast<uint64_t>(token) * n_head + head0) *
                 kv_lora_dim;
-            for (uint32_t j = 0; j < kv_lora_dim; j++) {
-                const float kv = cache_load(
-                    kv_lora_cache, cache_base + j, cache_f16);
+            /* Fast path: cache_base is always a multiple of kv_lora_dim,
+             * and kv_lora_dim is even for every model this kernel is
+             * enabled for (checked by cuda_attention_score_buffer_fits /
+             * the shape validation before launch), so a half2 load off
+             * cache_base is always 4-byte aligned. Processing two lanes
+             * per step does not reorder the accumulation (dot[h] still
+             * receives lane j then lane j+1, same as the scalar path),
+             * it only batches the load+conversion of the f16 pair. */
+            if (cache_f16 && (kv_lora_dim & 1u) == 0u) {
+                const __half2 *cache2 = reinterpret_cast<const __half2 *>(
+                    kv_lora_cache) + (cache_base >> 1);
+                for (uint32_t j = 0; j < kv_lora_dim; j += 2u) {
+                    const float2 kv2 = __half22float2(cache2[j >> 1]);
 #pragma unroll
-                for (uint32_t h = 0; h < kHeadTile; h++) {
-                    if (h < head_count) {
-                        dot[h] += low[static_cast<uint64_t>(h) *
-                                      kv_lora_dim + j] * kv;
+                    for (uint32_t h = 0; h < kHeadTile; h++) {
+                        if (h < head_count) {
+                            const float *low_h = low +
+                                static_cast<uint64_t>(h) * kv_lora_dim + j;
+                            /* Two separate += (not a fused a+b) to match
+                             * the scalar loop's addition order exactly -
+                             * float addition is not associative. */
+                            dot[h] += low_h[0] * kv2.x;
+                            dot[h] += low_h[1] * kv2.y;
+                        }
+                    }
+                }
+            } else {
+                for (uint32_t j = 0; j < kv_lora_dim; j++) {
+                    const float kv = cache_load(
+                        kv_lora_cache, cache_base + j, cache_f16);
+#pragma unroll
+                    for (uint32_t h = 0; h < kHeadTile; h++) {
+                        if (h < head_count) {
+                            dot[h] += low[static_cast<uint64_t>(h) *
+                                          kv_lora_dim + j] * kv;
+                        }
                     }
                 }
             }
```

Combined with the `ds4_gpu_synchronize()` fix from my previous comment: **120.9 tok/s @ ctx=8192, 92.3 tok/s @ ctx=16384** (up from 111.61 / 84.17), still fully deterministic.

I only vectorized the score-computation loop (the one where a thread already scans the whole `kv_lora_dim` range sequentially). There's a second, symmetric loop (weighted-sum accumulation) doing comparable work that looked like an equally good target, but my attempt at vectorizing it introduced real logit differences at ctx=16384 (not present at ctx=8192) that I couldn't fully explain, so I dropped it rather than ship something I wasn't sure about — flagging in case it's a more obvious fix for someone who knows this kernel better.

kyuz0:
Thank you, please do not waste any more time on this, I will close it and discard the work and start again from the correct GLM 5.3 Fkash implementation in main right now, this implementation is not correct.

PR #924, opened 2026-08-31
agent, engine: restore text tool observation path for non-GLM models

vanantwerp:
### Summary

Fixes an issue introduced in `47713296d` where executing any tool in `ds4-agent` under DeepSeek V4 models immediately failed with `ds4-agent: context full after compaction`.

### Root Cause

1. `ds4_agent.c:agent_tool_observation_build()` was refactored to route all tool observations through `ds4_chat_append_multimodal_message()`.
2. In `ds4.c`, `ds4_chat_append_multimodal_message()` enforced `if (DS4_MODEL_FAMILY != DS4_MODEL_FAMILY_GLM_DSA) return 0;`, failing unconditionally for DeepSeek V4 even on text-only tool results.
3. `agent_tool_observation_fits()` received `false` from the failed build and treated it as a context-capacity overflow, triggering context compaction and ultimately reporting a false `context full after compaction` error.

### Changes

1. **`ds4.c`**: In `ds4_chat_append_multimodal_message()`, handle `image_count == 0` by delegating to `ds4_chat_append_message()` so zero-image messages are valid across all model families.
2. **`ds4_agent.c`**: In `agent_tool_observation_build()`, fast-path `obs->image_count == 0` directly to `ds4_chat_append_message()`, avoiding unnecessary heap allocations and keeping the text-only path clean.

### Validation

- Tested with `ds4-agent -m ds4flash.gguf` on web search (`google_search`) and shell tools (`bash`), confirming tool execution succeeds without triggering compaction or context errors.
- Verified standard test suites pass: `make clean && make`, `make cpu`, and `make test` (all 8 test suites passed on macOS Metal Apple M5 Max).
- Verified GLM 5.3 Flash vision and multimodal tools continue to work.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -64144,6 +64144,14 @@ int ds4_chat_append_multimodal_message(
     }
     const bool tool = !strcmp(role, "tool") || !strcmp(role, "function");
     const bool user = !strcmp(role, "user");
+    /* A message without images is an ordinary chat message. Keep the native
+     * per-family rendering so a text-only tool observation never depends on
+     * vision support: DeepSeek text models render <tool_result> under the
+     * user role, not the observation tags used for image messages below. */
+    if (image_count == 0 && (tool || user)) {
+        ds4_chat_append_message(e, tokens, role, text_parts[0]);
+        return 1;
+    }
     if ((DS4_MODEL_FAMILY != DS4_MODEL_FAMILY_GLM_DSA &&
          e->vision_kind != DS4_VISION_DEEPSEEK4) || (!tool && !user)) {
         if (error && error_cap)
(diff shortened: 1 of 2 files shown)


rinaldofesta:
Hit the same thing independently on an M5 Max with DeepSeek V4 Flash 0731, no `--vision`: the first `list .` dies with `ds4-agent: context full after compaction` at 1.9k of 262k. This patch no longer applies on main after fc8bf3c moved the guard. I opened #963 with the same `ds4.c` delegation plus a regression test that fails on the old engine (`--chat-multimodal-text-only`). Happy to fold the test into this PR instead if you rebase it.

vanantwerp:
Thanks @rinaldofesta! I have rebased onto current `main` to resolve the conflict from `fc8bf3c` and incorporated your `--chat-multimodal-text-only` regression test, adding you as co-author on the commit. Verified locally with `./ds4_test --chat-multimodal-text-only`, `./ds4_test --server`, and `./ds4_agent_test` (all passing).

PR #927, opened 2026-09-01
server: reuse KV when tools append images

JordiPosthumus:
## Summary

Reuse an exact image-conditioned live KV prefix when a tool appends a new image
strictly after that prefix.

## Why

DS4 records the image spans and embedding fingerprints that conditioned a live
checkpoint. Reuse previously required the next request to contain exactly the
same number of images. A vision tool appending a new image therefore discarded
the whole live prefix even when every old token and image was unchanged and the
new image began after the checkpoint frontier.

The prefix is reusable only when:

- every token through the live frontier matches;
- every existing image span and fingerprint matches; and
- every appended span is nonempty, ordered, non-overlapping and begins at or
  after the live frontier.

Changed, removed, moved or replaced images remain cold, as do images inserted
inside already-computed tokens. Protocol reconstruction, text-prefix reuse and
rewinds still require an exact image-state match. Resident-slot routing uses the
same eligibility rule.

## Compatibility

Image-conditioned KV remains excluded from the text-keyed disk cache. The
multimodal protections from #916 are unchanged. There is no cache-format,
sampling, model-math, launcher or default change.

## Practical impact

The benefit is less repeated vision prefill when a tool appends an image after an unchanged live prefix. It does not make ordinary decode faster or enable appended-image disk recovery. The earlier reused-token measurements below demonstrate the affected workflow; their timings have not been remeasured against `f62ca29`. This live-prefix change and #961's persistent image-conditioned cache address different reuse boundaries.

## Combined runtime context

## Validation

One standalone commit, `f08cd7e54586`, directly on upstream `f62ca29a3087` (2026-09-07). Tests accept appended images and reject changed tokens/fingerprints, removed or moved images, insertion before the frontier, empty/overlapping spans and incompatible resident slots. Opaque session construction remains test-only.

Earlier M3 Ultra/Metal Vision-Exp MXFP4 evidence: appended images reused 10,884 and 11,288 tokens and evaluated only 231 and 387 new tokens; changing an existing image rebuilt cold. Mirrored 2K/16K text checks kept frontier logits exact, with timing differences below 0.5%. These measurements predate this rebase. #960 overlaps the reuse dispatch; preserve exact cancelled-retry priority and strict protocol-reconstruction matching when combining them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,6 +68,11 @@ DS4_LINK_LIBS ?= $(CUDA_LDLIBS)
 METAL_LDLIBS := $(LDLIBS)
 endif
 
+# ds4_test needs opaque-session constructors that must not ship in production
+# binaries. Build the core with test fixtures instead of exporting them from
+# the normal ds4.o.
+TEST_CORE_OBJS = ds4_test_core.o $(filter-out ds4.o,$(CORE_OBJS))
+
 .PHONY: all help clean test test-rocm test-glm53-kda-rocm test-metal-session-batch test-mxfp4-cuda test-mxfp4-rocm test-cuda-session-batch test-cuda-mixed-batch dspark-acceptance dspark-verify-depth mtp-verify-depth cpu cuda cuda-spark cuda-generic cuda-regression strix-halo rocm
 
 ifeq ($(UNAME_S),Darwin)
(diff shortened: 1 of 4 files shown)


JordiPosthumus:
Closing this PR as superseded by upstream commit 81a76585ff44e0494053fca9101130e2535648b9 (Preserve live vision context across server tool calls).

That change covers this PR's core appended-image live-KV reuse behavior, including multimodal resume and resident-slot selection, and adds authenticated vision continuation across the three API formats.

This PR also contains stricter image-span validation and adversarial tests that are not identical to upstream's implementation. Those can be evaluated separately as a small follow-up if useful; they do not justify keeping the overlapping implementation open.

PR #930, opened 2026-09-01
server: separate resident sessions from active requests

JordiPosthumus:
## Summary

Add `--max-active-requests N` so resident KV capacity and inference concurrency
can be configured independently.

## Why

`--batched-session` currently controls both how many KV sessions stay resident
and how many requests may occupy inference slots. A deployment may want several
warm conversations but only one active request to avoid mixed prefill/decode
contention.

For example:

```sh
./ds4-server --batched-session 10 --max-active-requests 1 ...
```

This allocates ten resident slots while admitting one assigned or running
request. Additional requests remain queued until that request completes or is
cancelled. `busy` already spans assignment, worker claim and execution, so the
new limit reuses the existing ownership state rather than adding another
scheduler.

This is admission control only. It does not change which slot the existing
router selects. As discussed in #765, reuse-aware slot selection is
complementary; this option alone does not guarantee that ten unrelated
sequential conversations all remain warm.

## Compatibility

Omitting the option preserves current behavior by setting the active limit to
the resident-session count. The option requires batched mode, must be positive
and cannot exceed the resident count. Model execution, cache formats, sampling
and existing routing are unchanged.

## Practical impact

This adds control over contention: a deployment can retain several warm sessions while admitting fewer simultaneous inference requests. It is not an unconditional throughput increase; reducing active requests can improve an individual request's latency while reducing parallel service capacity. The default is unchanged. The tests establish admission behavior, not an optimal setting or a measured speedup over `f62ca29`.

## Combined runtime context

## Validation

One standalone commit, `29e6104e792a`, directly on upstream `f62ca29a3087` (2026-09-07). Tests cover the unchanged default, one and two active requests, queuing, assignment, worker claim, completion and cancellation. Invalid CLI combinations are rejected before serving.

### DIFF
--- a/docs/SERVER.md
+++ b/docs/SERVER.md
@@ -60,6 +60,17 @@ preallocates independent KV states and queues requests when all slots are busy.
 Choose context and slot count together: a context that fits once may not fit
 four times. Idle slots can be cached before reuse; active requests are not evicted.
 
+`--max-active-requests N` separates resident KV capacity from compute
+concurrency. It requires `--batched-session`, may not exceed the resident
+session count, and defaults to that count so existing behavior is unchanged.
+For example, `--batched-session 10 --max-active-requests 1` allocates ten
+resident session slots while processing complete requests one at a time;
+additional requests wait in the server queue instead of occupying another
+active slot. This avoids mixed prefill/decode contention without reducing
+resident KV capacity. Request-to-slot selection is unchanged: this option
+controls admission concurrency, not which checkpoint the existing router
+chooses for a request.
+
 | Backend/model | Decode execution |
 | --- | --- |
 | Metal, resident Flash | Native shared-expert/QKV batching where supported |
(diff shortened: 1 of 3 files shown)


kmike:
> ELI20: --batched-session 10 currently does two things at once: it allocates
ten independent KV sessions and allows up to ten requests into the inference
scheduler. Setting it to one removes compute contention, but also leaves room
for only one warm conversation, so unrelated conversations repeatedly evict
each other's KV. This patch splits those controls. The server can retain ten
conversation checkpoints in memory while assigning only one request at a time;
queued requests do not consume or evict a resident slot until they are actually
scheduled.

I think it's currently a bit more nuanced (or, rather, a bit more broken :). Even with --batched-session 10 the in-memory KV cache is only kept for 1 slot, if there is 1 request at a time (no parallel requests), due to broken logic of "which KV cache slot to use". On-disk KV cache can somewhat mask the issue, but it's unnecesary SSD trashing. See https://github.com/antirez/ds4/pull/765.

JordiPosthumus:
Agreed — slot selection is a separate upstream problem, and the earlier description overstated what this PR guarantees. #930 limits admission only; it does not fix the routing behavior discussed in #765.

I have updated the code documentation and PR description accordingly. With current upstream routing, `--batched-session 10 --max-active-requests 1` allocates ten slots and limits compute to one request, but cannot guarantee that ten sequential conversations occupy ten different warm slots. A reuse-aware/staleness-aware routing fix is complementary work.

PR #939, opened 2026-09-01
glm: fix [gMASK]<sop> tokenization without BOS metadata

mvid:
## Problem

GLM chat templates start with `[gMASK]<sop>`. Some converted GGUFs omit `tokenizer.ggml.bos_token_id`, and the current fallback assigns `<sop>` to `bos_id`. That makes both direct chat construction and rendered server prompts tokenize the preamble as `<sop><sop>`.

The rendered-token table already maps `[gMASK]` to `bos_id`, so the fallback conflicts with the tokenizer's own intended mapping.

## Change

- Prefer the `[gMASK]` token when GLM BOS metadata is absent, retaining `<sop>` as a compatibility fallback.
- Append `<sop>` only when it differs from `bos_id`.
- Add a model-independent tokenizer test covering direct chat construction, rendered `[gMASK]<sop>` tokenization, and direct-path BOS/SOP deduplication.

Compatibility boundary: a tokenizer that lacks `[gMASK]` entirely still has different direct and rendered behavior. The direct path emits one `<sop>` after deduplication, while rendered `[gMASK]<sop>` still maps both literals to `bos_id == sop_id`. Official GLM Flash vocabularies contain `[gMASK]`; this PR retains the old `<sop>` fallback without redefining that no-`[gMASK]` rendered-template edge.

## Testing

Machine: Apple M5 Max, 128 GB, Metal.

Model used for the full suite: DeepSeek V4 Flash `Layers37-42Q4KExperts-OtherExpertLayersIQ2XXSGateUp-Q2KDown-AProjQ8-SExpQ8-OutQ8`.

```sh
make clean
make -j4
DS4_TEST_MODEL=/path/to/DeepSeek-V4-Flash-Layers37-42Q4KExperts-OtherExpertLayersIQ2XXSGateUp-Q2KDown-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix-fixed-0731.gguf make test
```

All tests passed, including the new `tests/test_tokenizer` target and the existing server, agent, Metal tensor-equivalence, sampling, and CLI suites.

The GGUF used during diagnosis is a genuine instance of this case: its raw KV header carries `general.architecture = glm-dsa`, `general.name = GLM 5.3 Flash`, and no `tokenizer.ggml.bos_token_id` at all. It is not loadable on current `main` for an unrelated reason: the `glm-dsa` arch tag routes it to the GLM 5.2 validator, which rejects its 46 blocks (`expected block_count=79 for GLM 5.2, got 46`), so tokenizer setup is never reached. The permanent regression therefore uses a synthetic GLM vocabulary and runs without a model file.

## Requirements

- I have read and agree with the contributing guidelines.
- AI usage disclosure: YES.


mvid:
Adding external evidence for the choice of `[gMASK]` as the fallback, since the PR previously argued this only from ds4's own rendered-token table.

Official reference tokenizer, [`zai-org/GLM-5.3-Flash`](https://huggingface.co/zai-org/GLM-5.3-Flash) `tokenizer_config.json`:

```json
{ "bos_token": null, "eos_token": "<|endoftext|>", "pad_token": "<|endoftext|>" }
```

`bos_token` is null upstream, which is why a converted GGUF can carry no `tokenizer.ggml.bos_token_id` at all. That is the case this PR fixes.

Where a converter does populate BOS, the canonical value is `[gMASK]`. GGUF metadata for [`unsloth/GLM-5.3-Flash-GGUF`](https://huggingface.co/unsloth/GLM-5.3-Flash-GGUF):

```
architecture:  glm5next
bos_token:     [gMASK]
eos_token:     <|endoftext|>
chat_template: [gMASK]<sop>\n...
```

So `[gMASK]` is the right fallback and `<sop>` is not, independently of ds4's internals. It also confirms the neighbouring `<|endoftext|>` EOS fallback already matches both sources.

To be explicit about scope: those Unsloth GGUFs do carry BOS metadata, so they take the metadata path and are *not* a reproducer for the fallback. They are cited here only as evidence of the canonical token. The regression test in this PR remains model-independent for that reason.

PR #941, opened 2026-09-01
fix(server): stream tool-enabled answers incrementally

apetersson:
## Motivation

Tool-enabled responses should reach clients as they are generated; buffering an ordinary answer until completion makes streaming appear broken and increases perceived latency.

## Summary

- limit the malformed second-reasoning-pass guard to one sampling update
- resume ordinary content streaming when a tool-enabled answer has no second reasoning marker
- preserve rerouting behavior when a real second reasoning close marker appears
- add regressions proving content arrives incrementally and before the final SSE event

Previously, an ordinary answer following reasoning could remain buffered until generation finished whenever tools were enabled.

## Testing

- `make -j ds4-server ds4_test`
- `./ds4_test --server`
- `git diff --check upstream/main...HEAD`

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -6332,6 +6332,7 @@ typedef struct {
     bool active;
     bool checked_think_prefix;
     bool guard_second_reasoning;
+    bool guard_second_reasoning_probed;
     bool sent_reasoning;
     bool sent_content;
     openai_tool_stream tool;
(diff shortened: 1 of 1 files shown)


apetersson:
Closing as a duplicate of #787, which addresses the same tool-enabled post-reasoning streaming regression and covers both the OpenAI and Anthropic streaming paths.

PR #944, opened 2026-09-02
server: accept inline images on tool/function role messages (fixes #933)

### LINKED ISSUE #933: ds4-server 400s valid OpenAI tool-role image messages as invalid JSON
## Summary

`ds4-server` rejects a well-formed OpenAI Chat Completions request that includes an inline PNG/JPEG data URI on a `role=tool` (or `role=function`) message. The HTTP status is 400 with `"invalid JSON request"`, even though the body is valid JSON.

This breaks OpenAI-compatible coding agents that attach tool-result screenshots as `image_url` content parts on the tool message, which is the normal Chat Completions shape. A `role=user` data URI of the same image is accepted.

## Repro

`POST /v1/chat/completions` with messages:

1. `role=user` text
2. `role=assistant` with a `tool_calls` entry
3. `role=tool` whose `content` is an array of `{type:text}` plus `{type:image_url, image_url:{url:"data:image/png;base64,..."}}` (or jpeg)

Observed: `400 {"error":{"message":"invalid JSON request","type":"invalid_request_error"}}` before vision encode.

The same data URI on `role=user` returns 200.

## Cause

In `parse_messages` / `parse_anthropic_messages`, any message with images whose role is not exactly `"user"` fails parse. That path always reports `"invalid JSON request"`.

The user-only restriction matches the GLM note that image tokens should land on a user-shaped turn, and `ds4-agent` already remaps `view_image` onto a user turn. The server still needs to accept the OpenAI wire format: DeepSeek already wraps tool results in a user turn (`<｜User｜><tool_result>…`).

File paths and remote URLs should stay rejected.

## Suggested fix

Allow inline images on `user`, `tool`, and `function`. Keep rejecting assistant/system images, paths, and `https` URLs. Prefer a specific error string for remaining schema failures instead of `"invalid JSON request"` (optional, not required for the role check).

I can send a PR with a parser unit test for the tool-role shape.

Flor1an-B:
Fixes #933. Credit to @aj-img for the precise diagnosis and suggested fix there — this PR implements exactly what was proposed.

## Problem

`parse_messages` rejected any OpenAI Chat Completions message carrying an inline image whose `role` wasn't exactly `"user"`, always with a generic `400 "invalid JSON request"`. That breaks the normal shape for coding agents that attach a tool-result screenshot as an `image_url` content part on a `role: "tool"` (or `role: "function"`) message — the same data URI on `role: "user"` was accepted.

## Fix

As diagnosed in #933: the user-only restriction was meant to keep image tokens landing on a user-shaped turn, but DS4's own chat template already renders tool results inside a user turn (`<|User|><tool_result>...`), so the API-level role tag doesn't need to match that internal shape. One line in `parse_messages`: allow `user`/`tool`/`function`, keep rejecting `assistant`/`system`, file paths, and `https` URLs exactly as before (`parse_anthropic_messages`'s identical-looking check is untouched — Anthropic's wire format has no separate tool role; `tool_result` blocks always nest inside a `user`-role message already, so it isn't exposed to this bug).

## Testing

- Added `test_openai_tool_role_image_content`, mirroring the existing `test_openai_inline_image_content` / `test_http_image_paths_and_urls_are_rejected` pair: asserts `user`/`tool`/`function` all parse and capture the image, and that `assistant`/`system` still fail.
- Verified it's a real regression test, not a happy-path-only check: reverted the fix locally, confirmed the new test fails at the exact new assertions, restored it, confirmed green.
- Verified end-to-end past the parser: built `ds4-server` with the fix, sent the exact three-message shape from the issue (`user` text → `assistant` with `tool_calls` → `tool` message with a real screenshot as `image_url`) over HTTP — `200`, and the model's answer correctly described the screenshot's actual on-screen content (not a hallucination). Confirmed the two negative cases still `400`: an image on `role: assistant`, and an `https://` URL on `role: tool`.
- Full `make test` green throughout, on `main` @ current head.

## Scope

The issue's "prefer a specific error string for remaining schema failures" suggestion is explicitly marked optional there and left out, to keep this to the one behavioral fix.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -2028,7 +2028,8 @@ static bool parse_messages(const char **p, chat_msgs *msgs) {
         (*p)++;
         if (!msg.role) msg.role = xstrdup("user");
         if (!msg.content) msg.content = xstrdup("");
-        if (msg.images.len && strcmp(msg.role, "user")) goto fail;
+        if (msg.images.len && strcmp(msg.role, "user") &&
+            strcmp(msg.role, "tool") && strcmp(msg.role, "function")) goto fail;
         chat_msgs_push(msgs, msg);
         memset(&msg, 0, sizeof(msg));
         json_ws(p);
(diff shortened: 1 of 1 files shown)


pswai:
Tested on top of current `main` (0aaea5a) on an M4 Max 64 GB with `Qwen3.8-Flash-Next-Q2.gguf --vision`: the diff applies cleanly, `DS4_SERVER_TEST` passes, and an `image_url` part inside a tool-role message now returns 200 with a correct description instead of `400 invalid JSON request`. This fixes a real client (Hermes Agent's `vision_analyze` tool sends exactly this shape). Details and a minimal three-message reproduction in #933. Running this as my daily server now. Would be great to see it merged.

Flor1an-B:
Bumping given the independent confirmation above — different hardware (M4 Max vs my original M5 Max), a real downstream client that depends on this (Hermes Agent's `vision_analyze` tool sends images on tool-role messages as the normal OpenAI wire format), and it still applies cleanly on current `main`. Happy to rebase again if anything's changed since. Whenever you get a chance to take a look, thanks!

PR #950, opened 2026-09-02
ds4-server: make OpenAI-compatible image input work end to end

Bendr0id:
Analyzing an image over the OpenAI chat completions API failed even though /read in the agent handled the same file, and every failure surfaced as a generic "invalid JSON request". Three independent causes, each sufficient on its own to break it:

The data URI parser matched a fixed set of literal prefixes and the base64 decoder accepted only canonical input, so payloads real clients emit were refused: varying scheme and media-type case, parameters such as ";charset=" before ";base64", MIME line breaks, the URL-safe alphabet and omitted padding. Rejections now explain themselves instead of collapsing into one opaque 400, and an untyped content block carrying an image_url is treated as an image rather than silently dropped, which had answered the prompt from the text alone as if no image had been sent.

/v1/models never advertised image input, so clients that gate on modality stripped the image before it left the client. Model metadata now carries architecture.input_modalities when an encoder is loaded. Clients also commonly suppress image input for anything they classify as DeepSeek, because the upstream API is text-only, and re-enable it only when the model id carries a "vision" token, so deepseek-v4-flash-vision and deepseek-v4-pro-vision are offered too. Both are reported only when started with --vision, including the per-model endpoint.

An image the server cannot decode is replayed in the transcript on every later turn, so a single WebP wedged a conversation permanently: every request failed regardless of what was actually being asked. An undecodable image now becomes a text note in the history and only the newest message rejects the request, since that is the one the caller just attached and needs to be told about. Dropping an image also removes the marker that referenced it, which rendering would otherwise emit into the prompt as literal control-character junk.


hertz-hwang:
Second independent report on the masking problem, from a harness-driven client rather than the DeepSeek-gating case in the description — plus one path this branch still leaves generic.

**Why the masking matters from here.** The client is a coding-agent harness on `api: openai-responses`. Full-screen macOS captures (3456×2234, over its 2048² normalization budget) get re-encoded to WebP for the alpha plane and were refused as `400 invalid JSON request`; window crops of the same screen stayed PNG and worked. Two hours went into body-size limits before the container turned out to be the axis. Anything that turns a payload-shape problem into a JSON-syntax report costs that same time to every caller — so the direction here is right, and `GIF images are not supported; send a JPEG or PNG` is a much better answer than what `main` gives.

**Verified on this branch** (`68c4306`, Apple M3 Max, macOS, Metal). `make ds4_test` builds warning-free with `-Wall -Wextra`. I then built a small probe against this `ds4_server.c` and drove the parsers directly:

| case | result |
| --- | --- |
| chat, `data:image/gif;base64,…` | `GIF images are not supported; send a JPEG or PNG` |
| responses, gif inside an item `content` | `GIF images are not supported; send a JPEG or PNG` |
| anthropic, `media_type: image/gif` | `GIF images are not supported; send a JPEG or PNG` |
| chat, **MIME line-wrapped base64** (newline every 76 chars) | reaches `image input requires starting ds4-server with --vision FILE` — i.e. accepted |
| responses, gif inside an item **`output`** | `invalid JSON request` ← still masked |
| truncated JSON `{"messages":[{"role":"user"` | `invalid JSON request` (correct) |

Two notes on that table:

**The wrapped-base64 row is the quiet win.** On `main`, `server_decode_base64()` rejects a length that is not a multiple of 4 and any non-alphabet byte, so a payload carrying a trailing newline is refused as malformed JSON. I hit exactly that by accident while testing (a shell here-string appended `\n`), and it sent me down a size-limit theory for a while — which is a decent proxy for what a real client with a line-wrapping encoder or an HTTP layer that folds headers-style will experience. Worth a named test so it does not regress: `data:image/png;base64,` + base64 broken by `\n` every 76 chars, asserting it decodes.

**The `output` row looks like an oversight rather than a policy choice.** `parse_responses_input()` mirrors the diagnostic at the `content` key but not at the `output` key — `ds4_server.c:5071`:

```c
} else if (!strcmp(key, "output")) {
    server_image_inputs_free(&output_images);
    json_ws(p);
    if (**p == '[') {
        if (!parse_responses_content_array_multimodal(p, &output, &output_images)) {
            free(key);
            goto item_fail;          /* output_images.error never mirrored */
        }
```

so a `function_call_output` / `custom_tool_call_output` item carrying an unreadable image is still reported as a broken request. The same three lines used at the `content` site should close it:

```c
        if (output_images.error[0])
            snprintf(msgs->image_error, sizeof(msgs->image_error), "%s", output_images.error);
```

Worth noting for anyone implementing the same idea elsewhere: `parse_responses_content_array_multimodal()` frees the container on the failure path *before* returning, and `server_image_inputs_free()` ends in `memset()`, so the diagnostic has to be carried across that reset. I found that trap the slow way — every propagation site looked correct and the message was still generic. This branch already handles it in the wrapper; the `output` site above is the one place it does not currently pay off.

*(Scope: parse-level cases only, driven directly against the parsers. I did not run a live server on this branch, and did not exercise the vision-gated model aliases or `input_modalities` advertisement it also adds.)*

PR #954, opened 2026-09-02
metal: pre-M5 optimizations — +3.65% MXFP4 prefill on M3 Ultra

ivanfioravanti:
Metal performance work for pre-M5 GPUs (M1–M4 resident path) that was benchmarked during the Aug decode campaign but never landed on `main`. All changes are guarded by rollback env switches and produce **bit-identical output**.

Main pieces:
- **Greedy chain decode** (`metal_graph_greedy_chain`): GPU argmax feeds the next token via a device ring, removes the per-token `waitUntilCompleted` + 517 KiB logits readback + CPU argmax boundary.
  Also wired into the session API (`ds4-eval`).
- **Raw-layer gathered attention**  the path cost ~65 µs vs ~32 µs per layer.
- **Prefill**: indexer-query pruning (pre-M5 + M5 port), staged indexed prefill attention rows (RB4 / heads16-dual), long-prefill acceleration, batch Q/KV finalizer overlap, MoE sum6→HC4 and Q8 attn-out→HC4 epilogue fusions.
- **Tooling**: commit-only GPU stage-counter profiler; DSpark decode-campaign measurements and handoff docs in `speed-bench/`.

## Performance (branch vs main, fresh builds, interleaved runs)

Hardware: **Apple M3 Ultra, 512 GiB** · Model:
`DeepSeek-V4-Flash-MXFP4…-chat-v2-mxfp4-0731.gguf`

### Greedy decode

`./ds4 -m <mxfp4.gguf> -p "Write a short story about a lighthouse keeper." -c 8192 -n 128 --temp 0`

| build | run 1 (t/s) | run 2 (t/s) |
|---|---|---|
| main | 42.49 | 43.93 |
| this branch | **46.14** | **46.14** |

**+5.0–8.6% tokens/s.** Both builds and both runs produced transcript md5 `db0c504c8203618552e685bd2c701e4f` — the campaign oracle hash, confirming bit-exactness end to end.

### Prefill

`./ds4-bench -m <mxfp4.gguf> --prompt-file speed-bench/promessi_sposi.txt --ctx-start 2048 --ctx-max 8192 --step-incr 2048 --gen-tokens 128`

| ctx tokens | main tok/s | branch tok/s | Δ |
|---|---|---|---|
| 2048 | 638.4 | 646.9 | +1.3% |
| 4096 | 578.6 | 604.3 | +4.4% |
| 6144 | 571.0 | 595.9 | +4.4% |
| 8192 | 566.4 | 589.1 | +4.0% |

First-token latency at ctx 2048: 28.1 ms → **24.3 ms**. Steady-state generation at longer contexts is equal or slightly better (39.33 → 39.47 t/s at 8K).

## Correctness evidence

- Transcript md5 oracle (`db0c504c…`) matches on both builds (above).
- Per-kernel A/B campaigns on this branch compared **millions of full-vocabulary logits bit-identical** against rollbacks, plus forced direct-kernel oracles (details in `speed-bench/README.md` on this branch).
- `make test` 44/44 at campaign close; prefix-to-decode checks matched all selected token IDs.

## Rollback switches

Every optimization is individually disableable, e.g. `DS4_DISABLE_GREEDY_CHAIN=1`,
`DS4_METAL_DISABLE_DECODE_RAW_GATHERED_ATTN`,
`DS4_METAL_DISABLE_DECODE_RAW_PACKED32`,
`DS4_METAL_DISABLE_DECODE_MOE_HC_FUSION`,
`DS4_METAL_DISABLE_PRE_M5_INDEXED_ATTN_PREFILL_RB4`,
`DS4_METAL_DISABLE_PRE_M5_INDEXED_ATTN_PREFILL_HEADS16_DUAL_RB4`,
`DS4_METAL_DISABLE_PRE_M5_BATCH_MOE_SUM6_HC_FUSION`,
`DS4_METAL_DISABLE_PRE_M5_BATCH_ATTN_OUT_HC_FUSION`.


JordiPosthumus:
Ivan, I tested the raw-only attention portion on an M3 Ultra / 512 GiB. There is a useful kernel-level gain here, and one small Metal API validation issue worth fixing.

**Scope:** I extracted the raw-only dispatch, staging and packed32 eligibility changes from `e04243b93f717038109756c6633cc1e0ebe20d33` onto our existing patched M3 build. These are synthetic direct-kernel comparisons against the original raw path, not a whole-PR-versus-main or real-model benchmark. I did not exercise the greedy chain or prefill changes.

**Binding issue:** with `n_comp=0` and `comp=nil`, the newly admitted fused staging path still binds `comp` directly. With `MTL_DEBUG_LAYER=1`, the first candidate call aborts:

```text
Compute Function(kernel_dsv4_flash_kv_stage_f16): missing Buffer binding at index 2 for comp_src[0].
```

Repro shape: `ds4_gpu_attention_decode_heads_tensor`, 64 heads × 512, raw capacity 2304, one visible raw row, no compressed tensor/rows, no mask, quality mode off. The original raw dispatch passes. The staging guard correctly allows the empty compressed side; Metal still requires a binding even though that side is not read. Using the same dummy-binding approach already used for mask/pad resolves the assertion:

```objc
[enc setBuffer:(n_comp ? comp : raw)
         offset:(n_comp ? comp_offset : raw_offset) atIndex:2];
```

With that adjustment, **88 direct cases were bit-exact** against the original path: raw lengths 1–512 including 31/32/33, 127/128 and 255/256/257; contiguous/wrapped rings; nonzero tensor-view offsets; both quality modes; and fused inverse RoPE versus the original standalone tail. Removing only the dummy binding reproduced the validator abort again.

For the eligible packed/fused path, median command-buffer GPU time per operation was **31.60 → 19.80 µs at 128 rows** and **33.13 → 23.06 µs at 256 rows** (10 alternating A/B rounds, 32 operations per buffer; Metal validation enabled). At 512 rows the result was effectively equal; non-fused measurements were mixed. These numbers establish a scoped kernel benefit, not a model tokens/s gain. Our long-context Flash workload has only two persistently raw-only layers, so I would expect a much smaller whole-model benefit from this part alone.

We have not installed the extraction or run a full-model/full-suite validation of this PR. Sharing the concrete binding fix and bounded positive result here rather than opening a competing optimization PR. Thanks for making the individual rollback switches available—they made the comparison straightforward.

JordiPosthumus:
Ivan, following up with measured results for the short-prefill MXFP4 dispatch portion of this PR. A focused extraction produced a repeatable 6–18% throughput gain on the measured short-prefill cases, including short suffixes at long context.

The dispatch change lets existing optimized paths, previously selected by default for at least 2,048 new tokens, also handle eligible 32–2,047-token prefills. This matters when most of a conversation is cached but its next turn adds only a short suffix.

The reference is our existing patched baseline `960e973`, compared with the same source plus only a 26-line dispatch delta in `ds4_metal.m` (16 additions, 10 deletions). It enables the existing compact tile, scatter map, tail culling and half-result lookup paths for eligible 32–2,047-token prefills, preserving the existing guards and adding one aggregate rollback. No kernel bodies, model math, server code or cache policy differ between these builds. This evaluates an adaptation of your work; it is not a whole-PR-versus-main result.

Benchmark setup: Apple M3 Ultra / 512 GiB, Metal; DeepSeek-V4-Flash-Vision-Exp with MXFP4 experts, F16 HC/compressor/indexer, Q8 attention/shared/output and its Vision encoder. Context 262,144, ten actual resident allocations, one active request, prefill chunk 4,096.

I used independent reference/candidate/candidate/reference processes, one excluded warm-up and four measured passes per mode. Each candidate process alternated enabled/rollback order. Rates are new tokens divided by timed engine-prefill duration, excluding tokenization, snapshot restoration and HTTP overhead. The table compares old/new dispatch within the candidate processes: eight paired observations per shape across two independent processes, not eight independent machine trials.

| New tokens | Cached tokens | Old dispatch tok/s | New dispatch tok/s | Paired gain |
|---:|---:|---:|---:|---:|
| 512 | 0 | 426.13 | 496.11 | +16.42% |
| 2,047 | 0 | 588.21 | 625.33 | +6.31% |
| 2,048 (control) | 0 | 644.29 | 643.88 | −0.06% |
| 91 | 110,290 | 118.21 | 139.31 | +17.85% |
| 128 | 261,888 | 134.53 | 156.03 | +15.98% |

For the 91-token continuation, this means approximately 0.77 → 0.65 seconds to process the new prompt tokens. These are prefill-throughput gains; they do not describe the total response latency.

The independent references agreed closely for the first four shapes, with before/after drift below 0.32%. The final near-capacity reference slowed 4.76%, inflating that pooled comparison to 19.61%; the paired result above avoids that comparison. The two candidate processes separately gained 16.07% and 15.90% there. Ordinary decode has no demonstrated speedup from this extraction.

Validation:

- Independent GPU oracles covered 100 boundary/routing/quality cases per build with four ordered rollback/on calls, guarded buffers and exact intermediate/output comparisons. Independent output dumps matched byte-for-byte; the normal Makefile regression target passed too.
- The initial vector plus 128 fixed teacher-forced continuation steps—129 full-vocabulary vectors of 129,280 scores—matched on every repeated text run and across builds, through the full 262,144-token boundary. A 642-token image-containing suffix after 32,768 cached text tokens also matched, including 33 full-vocabulary vectors.
- Ten live sessions, cancellation/replay, conversation isolation, appended/changed images, tool histories, capacity rejection and text/image checkpoint reuse in both build directions passed. An initial fixture incorrectly combined twelve conversations on ten slots; the corrected order passed with unchanged assertions and runtime.
- Native Metal and CPU builds plus focused server, agent and session-state tests passed. The full aggregate suite was not run.

The focused patch/test fixture passed an applicability check at main `9ab7053`, which contains the required kernels. I have not built or model-tested a new branch based on that head. The patch and regression fixture are ready to share if useful; I would coordinate any split with you. Thanks for making the changes attributable through rollback controls.

ivanfioravanti:
Great @JordiPosthumus I'm rebasing the branch and see how it works against the new main.

ivanfioravanti:
Rebased onto current `main` (`c0a6119`, +46 upstream commits including the Metal expert-caching and GLM/IQ2 prefill work) and re-ran the full comparison on the same M3 Ultra / 512 GiB / MXFP4 0731 setup. Clean detached worktree per commit, `make -j8`, ABBA order (main / branch / branch / main), one minute idle between sweeps, no warmup or diagnostic switches.

## Prefill (`ds4-bench`, promessi_sposi, 2K–8K, two-run pooled t/s)

| ctx tokens | main tok/s | branch tok/s | Δ |
|---:|---:|---:|---:|
| 2048 | 639.80 | 661.88 | +3.45% |
| 4096 | 584.88 | 608.05 | +3.96% |
| 6144 | 577.46 | 600.07 | +3.92% |
| 8192 | 571.44 | 593.16 | +3.80% |
| **pooled** | **592.26** | **614.63** | **+3.78%** |

Classic `ds4-bench` steady decode is flat (−0.1% to +0.4%); first-token evaluation at 2K unchanged (24.3–24.9 ms both builds, no claim made).

## Greedy decode (CLI, `--temp 0 -n 128`)

| build | run 1 (t/s) | run 2 (t/s) |
|---|---|---|
| main | 43.13 | 43.18 |
| branch | **45.74** | **45.66** |

**+5.9%**, same as before the rebase.

## Correctness

- All four decode transcripts md5 `db0c504c8203618552e685bd2c701e4f` (the campaign oracle), both builds.
- Untimed `--dump-frontier-logits-dir` runs (2K/4K frontiers): dumps byte-identical between builds.
- `make test` green on both trees; the branch's direct Metal oracles (static mixed attention, indexer leading query, MXFP4 short-prefill bundle, session batch vs serial) all pass.

## Review follow-ups

- @JordiPosthumus the nil-`comp` binding abort you reported is fixed on the branch (dummy raw binding at index 2 when `n_comp == 0`, mirroring the mask/pad trick); the md5 oracle above exercises it through the real graph. Thanks for the precise repro.
- The short-prefill MXFP4 dispatch extension you measured (+6–18% on 32–2,047-token prefills) is already in this branch since `d07ee2d` (`use_pre_m5_mxfp4_small_prefill_defaults`, aggregate rollback `DS4_METAL_DISABLE_PRE_M5_MXFP4_MOE_SMALL_PREFILL`) — your independent validation of the same change against `main` is much appreciated, and your extraction matches the shipped guards.

Raw CSVs, per-run samples and protocol: `speed-bench/m3_ultra_mxfp4_20260907/` on the branch (main-r1/branch-r1/branch-r2/main-r2.csv, summary.json, protocol.json); the September 6 comparison against `9ab7053` remains in `speed-bench/m3_ultra_mxfp4_20260906/`.

ivanfioravanti:
@JordiPosthumus your suggestion implemented here. @antirez ready to merge!

PR #959, opened 2026-09-03
metal: prune top-k argsort merge rounds to top_k

emilianbold:
## Summary

`kernel_argsort_merge_f32_i32` carried `work_width` elements per row through every intermediate merge round, so the full row was re-merged until the final merge clamped to `top_k`. It also divided thread work by the number of elements *read* rather than written, so once a round truncates, the high threads exit immediately and whole SIMD groups idle.

This PR prunes each round to `top_k` and splits the work over the outputs that actually exist. Same result, less work, better occupancy.

The work was started with DeepSeek Flash on ds4 then refined with Claude. 

## The change

`metal/argsort.metal`:

- Reads walk the input runs at stride `2*len`, bounded by a new `args.total` (the valid elements per row) instead of `args.ne0` (the buffer row stride).
- Writes pack at `start_write = im * new_len` where `new_len = MIN(2*len, keep_k)`, so a round keeps at most `top_k` per pair.
- The work split keys off `out = MIN(total, new_len)`, and one guard (`k0 >= out`) replaces the two old early returns.
- Two new kernel args: `total`, `keep_k`.

`ds4_metal.m` (`ds4_gpu_indexer_topk_tensor`):

- Tracks `len` / `total` / `nruns` per round and drives the loop off the run count rather than `len < work_width`.
- `total_next` is O(1): full pairs write `new_len`, the last partial pair writes `MIN(r, new_len)`, with `q` clamped to the dispatched pair count.

Odd run counts need no special case: the last run has no partner, which falls out of `len1 == 0`, and the merge copies it through.

## Why it is correct

Runs stay packed at stride `len` with every run full but the last, so:

- `total >= top_k` is preserved every round;
- the final round has `nruns == 2`, hence `total <= 2*len`, hence `2*len >= top_k`, hence `new_len == top_k` and exactly `top_k` elements are written to the output tensor;
- writes are bounded by `total_next <= total <= work_width`, so they stay in the row.

Truncating a pair to `new_len` is exact rather than approximate: a dropped element had `>= top_k` elements above it inside the pair, hence `>= top_k` in the whole row, so it was not in the top-k.

## Performance

Merge-element writes per row, across all rounds, at `nth = 1024`:

  | shape (n_comp / top_k) | npr | before | after | ratio |
  |---|---:|---:|---:|---:|
  | 8192 / 512    |  8 |   8704 |  3584 | 2.43x |
  | 16384 / 1024  | 16 |  50176 | 15360 | 3.27x |
  | 32768 / 2048  | 32 | 133120 | 63488 | 2.10x |
  | 8192 / 2048   |  8 |  18432 | 14336 | 1.29x |

Measured merge-GPU time per call, median over 7 runs on an idle GPU:

  | shape | n_tokens | before | after | speedup |
  |---|---:|---:|---:|---:|
  | 8192 / 512   |   1 |   13.5 us |  11.0 us | 1.22x |
  | 8192 / 512   | 256 |   76.1 us |  59.4 us | 1.28x |
  | 16384 / 1024 |   1 |   28.3 us |  16.3 us | 1.74x |
  | 16384 / 1024 | 256 |  402.2 us | 175.3 us | 2.29x |
  | 32768 / 2048 |   1 |   49.1 us |  23.5 us | 2.09x |
  | 32768 / 2048 | 256 | 1265.4 us | 560.9 us | 2.26x |

Two caveats on the measurement. It was taken with the merge rounds isolated in their own command buffer, which is not how they are encoded in production, and the harness that does that is not part of this PR.

End to end this is modest in an ordinary context: the saving is about 0.10 ms/token at Flash 32k summed over 43 layers, roughly 0.3% of a ~31 ms token. It grows with context though.

## Testing

`make test-argsort-metal` runs `tests/test_argsort_metal.c`


JordiPosthumus:
Tested `ea16db6` on an Apple M3 Ultra (80-core GPU, 512 GiB). My conclusion is that this is a useful exact optimisation, and it complements #832 rather than duplicating it: #832's stream-512 path only handles `top_k == 512 && n_tokens >= 32`, while this still shortens the merge cascade for decode (`n_tokens == 1`), small prefills, and other `top_k` values.

Correctness checks:

- The PR's seven Metal cases pass after the link fix below.
- Integrated on top of #830/#831/#832, the default path, canonical full argsort, and an independent CPU `(score desc, index asc)` reference matched for all 135,136 selected indices. This includes dense finite ties, odd row widths, the 31/32-token dispatch boundary, a 64K-context-shaped row, and the non-512 fallback.
- Separate baseline/candidate decode-shaped runs at 8,192, 16,384, 32,768, and 65,536 components produced byte-identical selected-index buffers for both ordinary rounded-float input and deliberately dense ties.

I have not repeated the PR's end-to-end model throughput measurement yet, so this is an independent correctness/integration result rather than a second net-speed claim.

One current-head issue: `make test-argsort-metal` fails to link on `b0a147a` because `ds4_metal.o` references `ds4_deepseek4_attention_bounds`, but the target does not link `ds4_image.o`:

```
Undefined symbols for architecture arm64:
  "_ds4_deepseek4_attention_bounds", referenced from:
      _ds4_gpu_fill_visual_mixed_batch_mask in ds4_metal.o
```

Adding `ds4_image.o` to that target fixes it; the rebuilt test then passes all cases. The isolated one-line commit is [6c19363](https://github.com/JordiPosthumus/ds4/commit/6c19363506e39c5386709e037dc88ca30b94a2e2) if useful.

JordiPosthumus:
Follow-up with a controlled real-model run on the exact current head (`ea16db6`) using an M3 Ultra (512 GiB) and DeepSeek V4 Flash Vision-Exp MXFP4/F16/Q8.

Correctness remained exact:

- full frontier-logit JSONs were byte-identical at 32K, 64K, and 128K contexts;
- a separate 64-step teacher-forced decode after a 32K prefill compared every logit (64 x 129,280 floats, 33,095,680 bytes) and was byte-identical, SHA-256 `fdef848d3bcb50ddb75d069842b0e9be42678543ac2e700b1e0fc4581441c871` in both builds;
- the focused Metal argsort suite and the broader canonical/default/CPU top-k matrix both pass.

Whole-model timings had a strong second-run effect, so I used ABBA order (baseline, candidate, candidate, baseline) rather than comparing unrelated absolute runs. Candidate decode won both pairs at every measured frontier; the geometric mean of the two paired ratios was +0.68% at 32K, +2.80% at 64K, and +5.01% at 128K. Prefill numbers are not evidence for this patch because our `top_k=512`, `n_tokens>=32` prefills take the separate stream512 path.

To isolate the affected fallback argsort from that run-to-run drift, I also used an evaluation-only same-process switch that alternated legacy/pruned merge widths in five `L P P L` quartets on the same pipeline and buffers. Median affected-kernel gains across 20 blocks were:

| candidates | legacy | pruned | gain |
| ---: | ---: | ---: | ---: |
| 8,192 | 319.322 us | 318.911 us | +0.13% |
| 32,768 | 355.230 us | 343.509 us | +3.41% |
| 65,536 | 388.570 us | 347.853 us | +11.71% |

Every legacy/pruned block produced the same selected-index checksum. The measurement switch is not proposed code; it was only used to eliminate cross-process GPU-state noise.

Conclusion from this machine/model: the optimization is lossless, small at shorter histories, and increasingly useful for long-context decode. We have installed it locally in production. Thanks for the focused improvement.

emilianbold:
@JordiPosthumus :

> One current-head issue: make test-argsort-metal fails to link on b0a147a because ds4_metal.o references ds4_deepseek4_attention_bounds, but the target does not link ds4_image.o

Thanks. This is due to a change introduced in main after working on the branch. Fixed now.

emilianbold:
I rebased this PR since it had conflicts after @antirez pushed DeepSeek 4.1, which has a change for top_k too.

I also have a 3rd refactoring commit which is optional. I think it makes things simpler to read, but changes more lines so it's up to the reviewer to choose.

@JordiPosthumus could you perhaps look at this PR again?

Note that DeepSeek 4.1 top_k uses `causal_start`/`causal_ratio` but not always (only in prefill?). Which means this PR would speed up DeepSeek 4.1 too, in all the other uses of top_k.

JordiPosthumus (commented):
Re-reviewed b4605a0919ac99a3aeba3242c52b5bc57f816e33 after the rebase, on an Apple M3 Ultra (512 GiB), Metal backend.

I found no correctness regression in the paths reviewed. The non-causal merge keeps at most top_k entries per run and accounts for the odd tail; the causal branch retains the original run offsets, prefix read limit and work distribution. I would keep the optional helper refactor: it preserves the path-specific arguments and thread caps while removing duplicated dispatch code.

You're right about the DS4.1 call sites: decode uses the generic non-causal top-k entry point, while prefill uses the causal batch entry point. The pruning therefore applies to DS4.1's non-causal calls too. I haven't measured an end-to-end DS4.1 speedup.

Validation in an isolated checkout:
- `make -j8` and `make test-argsort-metal`: passed (all 7 supplied cases).
- An expanded CPU-reference fixture: all 18 cases passed, including 12 non-causal cases and 6 causal cases with ratios 1 and 2, row widths crossing 1024/2048, odd tails and widths through 65537. Scores were unique; this does not validate tie ordering.
- The same 18 cases passed with both `DS4_METAL_DISABLE_V41_TOPK_PREFIX=1` and `DS4_METAL_DISABLE_V41_TOPK_SHUFFLE=1`.
- Both expanded runs also passed at e8c84dd, before the optional refactor.

One test addition I'd suggest: include causal coverage in `test_argsort_metal.c`, since the current committed test calls only `ds4_gpu_indexer_topk_tensor`. For example, call `ds4_gpu_dsv41_indexer_topk_batch` with `(width, rows, start, ratio)` = `(1056, 32, 1023, 1)` and `(1056, 32, 2047, 2)`, and compare each row against the CPU sort of its valid causal prefix. Repeat with the prefix/shuffle switches disabled. These exercise the rebase boundary that the existing suite misses.

No model was loaded and the production service was not changed. This was focused kernel correctness validation, not a full regression suite or model benchmark.

PR #960, opened 2026-09-03
server: restore the prompt frontier after cancelled generation

JordiPosthumus:
## Summary

Preserve the synchronized prompt when a client cancels generation, and retain
an exact OpenAI retry binding when visible replay differs from sampled tokens.

## Why

Generation mutates more than a token count: raw KV is a ring and compressor/
indexer accumulators change in place. Truncating the token list cannot recover
the prompt after those values have been overwritten. The request checkpoint
copies the mutable frontier; compressed history remains in place and is bounded
again by its saved counters. A checkpoint belongs to one session and is retired
before destructive history reconstruction. In the speculative
fallback, retirement now occurs before the rebuilding sync can overwrite
compressed history, on both success and failure; an in-place rewind retains
the request checkpoint.

Restoring KV is also insufficient if the client omits prior hidden reasoning.
The retry binding stores the rendered client prompt separately from the complete
restored model frontier. A retry must match the exact request identity, rendered
tokens, frontier tokens and image state, and dispatch must find one unique owner.
It does not re-append consumed tool results. Failed restoration clears the
binding. Successful response delivery is not undone by a late client FIN.

After rollback, a continued-save counter beyond the restored prompt is capped at
that prompt; lower counters are left alone. Otherwise a checkpoint written by
the discarded tail can suppress the next retry-tail checkpoint.

## Compatibility

Tensor rollback is limited to local DeepSeek Metal sessions. Unsupported or
invalidated rollback uses conservative invalidation; an untouched valid prompt
can be preserved without a snapshot. No model math, sampling, cache format or
arbitrary rewind change. The binding is exact, resident-only and uses existing
slot state and locks; it is not general history matching.

## Practical impact

The benefit is recovering a reusable prompt after cancellation and recognizing an exact retry without rebuilding that prompt. The tensor rollback applies to supported Metal sessions; this is not a CUDA rollback or general rewind feature. Snapshot creation has a cost, with earlier time and memory measurements below. Correctness was tested on the standalone patch, but net retry-latency benefit against `f62ca29` has not been timed.

## Combined runtime context

## Validation

One standalone commit, `6b7d00838e8e`, directly on upstream `f62ca29a3087` (2026-09-07). The new 32-case regression extracts the actual speculative-rebuild handoff: fast rewind retains rollback; successful and failed destructive rebuilds retire it before history replacement. Server tests cover exact retry ownership, cancellation/response races and continued-save counters. Upstream EOS coverage also passes.

Before the Metal-only upstream update, the standalone patch also passed `DS4_TEST_MODEL="$MODEL" DS4_TEST_BACKEND=metal ./ds4_test --cancel-checkpoint` on M2 Max with DeepSeek-V4-Flash-Vision-Exp IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8: all 607 full-vocabulary vectors matched after deliberately overwriting the raw ring; foreign-session restore was rejected. The fixture uses 6,144 context and 256-token prefill for adversarial coverage. Those are test settings. Earlier M3 MXFP4 measurements found about 2.3 ms and 22.39 MiB per request checkpoint at 8,193 prompt tokens, with allocations reclaimed on free; they are not new rebase timings. #927 overlaps reuse dispatch and must preserve exact retry priority.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -695,8 +695,12 @@ tests/test_prompt_prefix.o: tests/test_prompt_prefix.c ds4_prompt_prefix.h
 tests/test_prompt_prefix: tests/test_prompt_prefix.o ds4_prompt_prefix.o
 	$(CC) $(CFLAGS) -o $@ $^
 
+.PHONY: test-server-cancel-rebuild
+test-server-cancel-rebuild:
+	python3 tests/test_server_cancel_rebuild.py
+
 .PHONY: test-frontends
-test-frontends: ds4_test ds4_agent_test
+test-frontends: ds4_test ds4_agent_test test-server-cancel-rebuild
 	./ds4_test --server
 	./ds4_agent_test
 
--- a/ds4.c
+++ b/ds4.c
@@ -54505,6 +54505,29 @@ struct ds4_session {
 };
 
 static bool ds4_session_tp_leader(const ds4_session *s);
+struct ds4_cancel_checkpoint {
+    ds4_session *owner;
+    token_vec checkpoint;
+    ds4_vision_identity *images;
+    size_t image_count;
+    float *logits;
+    int pos;
+#ifndef DS4_NO_GPU
+    /* Compressed rows are append-only: restoring their row counts makes
+     * request-era rows unreachable.  The raw cache is a ring, so generation
+     * can overwrite prompt-era rows and its logical live window must be
+     * copied.  The small compressor accumulators are mutable in place. */
+    uint32_t raw_live;
+    uint32_t raw_cap;
+    uint32_t n_comp[DS4_MAX_LAYER];
+    uint32_t n_index_comp[DS4_MAX_LAYER];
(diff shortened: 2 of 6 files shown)


JordiPosthumus:
The recent upstream vision-continuation change (81a76585ff44e0494053fca9101130e2535648b9) overlaps this PR's reuse dispatch, so this branch needs rework against current main before it is ready to merge.

The remaining purpose is unchanged: restore the mutable prompt frontier after cancelled generation and retain an exact cancelled-request retry binding. Upstream's live-vision continuation and agent compaction recovery do not replace that behavior.

A future rework should preserve upstream's authenticated image continuation and slot selection while retaining cancellation-checkpoint ownership/invalidation rules, exact retry matching and the adversarial cancellation tests. Leaving this open for now; rework is deferred, with no commitment to a follow-up.

PR #961, opened 2026-09-03
server: persist vision KV with exact conditioning provenance

JordiPosthumus:
## Summary

Persist image-conditioned KV under a key containing the exact conditioning that
produced it, and retain that provenance through prefix reuse.

## Why

Equal text tokens do not imply equal KV when images contributed embeddings. The
key therefore includes image count, token spans, hashes of the final
conditioning vectors and canonical token text. Loading reattaches image metadata
only when the key and restored frontier agree.

A retained prefix still contains its old conditioning even if a fresh encoder
call produces different candidate bytes. Sync reports what was actually kept:
wholly retained images keep their old hashes, rebuilt images use new hashes, and
partial or unknown provenance cannot be persisted. A GLM dense-cache gap does
not itself erase the authoritative compact prefix; the dense repair bridge is
separately marked unknown.

## Compatibility

Text-only cache files and payload ABI remain unchanged. Changed, moved or
missing images cannot select the checkpoint. Appended-image disk recovery is
excluded; live appended-image reuse remains #927. No model math, sampling,
launcher or default change.

## Practical impact

The benefit is reusing an exact image-conditioned checkpoint across a process restart while rejecting different conditioning. It avoids repeated vision prefill; it does not accelerate decode or provide #927's appended-image live reuse. The standalone Metal restart test below demonstrates restored tokens and cold rejection of a changed image. No marginal throughput or latency uplift over `f62ca29` is claimed.

## Combined runtime context

## Validation

One standalone commit, `5c697a48858f`, directly on upstream `f62ca29a3087` (2026-09-07). Tests cover retained/rebuilt/appended image provenance, interrupted sync, partial-image rejection, malformed metadata and text-only compatibility.

Before the Metal-only upstream update, the standalone patch passed fresh private-cache tests on M2 Max/Metal with DeepSeek-V4-Flash-Vision-Exp IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8 and its encoder, 65,536 context, 4,096 prefill and two resident sessions. After graceful restart, text replay loaded all 1,816 tokens; exact-image continuation loaded 1,132 tokens and evaluated ten new tokens. A changed image stayed cold. Supported text-tool continuation also passed. Inline tool-image messages require separate #927 and are not attributed to this PR. GLM provenance is source/unit tested, not GLM-model validated. #943 overlaps disk lookup and needs both regression groups retained.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -54417,6 +54417,8 @@ typedef struct {
     uint32_t token_start;
     uint32_t token_count;
     uint8_t fingerprint[32];
+    uint8_t state_fingerprint[32];
+    bool state_fingerprint_valid;
 } ds4_vision_identity;
 
 struct ds4_session {
@@ -64444,6 +64446,9 @@ static int ds4_prompt_append_deepseek4_vision(
     free(embedding->data);
     embedding->data = block;
     embedding->token_count = layout.token_count;
+    memset(embedding->state_fingerprint, 0,
+           sizeof(embedding->state_fingerprint));
+    embedding->state_fingerprint_valid = false;
     embedding->layout = 0;
     embedding->grid_width = 0;
     embedding->grid_height = 0;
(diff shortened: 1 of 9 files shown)


JordiPosthumus:
The recent upstream vision-continuation change (81a76585ff44e0494053fca9101130e2535648b9) overlaps this branch, so it needs rework against current main before it is ready to merge.

This PR still addresses a separate gap: persisting image-conditioned KV with exact conditioning provenance and safely restoring it across a process restart. Upstream's live image/embedding reuse does not provide that disk-cache behavior; current main still excludes image-conditioned state from the text-keyed disk cache.

A future rework should build on upstream's new live-vision implementation while preserving authoritative retained-image provenance, rejection of changed or unknown conditioning, text-cache compatibility and restart regression coverage. Leaving this open for now; rework is deferred, with no commitment to a follow-up.

PR #952, opened 2026-09-02
AProjQ4: imatrix-guided Q4_K attention and GPU runtime optimizations

GiorgioOppo:
Updated 2026-09-11. PR head: `7a5002de`.

The latest full-model CUDA/ROCm reports below tested `d0bc6f07`; the Metal Q4/Q8 report tested `6a20b131`. Subsequent runtime changes require validation on the final merge candidate.

## Summary

This PR supersedes #621 and adds support for the DeepSeek V4 Flash **AProjQ4** layout: imatrix-guided requantization of 215 dense attention-projection tensors from `Q8_0` to `Q4_K`.

The matching model is **2.14 GiB smaller**. Tests report faster generation on Metal, CUDA and ROCm, with results tied to the revisions and workloads below. The latest CUDA incremental-prefill retest shows that most of the earlier regression has been recovered. **ROCm retains a measurable large-chunk prefill cost.**

The conversion preserves the original `-0731` checkpoint, expert and output-head quantization, tokenizer, GGUF metadata, and all unselected tensors. The selected families are `attn_q_a`, `attn_q_b`, `attn_kv`, `attn_output_a` and `attn_output_b`, across 43 layers.

| Layout | GGUF size |
|---|---:|
| AProjQ8 | 80.76 GiB |
| AProjQ4 | 78.62 GiB |
| Saving | **2.14 GiB / 2.65%** |

Beyond layout support and conversion, the current diff includes CPU/GPU execution changes, F16 compressor and aligned-Q8 paths, MoE and SSD-cache changes, tests and benchmark tooling. Later commits add execution-phase handling, HC normalization reuse and Metal indexer/top-k optimizations. The ROCm Q4 lookahead and prepared CUDA/ROCm indexer candidates remain outside automatic inference dispatch. #621 remains the earlier development archive.

## CUDA implementation comparison — tested at `d0bc6f07`

On DGX Spark / GB10, @adamlawi tested `d0bc6f07` against a production build based on `ad8149df` + PR #704 + two local CUDA stubs. **Both used the same AProjQ4 GGUF; the baseline already contains the earlier #621 work.** This measures the difference between those builds, not Q4 versus Q8 or this PR versus upstream main.

The 4.5-hour server campaign used eight counterbalanced blocks, matching prompts and 400 generated tokens per request.

| Context | Matched pairs | Baseline decode | `d0bc6f07` decode | Decode change | Prefill change |
|---|---:|---:|---:|---:|---:|
[...]

Issue #929, opened 2026-09-01
Help wanted: test DeepSeek V4 Flash Vision-Exp AProjQ4 on Metal, CUDA, and ROCm


GiorgioOppo:
## Summary

We are looking for volunteers to test the new **DeepSeek V4 Flash Vision-Exp AProjQ4** GGUF on:

- Apple Silicon / Metal
- NVIDIA / CUDA
- AMD Strix Halo / ROCm
- Fully resident and SSD-streaming configurations

The runtime implementation is tracked in [PR #621](https://github.com/antirez/ds4/pull/621).

This request complements [#838](https://github.com/antirez/ds4/issues/838), which covers the non-Vision AProjQ4 model, and follows the original Vision support request in [#851](https://github.com/antirez/ds4/issues/851).

You only need to test one hardware/backend combination. Reports of crashes, incorrect image understanding, quality drift, unexpectedly low performance, or successful runs are all useful.

## Model artifact

The experimental model is available from [Hugging Face PR #22](https://huggingface.co/antirez/deepseek-v4-gguf/discussions/22):

[`DeepSeek-V4-Flash-Vision-Exp-IQ2XXS-w2Q2K-AProjQ4-SExpQ8-OutQ8-imatrix.gguf`](https://huggingface.co/antirez/deepseek-v4-gguf/blob/refs%2Fpr%2F22/DeepSeek-V4-Flash-Vision-Exp-IQ2XXS-w2Q2K-AProjQ4-SExpQ8-OutQ8-imatrix.gguf)

```text
Size:    84,420,584,608 bytes (78.62 GiB)
SHA-256: a1d0fde967c910e37e4bea00c4db1b2e8a88978770828a95772bc66837386cb8
```

The model was produced by requantizing 215 dense-attention projections—five projections across 43 layers—from `Q8_0` to `Q4_K`, using the 220k routed-and-dense DS4 imatrix.

All tensors outside the selected attention projections retain the source model’s quantization layout.

The matching Vision-Exp encoder sidecar is also required:

```text
DeepSeek-V4-Flash-Vision-Encoder.gguf
```

## Download

Install the current Hugging Face CLI if needed, then run:

```sh
mkdir -p gguf

hf download antirez/deepseek-v4-gguf \
  DeepSeek-V4-Flash-Vision-Exp-IQ2XXS-w2Q2K-AProjQ4-SExpQ8-OutQ8-imatrix.gguf \
  --revision refs/pr/22 \
  --local-dir gguf

[...]

PR #967, opened 2026-09-04
Fix CUDA build: restore g_tp_block_ctx and add missing TP/kv-norm CUDA stubs

Matthley:
## Summary
- `make cuda-spark` (and any non-Apple GPU build) currently fails to compile: `g_tp_block_ctx` was accidentally scoped to `#if !defined(DS4_NO_GPU) && defined(__APPLE__)` in "Remove dead Metal TP experiments" (5d41a80), even though it's referenced from shared, non-Metal-specific code (`metal_graph_verify_suffix_tops_impl`, `ds4_engine_tp_bind`). Restoring the guard to `#if !defined(DS4_NO_GPU)` fixes the compile.
- The same Metal TP gate-flag-fold work added calls to `ds4_gpu_add_tensor_tp_flag` and `ds4_gpu_dsv4_qkv_norm_defer_kv_next` from shared `ds4.c` code, but only ever defined them in `ds4_metal.m`, leaving the CUDA build with undefined references at link time. This adds CUDA-side definitions (plus the related `ds4_gpu_tp_flag_fold_request`/`ds4_gpu_kv_norm_task_begin_concurrent`/`ds4_gpu_kv_norm_task_end_concurrent`) that take the same plain fallback path the Metal implementation itself uses whenever its fold optimization doesn't apply — no behavior change, just makes CUDA link again.

Verified on a DGX Spark GB10 after rebasing onto current `main`: clean `make cuda-spark` builds all five binaries, `./ds4 --help` runs, and full CUDA inference completes all 43 layers with the supported AProjQ8 GGUF at PR head `940f694`.

## Test plan
- [x] `make cuda-spark` builds all five binaries without errors
- [x] `./ds4 --help` runs on DGX Spark GB10 (CUDA, sm_121a)
- [x] Full inference run with a downloaded GGUF on DGX Spark GB10


Matthley:
Rebased onto current `main` (`bd66c402`) and force-updated safely. New PR head: `9b112916`.

Validation on NVIDIA DGX Spark / GB10, Ubuntu 24.04.4, CUDA 13.0, driver 580.173.02:

- `make clean && make -j4 cuda-spark`: PASS
- `make CUDA_ARCH=sm_121 test-cuda-q8-scratch`: PASS
- `make CUDA_ARCH=sm_121 test-cuda-dspark-moe`: PASS (zero exact assignment failures)
- `make CUDA_ARCH=sm_121 test-mxfp4-cuda`: PASS
- `git diff --check`: PASS

I also applied this fix cleanly on top of PR #952 and completed resident CUDA Vision AProjQ4 QA on the same GB10. The multimodal run passed without crashes, NaNs, loops, or grounding/OCR errors; full report: https://github.com/antirez/ds4/issues/929#issuecomment-5647953222

The existing `cuda-regression` target on current upstream still misses `ds4_image.o` at link time (`ds4_deepseek4_attention_bounds` undefined). Manually adding that object makes `cuda long-context regression: OK`; this is independent of the two-file patch in this PR.

PR #971, opened 2026-09-04
server: opt-in auto-reduction of over-budget image histories (DS4_VISION_KEEP_IMAGES)

nazerim:
Follow-up split from #968 per @JordiPosthumus's scope review: automatically
discarding older images is an intentionally lossy API policy, so it lives in
its own PR and is explicitly opt-in.

Behavior:
- Unset (default): unchanged upstream policy. Requests over 16 images are
  rejected; the error message now names the opt-in.
- DS4_VISION_KEEP_IMAGES=N (1 <= N <= 1024) at server start: a request whose
  history carries more than N images is SERVED by keeping the last N (the
  freshest views matter most to agent loops) and replacing each dropped image
  sentinel in the rendered transcript with a fixed text note, so the model
  still sees an honest history, marker/span counts stay consistent for the
  vision sync, and the reduced text remains a deterministic cache key.
- A 1024 hard guard still fails absurd requests before any work; invalid env
  values are ignored with a one-time warning and the reject policy stays.

Why: screenshot-per-tool-round agent sessions accumulate images without
bound. The current alternative for such clients is a session that stops
working at the 17th screenshot with no graceful path; truncation with an
honest transcript note keeps them running and cache-friendly.

Sliding-window note: as the keep-window advances, the newly omitted image's
text changes, invalidating cached prefixes only back to the previous
retained image.

Verification (on this exact submitted code):
- Unit test on bare main: drop-oldest rewrite (oldest sentinel replaced,
  newer markers survive verbatim, drop-all keeps surrounding text, lost
  sentinel reported instead of misrendered). ds4-server tests: ok.
- Live with DS4_VISION_KEEP_IMAGES=16 (the default opt-in value): 20-image
  request served; log: "multimodal image budget: 20 images, kept last 16,
  omitted 4 oldest".
- Live with unset default: 2-image request served normally; 17-image request
  rejected with HTTP 400 and the hint message naming the opt-in.
- Live with invalid DS4_VISION_KEEP_IMAGES=abc: one-time warning logged,
  reject policy retained.

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,7 +332,11 @@ library is required.
 Anthropic requests. HTTP images must be inline: use a PNG/JPEG data URI for
 OpenAI or base64 image source for Anthropic. File paths and remote URLs are
 rejected. A request may contain up to 16 images and the HTTP body is limited to
-64 MiB.
+64 MiB. Requests over that limit fail by default; agent loops that accumulate a
+screenshot per tool round can instead set the environment variable
+`DS4_VISION_KEEP_IMAGES=N` (1 <= N <= 1024) to opt into auto-reduction: the
+server keeps the last N images of the history and serves the request, replacing
+each dropped image sentinel in the transcript with a fixed text note.
 
 Vision runs on Metal, single-GPU CUDA, and ROCm. On a DGX Spark, use the CUDA
 command above and add `--vision FILE`. On the 128 GB Strix Halo reference host,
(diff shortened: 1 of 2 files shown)


nazerim:
Closing for rebase, not abandoning: the opt-in DS4_VISION_KEEP_images budget remains upstream-absent after 81a7658. Superseded-tree conflicts + our fork just merged upstream through 6289c51, so I'll re-validate on the merged tree and resubmit clean if it's still additive. Live-validated at keep=16 (20-image request -> kept 16/omitted 4 with drop-oldest note), invalid-env rejection, and default reject>16 with hint.

nazerim:
Superseded by #1022 - rebased onto current main (`fa0af86`), code hunk auto-merged against 81a7658, docs moved to docs/SERVER.md.

PR #976, opened 2026-09-04
Add WebP image decoding

Bendr0id:
Screenshot tooling for mobile and browser debugging emits WebP by default, and clients transcode to it on their own initiative, but an image sent that way was refused by media type before anything looked at it. The request failed as a whole rather than the image simply being read, which is the worst outcome: the caller attached a screenshot and got an error about the format instead of an answer.

Decoding is a new dependency-free single header beside the vendored JPEG and PNG ones, covering both compression modes the format defines. A still lossy image is a single VP8 key frame, so the lossy path is intra only and needs no motion compensation or reference frames: boolean entropy decoder, frame, segment, filter and quantizer headers, 16x16, 4x4 and chroma intra prediction, token decoding, dequantization, the inverse DCT and the Walsh-Hadamard pass, both loop filters, and the conversion to RGB. The lossless path is a separate algorithm: prefix codes, backward references, the color cache and the four inverse transforms, including palettes packed several pixels to the byte. RIFF and VP8X parsing skips ICCP, EXIF and XMP, which is how real payloads arrive. Animation is refused with a message saying so rather than silently showing the first frame, since a caller who sent an animation did not ask for one arbitrary frame of it. The alpha chunk of a lossy image is ignored, so those decode to opaque RGB.

All three front ends reach that decoder through one path, so `/read` in the CLI and `view_image` in the agent read WebP without a change of their own, and neither gains a way to reject a format the decoder handles. Only the server needed its gate widened, because it alone filters on a media type the client declares rather than on the bytes themselves, and it accepts `image/webp` for both the OpenAI data URI form and the Anthropic form, where the media type travels beside the data. The `/read` help text offered PNG and JPEG only and now names what it actually takes.

The probability, quantizer and mode tables are values fixed by RFC 6386 and were taken from the reference implementation rather than retyped, and `third_party/webp/LICENSE` records that.

Output is bit exact with libwebp's own `dwebp` across 244 images spanning tiny and odd dimensions, flat and noisy content, every quality setting, alpha, and each lossless transform. That mattered: it is what caught the Walsh-Hadamard pass needing to scatter DCs across the whole coefficient array, and the inner edge filter flag depending on whether a macroblock decoded to all zero coefficients rather than on its skip bit. Because this parses untrusted network input, 14640 truncated and corrupted inputs were also run under the address and undefined behaviour sanitizers, which is what found a distance code tree built over 256 symbols instead of the 40 its alphabet actually has.

Tests are in `tests/test_deepseek4_vision_image.c` for the decoder and in the server unit tests for acceptance over both request forms. The build is warning free under `-Wall -Wextra`.


Bendr0id:
Hi @antirez, can we get this in? 
Lot of harnesses are using webp for screenshots.  I've verified the branch and it's still working.

hertz-hwang:
Second independent report, from a different client than the ones described — and it looks nothing like a format problem from the outside, so I want to record the trigger before this gets closed as "works for me".

**Origin.** A coding-agent harness normalizes every pasted image before it reaches the provider: it passes the source through untouched only when it is already clean 8-bit sRGB and inside its budgets (2048×2048 px, 4 MiB), and otherwise re-encodes it, choosing the container by alpha presence. A full-screen macOS capture is 3456×2234 = 7.7 Mpx, i.e. **always** over the pixel budget, so it always gets re-encoded. On this machine the two stored results of that path are `2546×1646` WebP objects — exactly 3456×0.737, matching a 2048² downscale.

**Why the failure reads as a size limit.** A window or region crop of the same screen is under the budget, passes through as PNG, and works. The full-screen capture becomes WebP and is refused before any decoder is consulted, with `400 invalid JSON request`. Measured against `main` (8db1d1d):

| payload sent to `/v1/responses` | result on `main` |
| --- | --- |
| 3456×2234 PNG, 3.55 MB request body | accepted |
| 3456×2234 JPEG | accepted |
| 32 px PNG | accepted |
| `data:image/webp;base64,…` | 400 `invalid JSON request` |
| the same pixels re-encoded as PNG/JPEG | accepted |

So neither size nor base64 length was ever the axis — only the container. That matches your framing in the PR description and is a nice confirmation that refusing by media type is the whole bug.

**Verified on this branch** (`f40aedb`, Apple M3 Max, macOS, Metal). `make tests/test_deepseek4_vision_image` builds warning-free with `-Wall -Wextra` and passes. Decoding your `ds4_image.o` against the two real stored WebP objects from above:

```
2546x1646  has_alpha=0  -> decoded OK
1256x80    has_alpha=1  -> decoded OK
1440x900 PNG            -> decoded OK (regression)
animated                -> "animated WebP is not supported; send a still image"
```

The animation refusal matches what I reached independently: picking one frame silently is worse than refusing.

**One thing worth pinning down in a test.** I had previously tried a libwebp-backed variant (your vendored single header is the better fit for this tree — no new runtime dependency, and it matches the `third_party/iris` idiom), so I had a reference decoder to diff against. On **lossless** WebP, including an alpha-bearing one, your decoded raster is byte-identical to libwebp's and to the PNG of the same source. On **lossy** VP8 it is not:

```
2546x1646 lossy webp: 402231 / 4190716 pixels differ (~9.6%)
  all of them opaque (a=255); max |Δ| = 2 per channel
  first: libwebp (37,123,180) vs vendored (35,124,180)
```

That is a legitimate reconstruction difference (YUV→RGB rounding / chroma upsampling is not normatively fixed), not a defect — but `ds4_image.fingerprint` hashes the decoded RGB, and the fingerprint keys the vision-embedding cache and whatever disk reuse sits on top of it. Two consequences: a cache written by one decoder build is not addressable by the other, and any test that asserts exact WebP pixels against a libwebp- or Pillow-generated reference will fail for reasons unrelated to correctness. Suggest a one-line comment or a tolerance-based assertion so nobody later "fixes" this as a regression. Happy to send that as a small patch if useful.

*(Scope note: I ran the image test target and a decode probe against `ds4_image.o`. I did not exercise the server endpoints or the `ds4_agent.c` / `ds4_cli.c` / `ds4_help.c` doc changes on this branch.)*

PR #983, opened 2026-09-05
kvstore: stamp checkpoints with a behavioral tokenizer fingerprint

nazerim:
Persisted KV checkpoints store a token history that future requests
re-tokenize from text with whatever tokenizer the running engine has.
When tokenizer code or data changes (the JoyAI pre-tokenizer has seen
several rule fixes recently), old checkpoint token histories become
unreproducible: every resumed session diverges from its restored state
at the first changed merge boundary, keeps reloading the same stale
file through the text-keyed tiers, and re-imports the stale
tokenization - a permanent per-turn miss/reload/re-prefill loop.

Production symptom (this week, long multimodal agent session): 29
consecutive turns each live-missed at the same token, each reloaded
the same stale checkpoint by text key, each re-prefilled ~38k tokens.
The session was permanently pinned; only a manual cold rebuild healed
it. A plain text-equality check at load does not work as a detector:
it false-positives on legitimate preserved-thinking continuations,
whose stored sampled tokens intentionally differ from a whole-text
re-tokenization of their bytes (that difference is the bridge's entire
premise). The correct discriminator is engine identity.

This commit adds a behavioral tokenizer fingerprint: an append-only
probe set (apostrophe-punctuation seams, digit runs, indentation,
newline joins, UTF-8 letters, special-token edges) run through the
real tokenizer, hashed together with the token table. It covers
tokenizer data AND code changes with no human-maintained version
number, and it is cheap (computed once, lazily, under a mutex).

Mechanics:
- Every checkpoint trailer now begins with a TOKFP section (8-byte
  header + u64 fp) and the auxiliary ext bit DS4_KVSTORE_EXT_TOKFP.
- The loader pre-checks the stamp with one 16-byte seek at the trailer
  offset BEFORE touching the payload: mismatches cost microseconds,
  never a 600 ms load, and the session is left untouched.
- Rejection falls into the normal prefill path, which rewrites a fresh
  stamped file under the same text key: the poison class becomes
  self-healing instead of permanent.
- Unstamped legacy files keep today's trusted behavior exactly.

Verification (live agent session on Apple M5 Max):
- stamped restore: full 451k-token thinking-visible checkpoint accepted,
  38-token re-prefill, session continues at full reuse.
- synthetic skew (temporary probe-string change = simulated upstream
  tokenizer edit): old stamped files rejected instantly with a clear
  warning; fallback chain restored from the next valid candidate;
  fresh stores re-stamped correctly after restoring the probe.
- unit suite green on bare main (trailer size/write round-trip test
  updated for the new leading section).

Independent of #968/#971 (touches ds4.c/ds4.h/ds4_kvstore.*/server
trailer hooks only); no behavior change for files written today.

### DIFF
--- a/ds4.c
+++ b/ds4.c
@@ -56033,8 +56033,31 @@ uint64_t ds4_engine_tokenizer_fingerprint(ds4_engine *e) {
     pthread_mutex_lock(&e->tokenizer_fp_mu);
     if (!e->tokenizer_fp_ready) {
         uint64_t h = 14695981039346656037ull;
-        h = tokenizer_fp_hash_bytes(h, e->vocab.token,
-                                    (size_t)e->vocab.n_vocab * sizeof(ds4_str));
+        /* Hash tokenizer CONTENT, never in-memory struct bytes: pointer
+         * fields change with every process mapping, which would make the
+         * fingerprint restart-unstable and reject valid checkpoints after
+         * an ordinary server restart. */
+        h = tokenizer_fp_hash_bytes(h, &e->vocab.n_vocab, sizeof(e->vocab.n_vocab));
+        for (int ti = 0; ti < e->vocab.n_vocab; ti++) {
+            h = tokenizer_fp_hash_bytes(h, &e->vocab.token[ti].len,
+                                        sizeof(e->vocab.token[ti].len));
+            h = tokenizer_fp_hash_bytes(h, e->vocab.token[ti].ptr,
+                                        (size_t)e->vocab.token[ti].len);
+        }
+        /* Merge ranks: the open-addressed slot order is a deterministic
+         * function of the GGUF merge list, so hashing slots in index order
+         * (key bytes + rank) is stable across processes. */
+        h = tokenizer_fp_hash_bytes(h, &e->vocab.merge_rank.used,
+                                    sizeof(e->vocab.merge_rank.used));
+        h = tokenizer_fp_hash_bytes(h, &e->vocab.merge_rank.cap,
+                                    sizeof(e->vocab.merge_rank.cap));
+        for (uint64_t mi = 0; mi < e->vocab.merge_rank.cap; mi++) {
+            str_i32_entry *en = &e->vocab.merge_rank.entry[mi];
+            if (!en->used) continue;
+            h = tokenizer_fp_hash_bytes(h, &en->key.len, sizeof(en->key.len));
+            h = tokenizer_fp_hash_bytes(h, en->key.ptr, (size_t)en->key.len);
+            h = tokenizer_fp_hash_bytes(h, &en->value, sizeof(en->value));
+        }
         for (size_t pi = 0; pi < sizeof(DS4_TOKENIZER_FP_PROBES) /
                                    sizeof(DS4_TOKENIZER_FP_PROBES[0]); pi++) {
             ds4_tokens t = {0};
--- a/ds4.c
+++ b/ds4.c
@@ -38652,6 +38652,9 @@ typedef enum {
 } ds4_vision_kind;
 
 struct ds4_engine {
+    uint64_t tokenizer_fp;
+    int tokenizer_fp_ready;
+    pthread_mutex_t tokenizer_fp_mu;
     ds4_model model;
     ds4_model mtp_model;
     ds4_model vision_model;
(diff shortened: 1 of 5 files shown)


JordiPosthumus:
Hey Naz — I like the aim here. Preventing a session from repeatedly restoring an incompatible checkpoint is useful hardening. I spotted a restart-stability issue in the fingerprint on `f043133b` that is worth fixing before this lands.

In `ds4_engine_tokenizer_fingerprint()`:

```c
h = tokenizer_fp_hash_bytes(h, e->vocab.token,
                            (size_t)e->vocab.n_vocab * sizeof(ds4_str));
```

`ds4_str` contains `const char *ptr` and `uint64_t len`, so this hashes the **pointer addresses and lengths**, not the token-string bytes. The same unchanged model can map at a different address in another process/restart. That changes the fingerprint even though tokenization is identical, causing valid stamped checkpoints from the previous process to be rejected. The behavioral probes cannot cancel out that address-dependent input.

I think the fix is to hash a canonical serialization of the tokenizer data: token IDs/order, lengths and the actual bytes behind each pointer, plus the other data needed to identify its tokenization behavior. Avoid hashing raw structs/process representations.

Two particularly useful regressions would be:

- Identical tokenizer contents at different mapping addresses / in separate processes produce the same fingerprint, and a saved checkpoint survives an unchanged-server restart.
- Changing token bytes without changing their lengths changes the data fingerprint, independently of whether those tokens happen to appear in the probe set.

This is a static-code finding, not a claim that I've run your full model-backed test sequence. Thanks for working on this — the underlying problem is definitely worth addressing, and this looks like a fixable snag. 🙂

nazerim:
You're right, and thank you — this is the kind of bug my live validation was uniquely suited to hide. Same binary + same args + same load sequence on macOS reproduced the allocator layout across my restart test, so the pointer-hash passed by luck; on any different mapping (or a different host) every stamped file would silently reject after each restart. Static review caught what the runtime fluked.

Fix (will force-push once validated): hash tokenizer *content* — `n_vocab`, then each token's length + bytes in id order, then the merge-rank table (used/cap + every used slot's key bytes + rank in slot order, which is a deterministic function of the GGUF merge list, not of process state). The behavioral probes stay on top, covering code changes. No raw struct hashing anywhere.

Your two regressions become the validation plan, run live:
1. **Restart-stability:** store checkpoints, restart the server twice across unchanged binaries, confirm zero `fingerprint mismatch` lines and full-depth restores (this is exactly the case that silently fails with the pointer hash — and my earlier "verified" claim was lucky, not sound).
2. **Content sensitivity:** synthetic token-bytes mutation (same lengths, no probe overlap) → new fingerprint → old stamps rejected. (I can pair that with the existing synthetic probe-change test for the code dimension.)

I'll report both results on this PR before re-requesting review. The lesson is noted for the probe-set doc comment as well: pointer-derived inputs must never reach the hash.

nazerim:
Fixed and validated — head now `e692449` (two commits: original + the content-hashing fix you outlined).

What changed in `ds4_engine_tokenizer_fingerprint()`:
- `n_vocab`, then every token's **length + bytes** in id order (no raw `ds4_str` array hashing anywhere).
- The merge-rank table as content: used/cap plus each used slot's key bytes + rank. Open-addressed slot order is a deterministic function of the GGUF merge list, so it is stable across processes without hashing pointers.
- Behavioral probes ride on top, unchanged, covering code changes.

Your two regressions, run live on Apple M5 Max against the production model file:

1. **Restart stability.** Two separate process launches of the unchanged model produce the identical fingerprint (`e467e62ceeadf47e`), and a stamped checkpoint stored by the first process restored at full depth from the second (`kv cache hit text tokens=32768 ... load=69.4 ms`, re-prefill = only the 5253 tokens above the grid, **zero** mismatch lines). With the pointer hash this cycle silently rejected or accepted depending on allocator luck; it is now deterministic either way.

2. **Content sensitivity.** APFS-cloned the model, flipped one byte of a token string (`TABLE` -> `TASLE`, same length, no probe string contains `TABLE`): fingerprint moves to `5f83c7a25a3b184e`. Data changes are captured independently of probe coverage, as requested.

One confession your static finding flushed out: my original "validated across restart" claim in the PR description was real but *lucky* — same binary + same args reproduced the allocator layout on macOS, so the pointer-hash passed its own test. The fix and both regressions above are now the honest basis for that claim. Thanks for catching this before it landed.

nazerim:
Closing for rebase: the tokenizer fingerprint is still upstream-absent and our fork just merged upstream through 6289c51 (conflicts in ds4.c/ds4.h from 81a7658). Will resubmit rebased after soak-testing on the merged tree. Jordi's pointer-hash catch is fixed on the branch (e692449) with both regressions documented above; thanks again for the sharp review.

nazerim:
Superseded by #1020: rebased onto current main as `07e0478`+`aa0b462` and GitHub refuses reopen after the rebase force-push. Content identical in substance; production soak evidence updated there.

Issue #816, opened 2026-08-16
Stateless chat clients never extend the live KV session on Flash/Metal: measurements, the anchor that saves them, and the replay recipe that fixes it


adriangalilea:
This is an analysis with numbers, not a bug report. It may be worth a paragraph in the README's client recipes; happy to send that as a PR if wanted.

**Setup.** Mac Studio M3 Ultra 512 GB, Metal, DeepSeek V4 Flash MXFP4 0731, `84cc882`, `--batched-session 4 --ctx 300000 --kv-disk-dir ... --kv-disk-space-mb 16384`. Client: a Telegram group bot on OpenAI chat/completions, `think:false`. Each request is the room's transcript (system + memories + N user turns), ~7-9k tokens, plus whatever arrived since the last call. Verdicts run every few seconds when the room is active.

**Observation 1: the live session was never reused.** 787 out of 787 requests logged `live kv cache miss ... reason=token-mismatch`, e.g. `live=6752 prompt=6776 common=6676`. After a call the slot holds `prompt + sampled output`; the next prompt is `prompt + new user turns`, which shares `common` = the old prompt and then diverges, and `generate_job` takes the live session only on `common == old_pos && prompt.len >= old_pos`. The rewind path (`live_prefix_rewind_target`) is GLM-only and covers exact-prefix retries (#698/#710), not this. So every chat request fell through to disk.

**Observation 2: on disk, only the cold anchor can help such a client**, because evict dumps end in sampled output too. The cold anchor cuts at `kv_cache_chat_anchor_pos` (last user turn before the first assistant turn), so in a room where the bot spoke early it sits at ~5k of ~7k tokens: prefill from the anchor is ~1.7-3k tokens (~5-8 s), and once the disk was full the anchor was evicted on every store (#814).

**Observation 3: replaying the assistant turn byte-for-byte fixes it completely**, on both API shapes, without engine changes:

- tool call: replay `{"role":"assistant","tool_calls":[{id, function:{name, arguments}}]}` + a `tool` result; ds4 re-renders the exact sampled DSML from the tool-id map → `cached_tokens` = prior prompt + generation, e.g. `ctx=8228..8277:49`, only the new turn prefilled.
- plain text: replay `{"role":"assistant","content": <exact text as returned>}` → same, `ctx=7527..7558:31`.

[...]

Issue #781, opened 2026-08-11
When using Codex, once I get RESPPROTO TOOLS finish=stop the prompt has to be processed again, cache rebuilt.


brutuscat:
Hi @antirez thank you for this great project!

I finally was able to make it run and I am very happy with it. I'm using it with Codex.

But I found that from time to time the responses get to a `finish=stop` response "type" (I guess? that is what I see in the output logs), and when that happens in the middle of a session the cache has to be rebuilt. It is a bit annoying, but I also do not know if I am doing something wrong.

Here you can see (in the middle) what I am talking about:
```
0811 22:01:11 ds4-server: tool calls ctx=98994..99434:440 RESPPROTO n=1 raw_tool_text=1 ids=[call_d8f9c801309531b0052ba13aa441e7d0] names=[exec_command]
0811 22:01:11 ds4-server: chat ctx=98994..99434:440 gen=1094 RESPPROTO TOOLS DSML_START DSML_END finish=tool_calls 54.478s
0811 22:01:12 ds4-server: responses live continuation RESPPROTO match=visible-prefix ids=1 cached=100528 prompt=101055
0811 22:01:12 ds4-server: chat ctx=100528..101055:527 RESPPROTO TOOLS prompt start
0811 22:01:12 ds4-server: chat ctx=100528..101055:527 RESPPROTO TOOLS prefill chunk 0/527 (0.0%) chunk=0.00 t/s avg=0.00 t/s 0.000s
0811 22:01:16 ds4-server: chat ctx=100528..101055:527 RESPPROTO TOOLS prefill chunk 527/527 (100.0%) chunk=123.85 t/s avg=123.84 t/s 4.255s
0811 22:01:16 ds4-server: chat ctx=100528..101055:527 RESPPROTO TOOLS prompt done 4.256s
0811 22:01:18 ds4-server: chat ctx=101055..101105:50 gen=50 RESPPROTO TOOLS THINKING decoding chunk=21.53 t/s avg=21.53 t/s 2.322s
0811 22:01:21 ds4-server: chat ctx=101105..101155:50 gen=100 RESPPROTO TOOLS THINKING decoding chunk=21.57 t/s avg=21.55 t/s 4.640s
0811 22:01:23 ds4-server: chat ctx=101155..101205:50 gen=150 RESPPROTO TOOLS THINKING decoding chunk=21.51 t/s avg=21.54 t/s 6.964s
[...]

PR #985, opened 2026-09-05
Preserve live KV cache across normalized tool replay

rratmansky:
# Preserve live KV cache across normalized tool replay

## Summary

Preserve the resident KV checkpoint when a Chat Completions client replays a successful tool-call turn with semantically equivalent but byte-normalized assistant text.

The server already uses a remembered visible transcript to continue from richer live state after tool-less thinking. This extends that existing mechanism to successful tool-call turns:

- build the syntax-aware client-visible assistant transcript after a tool call;
- remember it alongside the sampled live token frontier;
- validate that the next rendered prompt exactly extends that visible transcript;
- if it does, retain the sampled live tokens and tokenize only the new text suffix.

The visible continuation check now runs before GLM token rewind. Requests that do not match still use the existing token, rendered-text, disk-cache, or cold-prefill paths.

## Motivation

A long GLM-5.3 agent session repeatedly lost its entire resident cache because parse-and-replay normalized one trailing newline immediately before a tool call:

```text
live_tokens_before: 161095
prompt_tokens: 161124
live_prompt_common: 161018
memory_miss_reason: token-mismatch
cache_source: none
cached_tokens: 0

first mismatch:
  live:   "):\n"
  prompt: "):"
```

Although the visible conversation was the same and more than 99.9% of the token history agreed, the request prefetched all 161,124 tokens from zero. Cache misses at this context took roughly 420–525 seconds; a neighboring successful resident-cache hit completed in about 14 seconds.

This is particularly painful for tool-using agents because every tool result immediately triggers another request.

## Implementation

- Generalize the existing visible-chat continuation description to cover normalized tool-call replay as well as omitted tool-less reasoning.
- Add a small pure visible-prefix predicate that requires:
  - a valid remembered state;
  - the same live token frontier;
  - an exact byte prefix of the newly rendered prompt;
  - at least one byte of new prompt suffix.
- After a successful non-Responses tool call, remember the syntax-aware visible transcript produced from the request prefix and parsed assistant/tool-call output.
- Try this validated visible continuation before GLM rewind.
- Report successful use as `cache_source=chat-visible` and log `visible chat continuation match=visible-prefix`.

Responses API behavior is unchanged; it retains its existing protocol-specific continuation mechanism.

## Safety

This does not perform an arbitrary token rewind and does not modify GLM KDA/indexer state.

The sampled live KV state is reused only when the next request exactly extends the remembered client-visible transcript at the same live frontier. Edited histories, unrelated branches, stale slots, non-extending prompts, failed/length-truncated tool turns, and image-identity mismatches continue through the existing fallback paths.

The implementation uses the existing syntax-aware tool renderer, so it covers both DeepSeek DSML and GLM `<tool_call>` replay.

## Related work

This is a narrowly scoped, current-`main` implementation of the visible-checkpoint approach discussed in several existing reports and PRs:

- #789 contains the closest implementation and is a broader superset, including tool-context final-answer checkpoints, streaming-error preservation, and KV waypoint changes. This PR isolates only the successful tool-call frontier fix and adds GLM-5.3 production validation.
- #709 handles the complementary DeepSeek case where clients omit `reasoning_content`; this PR was validated with a client that replays parsed reasoning.
- #611 / #609 bind OpenAI tool-result continuation through `tool_call_id`, an alternative protocol-specific solution to the same repeated-prefill symptom.
- #904 preserves GLM tool-call separators at rendering time and is complementary: it prevents one source of replay drift, while this PR safely retains live KV when semantically equivalent visible replay still differs from sampled tokens.

Related broader reports include #691, #816, #781, and #364.

## Tests

Machine and test configuration:

- Apple M5 Max
- Metal backend
- `ds4flash.gguf` -> `GLM-5.3-Flash-Q2.gguf`

Commands run:

```sh
make clean
make -j4
make -j4 ds4_test
./ds4_test --server
./ds4_test --tool-call-quality
```

Results:

```text
server: OK
tool-call-quality: OK
ds4 tests: ok
```

The server tests include:

- canonical visible transcript equality for DeepSeek tool syntax;
- canonical visible transcript equality for GLM tool syntax;
- acceptance of the observed normalized `"):\n"` -> `"):"` tool-turn replay;
- rejection when the live frontier differs;
- rejection when visible history changes;
- rejection when the request does not extend the remembered transcript.

A complete `./ds4_test --all` run was also attempted with the local GLM-5.3 Q2 model. Its server, tool-call quality, long-context, GLM continued-prefill, and Metal kernel sections passed. It reported 20 model/golden-vector discrepancies in logprob vectors, SSD-streaming vectors, a local golden vector, and long-prompt Metal tensor equivalence. Those checks compare the locally configured GLM model against existing expected vectors and are outside this server-only change; they are reported here as notable failures per `CONTRIBUTING.md`.

### DIFF
--- a/ds4_server.c
+++ b/ds4_server.c
@@ -10487,19 +10487,37 @@ static int responses_live_visible_prefix_prompt(server *s, server_slot *slot,
     return live_tokens->len;
 }
 
-/* Tool-less thinking continuation.
+static bool visible_live_matches_prompt(const visible_live_state *state,
+                                        int live_pos,
+                                        const char *prompt_text,
+                                        size_t prompt_len,
+                                        size_t *visible_len) {
+    if (!state || !state->valid || state->live_tokens != live_pos ||
+        !prompt_text || !state->visible_text || state->visible_len >= prompt_len ||
+        !byte_prefix_match(prompt_text, prompt_len,
+                           state->visible_text, state->visible_len))
+    {
+        return false;
+    }
+    if (visible_len) *visible_len = state->visible_len;
+    return true;
+}
+
+/* Visible chat continuation.
  *
  * Chat/completions and Anthropic do not have a previous_response_id object that
- * binds a later request to the last sampled turn.  Still, after a normal
- * tool-less thinking answer, the next prompt renderer intentionally omits that
- * hidden reasoning.  The live KV state is richer than the visible transcript.
+ * binds a later request to the last sampled turn.  A later rendered prompt may
+ * nevertheless normalize or omit sampled assistant bytes: hidden reasoning is
+ * omitted after tool-less answers, while reasoning immediately before a tool
+ * call may lose insignificant whitespace during parse-and-replay.  The live KV
+ * state is then richer than, but logically equivalent to, the visible replay.
  *
- * Remembering the visible transcript as a key lets us keep the sampled hidden
- * KV when the next request clearly extends that same visible history.  This is
- * the same byte-prefix idea used by the disk cache: the client-visible text
- * selects the checkpoint, while the payload stays the exact sampled token
- * frontier.  If the visible key does not match, callers fall back to ordinary
- * token/text/disk matching. */
+ * Remembering the visible transcript as a key lets us keep the sampled live KV
+ * when the next request clearly extends that same visible history.  This is the
+ * same byte-prefix idea used by the disk cache: the client-visible text selects
+ * the checkpoint, while the payload stays the exact sampled token frontier. If
+ * the visible key does not match, callers fall back to ordinary token/text/disk
+ * matching. */
 static int thinking_live_visible_prefix_prompt(server *s, server_slot *slot,
                                                const request *req,
                                                int live_pos,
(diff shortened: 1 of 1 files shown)


rratmansky:
Manual end-to-end validation on Apple M5 Max / Metal with `GLM-5.3-Flash-Q2.gguf` and a ~162K–170K-token tool-using agent session:

- 18 consecutive tool-result continuations selected `cache_source: chat-visible`.
- Reused live frontiers ranged from 162,940 to 170,109 tokens.
- Newly prefetched suffixes ranged from 23 to 714 tokens.
- Prompt-processing time for those continuations ranged from 0.304s to 2.803s.
- The initial cold prompt (161,995 tokens) took 446.468s to prefill.
- A later cold miss (166,208 tokens) took 468.552s to prefill.

Representative trace entry:

```text
prompt_tokens: 162980
effective_prompt_tokens: 162981
cached_tokens: 162940
live_prompt_common: 162385
cache_source: chat-visible
```

The stale ordinary token common-prefix position while the full live frontier was reused confirms that the new visible-transcript path—not an incidental exact-token hit—provided the speedup.

One boundary remains: after a tool-enabled turn finished with a normal final answer rather than a tool call, a later user turn cold-missed. This patch records successful tool-call frontiers, so it fixes the repeated tool-call/tool-result failure mode but does not claim to cover every non-tool final-answer transition.

Issue #867, opened 2026-08-26
Support for QWEN3.8-Flash-Next


doublerobust:
125B params plus 51B n-gram embeddings, only 6B active per token. Looks like an ideal model for local deployment and a serious contender to deepseek-v4-flash. 

https://www.modelscope.ai/collections/Qwen/Qwen38-Flash-Next

FramedStone:
yes please

dreamlate-ltd:
+1

maceip:
ive got qwen4exp running with its native multi-token prediction on Apple Silicon (MLX-side, under MTPLX). The interesting part is draft rejection: the DeltaNet layers carry a running recurrent state rather than a KV cache, so a rejected draft can't be truncated away: it's snapshot-and-restore. With MTP on, output is token-for-token identical to plain decoding testing

mentioning it here because i'd like to take a run at this model in ds4 next. The MLX implementation gives us a known good reference to diff against; not something to translate. From reading the ds4 impl, net new code needed is prob: the DeltaNet chunked-scan and per-step kernels, the sparse-attention block indexer, the n-gram embedding lookup, and teaching the --mtp-draft path to snapshot/restore recurrent state where it currently rewinds KV. this would be hand written against existing conventions, no new dependencies. If there's prior work in flight i should defer to, or opinions on where this should live, happy to coord before opening anything.

who am i and why do i care? just an unemployed replaced-by ai eng that likes ds4-agent :)

francosax:
FWIW there's already active work in progress: @Baekpica has a fork (github.com/Baekpica/ds4, branch `feature/qwen38-ssd-ple-handoff`, updated Aug 27) with a working proof-of-concept targeting CUDA/DGX Spark (GB10) — SSD-offload for the ~51B N-gram embedding table, 262K context confirmed working. CUDA kernels and the PLE runtime path are still being implemented/optimized per their own notes. Also discussed on the NVIDIA DGX Spark dev forum: forums.developer.nvidia.com/t/qwen3-8-flash-next-ds4-support/381417. Matching GGUF: huggingface.co/Baekpica/Qwen3.8-Flash-Next-Mixed-Quant-SSD-PLE-GGUF. Might be worth linking up rather than duplicating effort.

daaain:
llama.cpp support just got merged too: https://github.com/ggml-org/llama.cpp/pull/27742

Baekpica:
@francosax Thanks for the mention — really appreciate it.

[...]

PR #990, opened 2026-09-06
Add an experimental Qwen 3.8 Flash Next port on Metal

andrea82:
## Summary

An experimental Qwen 3.8 Flash Next port on the Metal backend, addressing #867.

- new graph: Gated DeltaNet recurrent layers alternating with QSA
  sparse-attention layers, hyper-connections in place of the plain residual,
  and per-layer n-gram embeddings (PLE) read from a second GGUF sidecar
- Metal kernels for all four mechanisms, each with a regression test against a
  scalar host reference written in the test itself
- the QSA block indexer: raw indexer keys cached per token, pooled rows per
  4-token block, and the model's top-512 block selection per query, wired
  through the GLM 5.3 indexer entry points for scoring and top-k; without it
  the model collapses past 2048 tokens
- the PLE sidecar served through explicit F_NOCACHE reads and a small row
  cache instead of being faulted through its mapping
- converter from the official FP8 snapshot, producing the model and the PLE
  sidecar; MXFP4 gains an encoder because the 640-wide expert rows are not
  QK_K aligned
- CLI and server support: ChatML rendering, reasoning effort, tool calls,
  model aliases
- CUDA and ROCm get link stubs and a load-time refusal, so they build and fail
  with a message instead of a wrong answer
- one commit that is not Qwen-specific: the Apple memory guard now trusts
  Metal's recommended working set when no wired limit was granted

This is unsolicited, so treat it as a proposal: happy to split it, cut parts of
it, or close it.

## Relation to work already in flight

#867 has other efforts on this model, and this one does not subsume any of
them.

@Baekpica has the most complete implementation: CUDA and DGX Spark, the PLE
table offloaded to SSD with latency reporting, 262K context, vision, and
two-bank batching throughout, with the PLE as its own `ds4_ple.c` module and a
CUDA counterpart. It lives on a large fork (896 commits ahead of this repo's
main and 409 behind) and has not been proposed here as a pull request.

@maceip is working on Metal with MTP and a 1M yarn context, currently in MTPLX
rather than in this repo. llama.cpp merged its own support in
ggml-org/llama.cpp#27742, which is a useful independent reference.

This port is deliberately the narrow one: Metal only, resident on a 64 GB
machine, the PLE read on demand from the sidecar rather than sharded across
files, no vision, no MTP. Its argument is scope and fit with the existing
conventions, not coverage. If you would rather have @Baekpica's
direction, this should give way to it; the parts that are not model-specific
(the MXFP4 encoder, the CUDA/ROCm stubs, the memory guard) stand on their own
either way.

#844 is a different model, Qwen 3.8 27B hybrid, but touches some of the same
Metal and tokenizer surface, so the two will need rebasing against each other.

Two points of independent agreement are worth noting as correctness evidence.
llama.cpp's PLE is "host-side row indices in a set_input, then get_rows", the
same design as here, and it found the same GDN detail this port spent time on
(a sigmoid output gate where Qwen 3.5 had SiLU). Where this port stores the PLE
table in Q8_0 (51 GiB), both llama.cpp and @Baekpica keep it in F16 (about
98 GiB); the quantized table is what lets model and sidecar coexist on a 64 GB
machine. The indexer follows the same pooled selection llama.cpp implements
and measures at a 0.975 mean Jaccard against the reference selection; the
equivalent measurement here is the unit test's host model plus the
bit-identical forced-sparse run below the budget.

## Validation

Machine: MacBook Pro M5 Pro, 20-core GPU, 64 GB, macOS 26.6.2. Backend: Metal.
Model: locally converted Qwen 3.8 Flash Next Q2 (IQ2_XXS gate/up, MXFP4 down)
plus the Q8_0 PLE sidecar.

Clean build, 0 warnings and 0 errors:

```text
make clean && make -j8
make cpu
```

Passing:

```text
./ds4_test --server         # OK, runs the new Qwen server tests
./ds4_test --metal-kernels  # OK
make test-qwen38-gdn        # PASS
make test-qwen38-qsa        # PASS
make test-qwen38-hc         # PASS
make test-qwen38-ple-hash   # PASS
make test-qwen38-ple-store  # ok
make test-qwen38-indexer    # PASS
```

The new suites need Metal but no model file, so they run anywhere with a GPU.
The indexer suite checks the pooled cache, the score matrix (including -inf for
blocks not yet complete for a row), the selection sets and the attention walk
over the selection against a scalar host reference, across a prefill chunk
boundary that splits a block, and that under a budget covering the whole
sequence every row selects its entire prefix.

Indexer, end to end. Below the 2048-token budget the engine uses the dense
walk (every block would be selected); `DS4_QWEN38_FORCE_INDEXER=1` forces the
sparse path there and is the equivalence check: on a 1500-token prompt the
greedy output is **bit-identical** to the dense run, 123 t/s prefill both. On a
3500-token prompt the sparse path gives a coherent summary at 121 t/s prefill
and 28.9 t/s generation, while the pre-indexer dense binary on the same prompt
emits `!!!!...`: past the budget the dense attention this port first shipped
with is not an approximation but a collapse. The old limit is gone; what
remains above the budget is that I have no bit-exact reference for the
selection itself beyond the unit test's host model.

End to end on this branch as pushed, with no environment overrides (load,
generation, memory guard):

```text
./ds4 -m gguf/Qwen3.8-Flash-Next-DS4-Q2.gguf --ctx 8192
./ds4 -m gguf/Qwen3.8-Flash-Next-DS4-Q2.gguf --ctx 16384
```

The KV cache is small for this model, so 16384 costs only a couple of hundred
MiB over 8192 and both fit the same host.

Before the memory guard commit the same command was refused with the
diagnostic quoted in that commit's message, so the two belong together on a
64 GiB host unless the user sets `DS4_GLM_MEMORY_GUARD_RESERVE_GB` by hand.

Behaviour: greedy decoding is token-identical to a CPU reference built from the
official FP8 snapshot. Resident in 44.4 GiB, 5.2 s to residency, 30-33 t/s
generation and 58-73 t/s prefill.

Quality, `gguf-tools/quality-testing/score_official` at ctx 4096 against 100
continuations from the FP8 oracle (ChatML nothink, greedy, 24 tokens,
full-vocabulary top-20 logprobs):

```text
summary cases=100 tokens=2252 avg_nll=0.295629595 first_match=66 avg_lcp=7.550
api_summary top1_rate=0.920071048 topn_recall=0.826805086
            pair_rate=0.847376658 target_mae=0.128
```

GLM-5.3-Flash Q2 scores 0.458 NLL / lcp 7.37 and Q4 0.300 / 9.66 against a
hosted FP8 oracle, so this Q2 recipe sits at GLM-Q4 NLL level. `first_match`
66/100 is the lagging metric and the misses cluster on near-equiprobable
openers; imatrix calibration of the IQ2_XXS gate/up tensors is the obvious next
step. The 100-case fixture is 5.1 MB and is not included; say the word and I
will add it.

## Hardware it needs, and what it was proven on

Nothing in the port is tuned to a particular GPU: the Metal dispatch geometry
derives from the model's head counts and head dimension, the prefill chunk is a
fixed 512, and the only host-dependent quantity is Metal's own
`recommendedMaxWorkingSetSize`, queried at runtime. So it is not written for one
machine.

What it does require is that the model stay resident. There is no SSD-streaming
path, so about 45 GiB have to be free at `--ctx 8192`, which in practice means a
64 GB Apple Silicon machine or larger.

What it has been *proven* on is exactly one host: an M5 Pro, 20-core GPU, 64 GB.
Other Apple GPU generations and larger hosts are untested, and so is the memory
guard commit's behaviour on 128 or 256 GB machines, where the recommended
working set is much larger than the reserve it replaces.

## Not validated

`make test` and `./ds4_test --logprob-vectors`, `--long-context` and
`--tool-call-quality` all load `ds4flash.gguf`, which is 81 GiB on this 64 GiB
machine. Attempting it took the host down, twice. CUDA and ROCm are untested
for the same kind of reason: no hardware.

**So the DeepSeek and GLM non-regression is argued below rather than measured,
and that is the part of this PR I would most want checked on your machines.**

## What this does to the existing models

Everything Qwen-specific is behind `ds4_model_is_qwen38()`. The shared-code
changes, worth reviewing first:

- **The GLM4 pre-tokenizer is parameterized.** Qwen breaks digit runs after
  every digit, GLM after every three; GLM keeps `max_digit_run = 3`, so its
  token stream is byte-identical.
- **The routed-MoE block-size check** in `ds4_gpu_routed_moe_one_tensor`
  derives the block length from the gate and down quant types instead of
  assuming 256. Identical for 256-block types; it only additionally admits the
  Q8_0/MXFP4 pairing this model needs.
- **The swiglu clamp** moves to a per-graph field. Pure refactor.
- **Static caps grow** to 512 routed experts and 10 selected, on both the
  engine and the Metal side: about 1.5 MB more BSS for every model.
- **The PLE row store** adds eight reader threads and a 64 MiB cache per
  loaded Qwen model; nothing else sees it.
- **The Apple memory guard change is a real behaviour change** and is a
  separate commit for that reason. With `iogpu.wired_limit_mb` at 0, which is
  its value on every boot, the guard used a blanket reserve and refused a
  45.05 GiB graph on a host where Metal reports a 51.84 GiB recommended
  working set. GLM and DeepSeek graphs that the guard refuses today would
  start loading on Apple. Drop that commit if you would rather decide the
  memory policy separately; everything else stands without it, at the cost of
  making users set `DS4_GLM_MEMORY_GUARD_RESERVE_GB` by hand.

## Known limits

- SSD streaming, `--kv-disk-dir` and `--batched-session` are refused: the
  recurrent and PLE session state cannot be serialized yet.
- No MTP, no vision, no CUDA or ROCm kernels.

There is no download target. `docs/MODELS.md` documents converting from the
official FP8 snapshot with `gguf-tools/qwen38_quantize.py`, which also produces
the PLE sidecar.

## Commits

Eleven commits, each reviewable on its own. The first two touch only
`gguf-tools`, and the CUDA/ROCm stub commit and the memory guard commit are
independent of the rest; all of them cherry-pick cleanly if you want some
without the port. The last two, the PLE row store and the block indexer, sit
on top of the base port and are the ones with measurements in their messages.

---


Baekpica:
@andrea82, Thank you for mentioning. I will add the MTP feature and CUDA kernels to this upstream if I can.
And maybe I could make a more aggressive version of Qwen for 64GB macOS. The target precision could be lower and the PLE artifacts goes from 16bit to 8bit for saving the disk.

PR #1011, opened 2026-09-09
ROCm: accelerate GLM 5.3 prefill and reduce MTP overhead on gfx1151

kyuz0:
GLM-5.3-Flash-Q2 prefill improves 2.20–2.46× on Strix Halo (gfx1151), and embedded-MTP decode improves 22.6% on the completed coding example. Ordinary decode is essentially unchanged.

Resident weights, ROCm 10, greedy native benchmark measurements against main `6289c51`; MTP uses a benchmark adapter. Rates are tokens/s:

| Workload | Main | This branch | Change |
|---|---:|---:|---:|
| Fresh 4K prefill | 76.64 | 168.41 | 2.20× |
| Fresh 16K prefill | 63.77 | 156.89 | 2.46× |
| Ordinary decode, 4K / 16K | 11.82 / 11.17 | 11.82 / 11.16 | ~unchanged |
| Completed coding turn, MTP decode | 12.21 | 14.97 | +22.6% |

Single-run measurements. Coding answers contain 241→267 useful tokens, so throughput is not answer latency. Prefill/ordinary figures precede the final MTP-only commit, which leaves those paths unchanged.

Changes:

- Enable existing IQ2 expert and Q4 projection MMQ paths for measured GLM 5.3 prefill shapes; parallelize grouped-indexer scoring with wave32.
- Reuse Q8 weights across two-token projections and skip an unused accepted-draft prediction while preserving its KV update. The final change alone improves coding 14.52→14.97 tok/s with identical output.
- Invalidate private MTP history after target-only append/evaluation to prevent unwritten draft rows entering its cache window.

Quality evidence:

- Independent kernel references, complete outputs, canaries, boundary sizes and dispatch profiling before model-level tests.
- Main and the kernel/history stage each pass 138 practical checks spanning ordinary/MTP, greedy/sampled generation, long prompts and tool round trips, plus 12 vision fact checks.
- Official long-prompt scoring: 8 cases / 740 targets, weighted NLL improves 1.96%; four cases improve and four worsen. Final-head ordinary short scoring matches main exactly over 100 cases / 11,559 targets.
- The final MTP-only change preserves 717 full-vocabulary state vectors against the preceding candidate, including snapshot/replay and rewind checks, plus tested native IDs/frontiers. Selected DS4F checks preserve logits/output without a material measured slowdown. All five executable help/link checks pass.

CUDA comparison used the identical GGUF and 4,096 raw token IDs at the same context frontier. Across all 154,880 logits:

| Difference from CUDA main | ROCm main | This branch |
|---|---:|---:|
| Full-logit RMS difference | 0.3491 | 0.2769 |
| KL(CUDA ∥ ROCm), nats | 0.0403 | 0.0412 |

All three select the same top token and top-ten set. Distances are mixed; CUDA also uses approximate arithmetic, so this checks consistency rather than establishing higher precision or quality equivalence.

Limits: gains are workload-specific; a separate raw-text MTP control is 19.3% slower than main MTP. Vision was not rerun after the final MTP-only commit, and its checks grade facts rather than strict formatting. Performance paths are guarded for resident ROCm/gfx1151 GLM 5.3.


saxs2224:
Tested this independently on my own Strix Halo box (AMD Ryzen AI Max+ 395 / Radeon 8060S, gfx1151, 128GB unified memory, ROCm 10.0 / HIP 7.15) — same GLM-5.3-Flash-Q2.gguf, on top of `main` `6289c51` as stated.

**Speed** (`ds4-bench --gen-tokens 0`, fresh prefill per context length):

| ctx | main | this PR | speedup |
|---:|---:|---:|---:|
| 4096  | 76.78 | 169.27 | 2.20x |
| 8192  | 71.09 | 162.6-163.5 (3 runs) | 2.29x |
| 16384 | 63.85 | 156.7-157.4 (2 runs) | 2.46x |

Matches your numbers closely (168.41/2.20x @4K, 156.89/2.46x @16K).

**Correctness** — after getting burned by the non-determinism in #923, I checked this one thoroughly:

- `test-glm-attention-rocm`, `test-glm53-kda-rocm`, `test-mxfp4-rocm`: all PASS, 0 failures. (Minor build note: all three standalone test targets fail to link against `main` — `undefined symbol: ds4_engine_is_glm53`, referenced from `ds4_rocm_matmul.cuh` but the Makefile test rules don't pull in `ds4.o` where it's defined. Worked around it locally by stubbing the symbol in the three test .c files; not a code bug, just a Makefile gap worth a follow-up.)
- Determinism: reran the exact same prefill 3x at ctx=8192 and 2x at ctx=16384 with `--dump-frontier-logits-dir`, diffed every pairing — **0/154892 differing logits every time**.
- End-to-end: sent the same ~30.6K-token prompt through `ds4-server` twice (greedy, 150 output tokens), full chat completion — **byte-identical output both times**.

No issues found. This one looks solid — nice work on tracking down the invalidation bug from the last round.

kyuz0:
Thank you for checking and testing this PR! @antirez , whenever you have a second, this looks like a good one to marge.

Piega:
Independent verification of the prefill claim on different hardware — reproduced.

**Hardware**: Bosgame M5 — Ryzen AI Max+ 395 (Radeon 8060S, gfx1151), 128 GB LPDDR5x, 2 TB YMTC PC41Q-2TB-B NVMe, ROCm 10.0 (`rocm/dev-ubuntu-26.04:10.0.0-full`), **stock kernel boot parameters** (no `amdgpu.gttsize` / `ttm.pages_limit` overrides, unlike the setup documented in docs/STRIX_HALO.md).
**Model**: `GLM-5.3-Flash-Q2.gguf`, resident weights, `--power 100`.

Scope note for the resident setup: with stock GTT/TTM defaults this machine exposes only ~62 GiB to the GPU, so the 90 GiB resident GLM arena OOMs on an unmodified main. I measured with the host-memory weight-arena path (PR #1028) added on top of this branch. That arena change is orthogonal to the kernels here, and on systems that apply the documented `amdgpu.gttsize=126976 ttm.pages_limit=... ttm.page_pool_size=...` (≈124 GiB GTT) it should not be needed at all. Mentioning it only so the numbers below are reproducible.

`ds4-bench --prompt-file long.txt --ctx-max 4096 --gen-tokens 128 --rocm --power 100`, same prompt and settings for both builds:

| ctx | main | this branch | change |
|---:|---:|---:|---:|
| 2048 | 64.95 tok/s | **124.00 tok/s** | **1.91×** |
| 4096 | 55.58 tok/s | **118.70 tok/s** | **2.14×** |

Steady generation: 10.86 / 10.68 tok/s (main) vs 10.84 / 10.23 tok/s (this branch) — unchanged, as described.

So the prefill gain holds across a third machine, slightly below your 2.20-2.46× but in the same range (different context points and prompt).

MTP: I could not reproduce the decode delta on this box. `--mtp` is not a `ds4-bench` option (it printed usage), and a CLI comparison on a short NIS2-list prompt gave generation 11.37 tok/s without `--mtp` vs 11.26 tok/s with it — i.e. no measurable change. Your +22.6% is measured on a completed coding turn through a benchmark adapter; is there a supported way for others to reproduce that number (adapter/recipe in the repo, or a specific prompt shape)? Worth documenting so others don't misread the MTP result as a general decode speedup.

PR #1014, opened 2026-09-09
Optimize Q8 decode projections on CUDA, Metal and ROCm

GiorgioOppo:
Q8 decode still incurs separate compressor state writes on CUDA/ROCm and extra synchronization inside Metal's existing fused kernels. On GB10, the shared-expert Q8+HC projection also misses an already available aligned weight layout. This PR addresses those paths directly on upstream `main` (`6289c51`), without the attention-Q4 commit series.

### Changes

- **CUDA / single GB10:** fuse paired F16 compressor projections with the ring-state write; use a bounded, immutable transposed half2 cache with capture-safe fallback. Use existing aligned Q8 weights for shared-expert down+HC, retaining the upstream activation quantizer, F32 scales and HC epilogue. The F16 paths originate in `d12f4480`/`ff749b84`, restored in `4a87d9c`.
- **Metal:** reduce F16 pair, quad and Q8 QKV+quad compressor synchronization from five threadgroup barriers plus a device barrier to one threadgroup barrier. Write projections and state from the same FP32 results. Remove one of two barriers in the shared, general and vector Q8+HC kernels, keeping products, reductions and epilogues unchanged.
- **ROCm / gfx1151:** add the state write to the existing shared-X F16 pair, preserving lane-strided products and the wave32 reduction. Enable it automatically for K=4096, width 256/1024 with ratio 4 and width 512 with ratio 128. Keep the reference fallback for quality mode, graph dumps and unsupported devices/shapes/APE types.

Metal/ROCm add no new cache or opt-in switch. ROCm's existing Q8+HC kernel is unchanged. Bounds/alias checks precede the new ROCm writer; an error after submission is fatal to prevent fallback replay. CUDA's cache and capture behavior are covered by dispatch fault tests.

### Validation

- CUDA host oracles: 72 F16 cases in each contraction mode and 106 policy checks; 40 Q8 kernel cases, 8192 independent integer-dot comparisons and 31 dispatch/fault checks in strict/fast builds. ASan/UBSan pass.
- Metal on **Apple M1 Max**: **68 strict + 68 fast cases pass bitwise** across all six fused entry points. A real-backend quad integration check passes **64 cases / 512 guarded tensor comparisons**, including batched commands, offset views, APE F16/F32 and boundary positions. The complete Metal backend builds and its shader library initializes successfully.
- ROCm host oracles: **216 arithmetic/state cases in each contraction mode**, with ASan/UBSan, plus **87 dispatch/failure checks**. Native HIP source generation passes; HIP compilation and GPU execution remain unverified.
- Native CUDA/HIP tests include fresh-input graph replay checks, but neither backend was compiled or executed on NVIDIA/AMD hardware locally. ROCm's model-level graph API remains unsupported; its new graph fixture exercises kernels directly.

Reproducible commands and backend eligibility are in [`docs/CUDA_GB10_DECODE.md`](https://github.com/GiorgioOppo/ds4/blob/q8-gb10-decode/docs/CUDA_GB10_DECODE.md), including the native GPU tests/benchmarks to run on GB10 and Strix Halo.

### Local component measurements

Actual old fused shaders (`4450d1f`) versus this change on M1 Max, with warm weights, GPU timestamps and eight rounds alternating version order:

| F16 kernel | Old median | New median | Median paired time reduction |
| --- | ---: | ---: | ---: |
| Pair, K4096 / width512 | 33.418 us | 26.868 us | 20% |
| Quad, K4096 / widths1024+256 | 63.660 us | 48.591 us | 24% |

Both F16 cases are faster in every measured round. **Q8+HC shows no stable measurable improvement** after longer repeated-dispatch samples. These are isolated component measurements, including direct shader variants whose host gates differ across Apple generations; they do not establish a model token/s gain or CUDA/ROCm/M3/M5 performance. End-to-end measurements with the same Q8 GGUF remain necessary.

### Reproducible Metal benchmark and real-decode attribution

The component benchmark now ships in the repository, together with a session profiler. This addresses the missing reproduction harness noted in the [external M5 Max report](https://github.com/evanwtf/local-llm/issues/274#issuecomment-5608344436), which passed the 68+68 bitwise checks but did not establish an end-to-end speedup.

```sh
make bench-metal-f16-decode METAL_F16_BENCH_ARGS="--output /tmp/ds4-f16-run1 --samples 8 --repeats 512 --warmup 2 --math both"
make metal-decode-profile
```

The benchmark extracts actual baseline/current shaders, compares five F16/compound cases, alternates version order and saves raw CSV plus JSON provenance. The model profiler tags prefill/warmup/decode batches in the existing encoder timeline and separates exclusive compressor time from compound QKV+F16 time. It also fixes trace contamination by owned/background dispatches and missing timeline attachment after selected-expert readback. No new environment variable or kernel-selection switch is added.

[Full commands, output interpretation and limitations](https://github.com/GiorgioOppo/ds4/blob/q8-gb10-decode/docs/METAL_DECODE_PROFILING.md) include a trace-off control. Reported GPU shares cover traced command buffers only; instrumentation changes encoder boundaries and must not be used as the throughput source for the baseline/PR comparison.

Validation of the tooling on M1 Max:
- Five benchmark cases in strict/fast smoke runs and the original 68+68 bitwise shader tests pass.
- The native timeline test passes six GPU intervals across three restart mechanisms, immutable phase/position tags, guarded numerical comparisons and exclusion of untraced dispatches.
- All 21 parser fixtures pass, including invalid counters, mixed/compound passes, dropped records and CSV joins.
- A short SSD-streamed run using the available local AProjQ4 Vision GGUF records 42 quad and 40 ordinary F16 pair dispatches over two measured tokens. Its trace-off control generates identical token IDs. This is a functional check, not a performance result or a reproduction of the tester's Q8/M5 workload.


GiorgioOppo:
@adamlawi @iammac2 @evandhoffman can you try this pr?

PR #1017, opened 2026-09-09
server: add --reasoning and --alias CLI options

aaa2015:
## Motivation

Clients that expose a model picker can inject thinking defaults that conflict
with the mode a user selected. This adds server-side defaults plus a way to
advertise an instance under a custom model id.

## Changes

- `--reasoning on|off` / `--thinking on|off` — default thinking state when a
  request omits one. The two are synonyms; the error message reports the flag
  the user actually typed.
- `--no-thinking` — shortcut for `--reasoning off`.
- `--reasoning-effort none|minimal|low|medium|high|xhigh|max` — default effort.
- `--alias NAME` — advertise an extra model id in the catalogue so a picker can
  target this instance by name. The id is also served by `GET /v1/models/NAME`.

Alias suffixes select the default thinking mode: `-chat`/`-nothink` disable it,
`-reasoner` enables it. The match is on the suffix rather than a substring, so a
local quant name such as `local-glm-5.3-flash-q2` stays unaffected while a
`...-chat` alias still opts out.

### Turning the default off must not disable thinking permanently

`think_mode_from_enabled()` returns `NONE` whenever the effort is `NONE`, so an
earlier revision that set `default_think_mode = NONE` on `--reasoning off`
pinned every request to non-thinking. A client that opted back in through a
`-reasoner` alias or an explicit `thinking: true` was silently ignored.

Off now flips only the boolean and keeps the configured effort:

```
--reasoning off                  → enabled=false, effort=HIGH
  request without thinking field → NONE
  request with thinking:true     → HIGH   (was NONE before this fix)

--reasoning off --reasoning-effort max
  request with thinking:true     → MAX    (was NONE before this fix)

--reasoning-effort none          → deliberate opt-out, stays NONE
```

## Catalogue bug fix

`send_models()` picked the advertised ids with a family test, but GLM-5.3 and
GLM-5.2 share `DS4_MODEL_FAMILY_GLM_DSA`. A GLM-5.3 engine therefore advertised
`glm-5.2`, `glm-5.2-chat` and `glm-5.2-reasoner`, misreporting the loaded model
to a picker. The ids are now selected by variant, and a single list drives both
the catalogue and the alias-deduplication check.

## Testing

Machine: Apple M4, macOS 26.4.1, 16 GB. Backend: Metal (default build).

```
make ds4_test
./ds4_test --server
→ server: OK
```

`--server` covers the request parsing, chat rendering and thinking-control paths
this change touches. It needs no model file. The model-backed suites
(`--logprob-vectors`, `--long-context`, ...) were not run because this machine
has no GGUF checked out.

New tests:

- `test_server_reasoning_options` — flag parsing, the off/on effort combinations
  above, plus negative alias-suffix cases (`my-chat-model`,
  `local-chat-assistant` must not disable thinking).
- `test_model_catalog_matches_loaded_variant` — the three variant catalogues,
  and every advertised id is accepted by `server_model_alias_known()`.
- `test_send_models_json_shape` — drives the real `/v1/models` body over a
  socketpair and checks the JSON shape, plus that a built-in alias is not
  advertised twice.

No inference-backend code is touched, so no speed regression is expected.


aaa2015:
Rebased onto `8db1d1d` (was `6289c51`, 55 commits behind). Two conflicts, and the second one was not mechanical — worth spelling out, because taking this branch's side verbatim would have introduced a regression.

**Conflict 1 — `model_alias_{disables,enables}_thinking()`.** Both sides extended the same two functions: upstream listed the `qwen3.8-flash-next-*` aliases, this branch added suffix matching. Both are kept, so the suffix rules and the full exact lists coexist and the aliases upstream added resolve exactly as before.

**Conflict 2 — `send_models()`.** Upstream rewrote the body with per-engine branches; this branch rewrote it to drive the catalogue and the alias-dedup check from one id list. This branch's shape won, since one list is the point — but its table only knew glm53 / glm52 / deepseek. On current main that means a Qwen3.8 or DeepSeek V4.1 engine would fall through to the DeepSeek ids and advertise `deepseek-v4-flash` / `deepseek-v4-pro` for a model it is not running. Upstream grew both engines after this was written, so the table now covers all five families in the same precedence the engine-specific branches used, with glm53 ahead of the broader glm_dsa test.

`test_model_catalog_matches_loaded_variant` is extended to pin qwen4 and V4.1 as well, so the fall-through cannot come back silently.

`ds4_test --server`, `make test-session-state` and `make test-tp-commands` pass.

Issue #998, opened 2026-09-07
GLM-5.2 Q4: prefill fails at ≥64K context on M3 Ultra 512GB (KV payload staging failed)


wbllcbbg-ai:
## Environment
- **Hardware**: Apple M3 Ultra, 512GB unified memory, 819GB/s bandwidth
- **Model**: GLM-5.2-UD-Q4_K_RoutedQ4K.gguf (404GB, from antirez/GLM-5.2-GGUF on HuggingFace)
- **Engine**: ds4-server, built from latest main (Metal backend)
- **Context**: --ctx 524288 (512K)
- **Command**: ./ds4-server -m GLM-5.2-UD-Q4_K_RoutedQ4K.gguf --ctx 524288 --port 8321 --kv-disk-dir ~/.ds4/server-kv --kv-disk-space-mb 262144

## What works
- Short context (2K, 16K): prefill and decode work correctly
  - 2K: prefill 116 t/s, decode 16.0 t/s
  - 16K: prefill 112 t/s, decode 14.2 t/s
- Memory is sufficient (model 404GB + graph ~50GB = ~455GB, well within 512GB)
- Quality: needle tests pass when thinking mode is disabled

## What fails
At ≥64K context, prefill fails with:
```
ds4-server: kv cache skipped tokens=10240 reason=continued because KV payload staging failed: se...
ds4-server: kv cache discarded reason=prefill-failed
```

After the failure, the engine stops responding to new requests (appears hung).

Reproducible on a fresh process. Same hardware runs DeepSeek V4 Flash MXFP4 at 1M context without issues.

## Log excerpt
```
0907 02:11:17 ds4-server: kv cache skipped tokens=10240 reason=continued because KV payload staging failed: se...
0907 02:35:35 ds4-server: chat ctx=2048..79031:76983 prompt start
0907 02:35:35 ds4-server: kv cache discarded reason=prefill-failed file=...kv
```

## Notes
- This might be related to GLM's IndexShare architecture or the compact DSA KV layout at larger contexts
- The model loads fine and memory guard passes at 512K ctx (required: ~455GB, budget: 480GB)
- Happy to provide more logs or test patches

wbllcbbg-ai:
Update: resolved on current main (`f62ca29`) on the same M3 Ultra 512GB.

[...]

PR #1024, opened 2026-09-10
Enable GLM 5.3 tensor parallelism on the ROCm backend

davidcanar:
Enables GLM 5.3 (GLM_DSA family) on the two-node ROCm tensor-parallel backend from #813, then makes it fast enough to use. Before this, GLM TP never armed on ROCm: the GLM graph's TP initialization was guarded `#if defined(__APPLE__)`, so `tp_world` stayed 0 and every rank silently computed the full expert set against its half-resident weights (word salad, no error). With this PR, GLM-5.3-Flash Q4_K (177.76 GiB, unrunnable on one box) runs across two 128 GB Strix Halo boxes, including 47k-token contexts and ds4-server tool-call conversations.

Opt-in, like the backend itself: `DS4_ROCM_GLM_TP=1` on both ranks.

## Branch layout (stacked)

Integration lineage of `rocm-tensor-parallel` (#813) with upstream main merged in (`bd177c9`/`da15223`, `8025b22`, `0fff703`, `5dc45b7`). #813 stays at its original trunk point as the historical base (it shows conflicts with `main` and is not being rebased), so its ROCm TP transport is reviewable here; #835 and #1018 are kept mergeable on their own. Stack order: #813 → #835 → #1018 → this PR. Commits new here:

- Enablement: `ea7057d` merge fixups, `4dd3f4d` GLM TP on ROCm (the core change), `848fb67` lock static weight spans under streaming, `1a646f6` avoid the faulting Q4_K sorted down kernels, `682cdb4` restore the CPU build.
- Transport: `0e02528`, `4402b61`, `cb7cb70` verbs fixes (RDMA bulk-window reset, `DS4_TP_GATE_TIMEOUT_MS`, RC queue pairs, real `ibv_post_*` codes); `7598395` OdinLink loss recovery (same as #1018); `2a0977d` odl batch/big gates without the TCP header rendezvous (same as `15b6d6b` in #1018).
- Decode performance: `5272bda` pre-quantized k-slice Q8_0 matvec, `5b31f6d` gate arrive/wait split with the shared expert overlapping the FFN exchange, `b6f965b` vectorized small-output BF16 row matvec, `ebfe289` no periodic decode-stream flush on ROCm, `9b708e9` vocab-split output head, `50a3f9f` fewer and wider small launches, `5da2545` tiled long-context DSA kernels, `8f3505c` KDA attention split by heads, `4dac2b5` one-wave-per-row Q4_K expert gate/up kernel.
- Prefill performance: `10f5531` Q4_K routed experts through the vendored MMQ tensor-core GEMMs, `e9cfbde` indexer scores as batched f32 GEMMs, `e39e8dd` batch FFN big gate overlapped with the shared expert.
- Portability: `91ae988` guards the ROCm-only split gate wait (Metal and CUDA keep the single-call gate); the CPU build stub for the KDA state sync came in with `0fff703`.
- Disk KV cache under TP: `70b16d9` restores snapshots on both ranks (the leader restores, then streams the same payload to the worker over the control socket; before this a TP disk hit rebuilt both ranks with a full prefill, so a fresh 15k-token session took 123 s and now takes 4.3 s with a byte-identical answer); `5801a05` stages snapshots in the cache directory instead of /tmp (RAM-backed tmpfs, which OOM-killed the server once with a 6 GiB snapshot).

## Changes

- GLM graph TP init guard extended to `DS4_ROCM_BUILD`; the refusal gated behind `DS4_ROCM_GLM_TP=1`; GLM exempted from the Q8_0 output-head requirement (its head is replicated under TP).
- Streaming: operator-precedence fix in the ROCm GLM streaming gate; `mlock` of SSD static spans on ROCm and post-upload discards that skip locked spans (the GPU cannot fault file-backed pages back in).
- Q4_K sorted-pairs down GEMM routed through the proven per-pair kernel, `moe_sum` epilogue kept (see failures below).
- Decode path on ROCm TP: Q8 k-slice on prequantized activations; gate arrive/wait split; each rank projects half of the 154880-row head during single-token decode and the session merges the halves (off with `--mtp`); one launch for the KDA q/k/v projections and one for f_a/g_a/beta; 1024-lane float4 RMS norms; float4 router matvec.
- Long-context DSA layers: the indexed decode attention is now three kernels (a 16-key LDS tile scored against every head with RoPE applied once per key, a pv kernel handling all of the rank's heads per 128-key chunk, a chunk-order reduce), the indexer scorer runs one wave per key row. Old kernels stay as fallback (`DS4_ROCM_GLM_LEGACY_INDEXED_ATTN=1`). Kernel time per DSA layer 1000 + 776 us → 98 + 66 us.
- KDA head split (single-token decode only): each rank runs 32 of the 64 KDA heads (half-row projections, the conv + delta-rule recurrence on its half of the layer state, output projection K-sliced into the layer's ATTN gate, now in the gate schedule: 87 gates per token). Replicated KDA paths (prefill/verify rows, batched-session decode, session save via the new `DS4_TP_FRAME_GLM_KDA_SYNC` frame) re-gather the state halves first with one big-gate swap per layer. Off with `--mtp`, SSD streaming, an unsupported `kda_output` type, or `DS4_GLM_TP_KDA_HEAD_SPLIT_DISABLE=1` on both ranks. BF16 gets `ds4_gpu_glm53_matmul_bf16_kslice`.
- Prefill experts: Q4_K batches of 16+ tokens compact the rank's owned (token, slot) pairs and run gate/up and down as `ds4_mmq_q4_K_moe_pair` / `ds4_mmq_q4_K_moe` calls with one expert per row, then the existing scatter + `moe_sum` epilogue (gfx1151 default, `DS4_ROCM_MMQ_Q4K=0` keeps the tile kernels). The tile kernels ran at a few percent of both roofs.
- Prefill indexer: the GLM-5.3 indexer scores (every query against every pooled key, 34% of a 24.7k-token prefill in a one-block-per-pair kernel) run as strided-batched f32 hipBLAS GEMMs over the heads plus a reduce kernel for relu, per-head weights and the causal pooled mask; same f32 products, `DS4_ROCM_GLM_INDEXER_GEMM_CHECK=1` reports a relative deviation of ~3e-7 against the block kernel. `DS4_ROCM_GLM_INDEXER_GEMM=0` restores it.
- Batch FFN under TP: the routed-partial swap is kicked before the shared expert is encoded and waited on after it (`ds4_gpu_tp_big_gate_kick_nowait` / `ds4_gpu_tp_big_gate_wait_encode`); payload and add unchanged.

## Test evidence (per CONTRIBUTING.md)

Machine: 2x AMD Ryzen AI Max+ 395 (Radeon 8060S, gfx1151, 128 GB unified), Fedora 44, ROCm, Thunderbolt 5 direct link (odl transport from #1018; rdma/tcp also work). Model: GLM-5.3-Flash Q4_K (`glm53-q4` GGUF, sha256 c7a0d950...). Build on both boxes:

```sh
make strix-halo -j32 CFLAGS="-O3 -ffast-math -g -march=native -Wall -Wextra -std=c99 -fPIC -D_GNU_SOURCE -fno-finite-math-only" \
  ROCM_LDLIBS="-lm -pthread -lamdhip64 -lhipblas -lhipblaslt -lrocblas"
make cpu   # also verified clean
```

```sh
# box2                                                  # box1
env DS4_ROCM_GLM_TP=1 ~/ds4/ds4 -m GLM-5.3-Flash-Q4_K.gguf --tensor-parallel --role worker \
  --coordinator 10.0.2.1 7300 --transport odl --ctx 8192
env DS4_ROCM_GLM_TP=1 ~/ds4/ds4 -m GLM-5.3-Flash-Q4_K.gguf --tensor-parallel --role coordinator \
  --listen 10.0.2.1 7300 --transport odl --ctx 8192 -p "<prompt>"     # or --prompt-file
```

Correctness runs, all clean (no faults, desyncs or timeouts on either rank): short exact-prefill prompts ("two plus two" → "Four"); 227- and 323-token prompts with accurate summaries; a 47,288-token prompt at ctx 49152 with an accurate summary of the fed source file; ds4-server with `--batched-session 4` (two-turn weather-tool conversation with KV reuse, the same under a concurrent generation, a 6,181-token multi-chunk tool prompt reusing 6,192 cached tokens); after the head split, a server session with three turns, two concurrent requests, a follow-up, and an 810-token conversation saved to the KV disk cache.

Numerical checks: `DS4_ROCM_GLM_DSA_CHECK=1` and `DS4_GLM_TP_KDA_CHECK=1` run the previous or replicated kernels next to the new ones on the same inputs; over ~1000 (DSA) and 2800 (KDA) layer calls per rank the relative L2 deviations stay at 1e-7 to 1e-6, no non-finite values. Greedy texts flip at near-tie tokens against earlier builds (changed accumulation order; int8 activations in the k-slice path), with both continuations correct; the 45-token test prompt reproduces earlier outputs byte for byte across several builds. The MMQ prefill path was checked by running the same prompts with the path on and off.

## Performance

2x gfx1151 over odl, greedy. "Q8-attention GGUF" is a data-only requant of the official file (KDA projections + head to Q8_0) used to measure the code with the BF16 tensors out of the way; it is not part of the PR.

| Test | first submission | now |
|---|---|---|
| official Q4_K GGUF, 45-token prompt decode | 8.9 t/s | 13.1 t/s |
| Q8-attention GGUF, 45-token prompt decode | 15.4 t/s | 19.4 t/s |
| Q8-attention GGUF, decode after 24,764 tokens | 11.8 t/s | 19.0 t/s |
| Q8-attention GGUF, decode after 44,267 tokens | 11.1 t/s | 15.4 t/s (at `5da2545`, not re-measured since) |
| Q8-attention GGUF, prefill 323 / 2,500 / 24,764 tokens | 69 / 72 / 52 t/s | 146 / 140 / 122 t/s |
| Q8-attention GGUF, MTP 1 draft | 10.4 t/s | 11.6 t/s (plain decode remains faster) |
| row-gate exchange (odl), average | 172 us | 86-106 us |

Where the time went, from roctracer timelines on both ranks: the first decode step (114 ms on the official GGUF) was 57 ms of BF16 KDA projections streamed at 160 GB/s, 13 ms of experts, 8.7 ms of gate spin, 5 ms of an 80 GB/s k-slice kernel, ~17 ms of ~1,500 small launches. After the rounds above a Q8-model step is ~52 ms: ~35 ms of memory-bound kernels near the 227 GB/s roof, ~6 ms of gates (84 exchanges of ~90 us, the odl round trip plus rank skew), ~4 ms of launch gaps, ~6 ms of small kernels. Prefill on a 2,500-token prompt spent 59% of its time in the Q4_K expert tile kernels before `10f5531`; at 24.7k the indexer scoring kernel was another 34% before `e9cfbde`. What remains of a 24.7k prefill (~200 s): MMQ expert GEMMs ~47 s, the exposed part of the big-gate exchanges ~20 s, the sequential KDA prefill recurrence 23 s, hipBLAS attention and dense GEMMs ~70 s.

Two intermediate pv-kernel designs for the long-context attention both sat at 200-240 us per layer: the scattered KV rows come from DRAM every token and both designs kept the same number of loads in flight per CU; reading each row once for all heads over 16 blocks brought it to 68 us.

## Transport reliability (`7598395`)

The OdinLink transport lost whole 1 MiB gate messages whenever a Thunderbolt frame failed CRC (`odl_tb5: RX CRC error ... dropping message`), which parked both ranks on their spin kernels until the 300 s bulk timeout and killed the session (seen 4096 tokens into an 11k-token ds4-server prefill). The loss-recovery commit reviewed in #1018 (per-message headers, receiver-driven retransmission from retained frames, idle ranks serving requests from their control-channel waits) is cherry-picked here. Validated with `DS4_TP_ODL_DROP_TEST`: 15.35 vs 15.23 t/s decode with every 400th gate message dropped (output byte-identical), 69.1 vs 67.5 t/s prefill with every 50th dropped (output identical).

## Speculative decoding (MTP)

Works with two-rank TP (CLI and ds4-server, 226 tokens to 47k, tool-call conversations) with no code beyond this PR; runtime settings: `--mtp` on both ranks, `DS4_GLM_MTP_DRAFTS=1`, `--batched-session 1` on ds4-server. It remains a net slowdown here: the 2-row verify span costs about twice a plain decode step and draft agreement stays below break-even. Default-off is right for this hardware. The KDA head split and the vocab-split head turn themselves off with `--mtp`.

## Notable failures root-caused on the way

1. **Q4_K sorted-pairs down-projection fault.** Batch prefills of 32+ tokens aborted both ranks with `HSA_STATUS_ERROR_MEMORY_APERTURE_VIOLATION` (surfacing as "GLM TP batch gate failed" in ds4-server). Stage-skipping isolated it to the sorted/tile down kernels (`moe_down_q4K_expert_tile{4,8}_row32`, `moe_down_q4K_sorted_qwarp32`); every input validates clean and the per-pair kernel reads the same weights without faulting. `1a646f6` routes the sorted-path down GEMM through the per-pair kernel and keeps the `moe_sum` epilogue for q4k (without it, batches of 128+ tokens silently dropped the MoE output). Why the tile down kernels fault is still open (possibly related to #890 / #998 / #608 / #595); with `10f5531` the prefill down projection runs through MMQ anyway.
2. **Static-weight pageout under streaming**, fixed by the mlock extension and discard guards in `848fb67`.

## Not done / caveats

- Balancing the routed experts by splitting each expert's FFN dimension across ranks: the Q4_K `down` matrices' K-halves are strided per row, so it needs whole `down` matrices per rank or a ~27 GB repack at load.
- Prefill stays replicated for the KDA layers (a split would need a Q8 K-slice GEMM on ROCm and 34 more big gates per chunk).
- The KV disk-cache restore of a session under TP was not exercised after the KDA head split (saves were; the restore path clears the split flag).
- Not run: the full `make test` suite (needs the CPU test fixtures); everything above was validated end-to-end on real hardware on both ranks.


MerlijnWajer:
Thanks for working on this, I'm excited to try to get this working locally (using `docker.io/kyuz0/strix-halo-ds4-toolbox:rocm-10.0` as docker base). I gave this a try with GLM 5.3 Flash (`GLM-5.3-Flash-Q4_K.gguf`) and I get this error on the coordinator:

```
0911 11:07:21 ds4-server: KV disk cache /tmp/ds4-kv (budget=100000 MiB, cross-quant=accept, min=512, cold_max=100000, continued=10000, trim=32, align=2048, hit_half_life=21600s)
0911 11:07:21 ds4-server: listening on http://0.0.0.0:8000
0911 11:07:53 ds4-server: chat ctx=0..1322:1322 TOOLS prompt start
ds4: ROCm GLM causal indexed prefill using fp16 hipBLAS attention GEMMs (tokens=1318 rows=1318 cache=f16)
ds4-tp: big gate window barrier failed
ds4: GLM TP batch gate failed (layer 3)
```

And this error on the worker:

```
ds4: memory: KV 0.56 GiB (raw 0.00 + compressed 0.56) + buffers 3.01 GiB + resident model 96.13 GiB = 99.69 GiB planned
ds4: memory detail: ctx=50000 prefill_cap=2048 raw_kv_rows=0 compressed_kv_rows=50000 backend=rocm
ds4-tp: rdma device rocep3s0f0 (port state 4)
ds4-tp: rdma device limits: max_qp_wr 8192, max_sge 30, max_cqe 4194303, max_mr_size 18446744073709551615
ds4-tp: leader connected, transport=rdma gate-timeout=750ms
ds4-tp: rdma warm-up ok (1 attempt), queue depth recv 1024 send 1024
tensor parallelism bound: rank 1, 50/50 expert split, rdma transporttp worker ready for mirrored sessionsds4: GLM session ctx=50000 (model max=1048576); prefill/work cap=4096; dense attention limit=2051
ds4: GLM graph using compact DSA KV only; expanded full-attention KV cache is skipped
ds4: GLM graph allocating compact DSA cache: rows=50000 logical_ctx=50000 kv_layers=45 indexer_layers=11 f16 0.56 GiB
ds4: GLM compact indexed prefill chunk=2048 score_rows=2048 score_scratch=97.66 MiB
ds4: ROCm GLM causal indexed prefill using fp16 hipBLAS attention GEMMs (tokens=1318 rows=1318 cache=f16)
ds4-tp: timeout in big gate window (237/256 recvs, 4/4 sends, 256 sent)
ds4: GLM TP batch gate failed (layer 3)
```

Commands that I use:

```
merlijn@strixhalo:~$ cat ./run-ds4-tp.sh
#!/bin/sh
sudo docker run --rm -it --net host --ulimit memlock=-1 -e DS4_ROCM_GLM_TP=1   --device /dev/kfd --device /dev/dri --device /dev/infiniband   --group-add video --group-add render   --ipc=host --cap-add=SYS_PTRACE --security-opt seccomp=unconfined  -v ~/ds4:/ds4  -v ~/ds4/gguf:/models -v ~/ds4/ds4-kv:/tmp/ds4-kv  ds4-glm   /ds4/ds4-server -m /models/GLM-5.3-Flash-Q4_K.gguf --kv-disk-dir /tmp/ds4-kv --kv-disk-space-mb 100000 -t 32 --ctx 50000 --role coordinator --tensor-parallel --transport rdma --listen 0.0.0.0 1234  --host 0.0.0.0  --kv-cache-cold-max-tokens 100000
```


```
merlijn@strix-halo-2:~$ cat ./run-ds4-worker-tp.sh
#!/bin/sh
sudo docker run --rm -it --net host --ulimit memlock=-1 -e DS4_ROCM_GLM_TP=1  --device /dev/kfd --device /dev/dri --device /dev/infiniband   --group-add video --group-add render   --ipc=host --cap-add=SYS_PTRACE --security-opt seccomp=unconfined   -v ~/ds4/gguf:/models -v ~/ds4:/ds4   ds4-glm   /ds4/ds4-server -m /models/GLM-5.3-Flash-Q4_K.gguf -t 32 --ctx 50000 --kv-disk-dir /tmp/ds4-kv --kv-disk-space-mb 8192   --role worker --tensor-parallel --coordinator 192.168.100.1 1234  --kv-cache-cold-max-tokens 100000
```


I see a different error when using the 0731 model (`DeepSeek-V4-Flash-Q4KExperts-F16HC-F16Compressor-F16Indexer-Q8Attn-Q8Shared-Q8Out-chat-v2-imatrix-0731.gguf`):

coordinator:

```
ds4: sharded warm done in 26.1s (checksum=2515874312)
ds4: ROCm preparing model tensor mappings: 8.20 GiB
ds4: ROCm startup model preparation covered 8.20 GiB of tensor spans in 0.001s
ds4: rocm backend initialized for graph diagnostics
ds4: memory: KV 1.00 GiB (raw 0.36 + compressed 0.64) + buffers 0.38 GiB + resident model 80.76 GiB = 82.14 GiB planned
ds4: memory detail: ctx=50000 prefill_cap=4096 raw_kv_rows=4352 compressed_kv_rows=12502 backend=rocm
ds4-tp: waiting for worker on 0.0.0.0:1234 ...
ds4-tp: rdma device rocep3s0f0 (port state 4)
ds4-tp: rdma device limits: max_qp_wr 8192, max_sge 30, max_cqe 4194303, max_mr_size 18446744073709551615
ds4-tp: worker connected, transport=rdma gate-timeout=750ms
ds4-tp: rdma warm-up ok (1 attempt), queue depth recv 1024 send 1024
tensor parallelism bound: rank 0, 50/50 expert split, rdma transport0911 11:14:00 ds4-server: context buffers 1414.48 MiB (ctx=50000, backend=rocm, prefill_chunk=4096, raw_kv_rows=4352, compressed_kv_rows=12502)
0911 11:14:01 ds4-server: KV disk cache /tmp/ds4-kv (budget=100000 MiB, cross-quant=accept, min=512, cold_max=100000, continued=10000, trim=32, align=2048, hit_half_life=21600s)
0911 11:14:01 ds4-server: listening on http://0.0.0.0:8000
0911 11:14:39 ds4-server: chat ctx=0..1534:1534 TOOLS prompt start
ds4-tp: timeout in big gate window (151/256 recvs, 4/4 sends, 256 sent)
ds4-tp: timeout in big gate window (124/256 recvs, 4/4 sends, 256 sent)
ds4: TP prefill attention row gate failed (layer 1)
ds4: gpu layer 1 attention batch encode failed
ds4: gpu layer-major prefill layer 1 encode failed
```

worker:

```
ds4: warming sharded tensor pages (rank 1): 80.76 of 153.32 GiB
ds4: sharded warm done in 19.5s (checksum=2515656570)
ds4: ROCm preparing model tensor mappings: 8.20 GiB
ds4: ROCm startup model preparation covered 8.20 GiB of tensor spans in 0.001s
ds4: rocm backend initialized for graph diagnostics
ds4: memory: KV 1.00 GiB (raw 0.36 + compressed 0.64) + buffers 0.38 GiB + resident model 80.76 GiB = 82.14 GiB planned
ds4: memory detail: ctx=50000 prefill_cap=4096 raw_kv_rows=4352 compressed_kv_rows=12502 backend=rocm
ds4-tp: connecting to 192.168.100.1:1234 ... (Connection refused)
ds4-tp: rdma device rocep3s0f0 (port state 4)
ds4-tp: rdma device limits: max_qp_wr 8192, max_sge 30, max_cqe 4194303, max_mr_size 18446744073709551615
ds4-tp: leader connected, transport=rdma gate-timeout=750ms
ds4-tp: rdma warm-up ok (1 attempt), queue depth recv 1024 send 1024
tensor parallelism bound: rank 1, 50/50 expert split, rdma transporttp worker ready for mirrored sessionsds4-tp: timeout in big gate window (145/256 recvs, 4/4 sends, 256 sent)
ds4: TP prefill attention row gate failed (layer 1)
ds4: gpu layer 1 attention batch encode failed
ds4: gpu whole-prefill layer 1 encode failed
ds4-tp: timeout in big gate window (140/256 recvs, 4/4 sends, 256 sent)
```

EDIT: I'll try to use an agent to see if I can track this down to an issue on my side or not.

davidcanar:
Thanks for the detailed report — it's diagnosable from your logs, and the good news is this doesn't look like the GLM code in this PR.

**What your logs show.** All three lines are one event. Your coordinator arrived at a TP bulk-gate window barrier first and waited 750 ms for the worker: the barrier is a one-byte handshake on the control socket, and that socket's receive timeout is `DS4_TP_GATE_TIMEOUT_MS` (750 ms by default). The worker arrived late — its handshake byte was already buffered, so it proceeded — and streamed its half of the exchange into a partner that had already bailed. That's the worker's `timeout in big gate window (237/256 recvs, ...)`: that timeout only fires when the RDMA completion queue is completely idle, i.e. the coordinator had genuinely stopped sending. `GLM TP batch gate failed (layer 3)` is just the graceful fallout on the coordinator. In short: per-layer arrival skew between the two ranks exceeded the 750 ms gate deadline.

**Why I don't think it's the GLM changes:** your DeepSeek-V4-Flash run fails identically (`TP prefill attention row gate failed (layer 1)`, same window timeouts) — and that code path predates this PR entirely. When both models fail the same way on the first prefill chunk, the cause is transport/environment/timing rather than the GLM enablement. We see the same failure family on our own boxes at very large contexts: at 128k the per-layer GPU work between gates grows past 750 ms and the pair dies exactly like yours.

**Things to try, in order:**

1. Add `-e DS4_TP_GATE_TIMEOUT_MS=10000` to **both** docker runs. This is the knob for exactly this signature; at 128k context it took us from hard failure to fully clean runs.
2. Run the worker as the CLI binary instead of `ds4-server`: `/ds4/ds4 -m /models/GLM-5.3-Flash-Q4_K.gguf --role worker --tensor-parallel --coordinator 192.168.100.1 1234 --ctx 50000` (plus the same env). That is the configuration validated in this PR; `ds4-server --role worker` is permitted but was not part of our validation matrix. Also drop `--kv-disk-dir` and `-t 32` from the worker — it needs neither.
3. A/B with `--transport tcp` on both sides. If TCP works, the RDMA path is implicated — RoCE without priority flow control is notoriously stall-prone under bursty 16 KiB message streams, and Docker's `/dev/infiniband` plumbing can contribute.
4. If it still fails at a modest `--ctx` (say 8192) with the larger timeout, please share `rdma link show` and `ibv_devinfo -d rocep3s0f0` output (hardware NIC vs rxe/soft-RoCE, MTU, direct-attached or via a switch).

I've also pushed a small hardening commit to this branch (0e02528): the window-timeout path now resets the RDMA bulk state — previously only the completion-error path did, so after one timeout every later window failed worse, which matches the degrading recv counts in your DeepSeek log (151 → 124) — and the barrier/window timeout messages now name `DS4_TP_GATE_TIMEOUT_MS` so the knob is discoverable from the log itself.

MerlijnWajer:
Thanks for the reply. I did try the higher timeout (in both the coordinator and the worker) and that alone didn't help.

Please find below the output that you asked for from both machines. I have a local deepseek v4 flash trying to see if it can figure out the problem as well (but I don't know if it will).

```
merlijn@strixhalo:~$ rdma link show
link rocep3s0f0/1 state ACTIVE physical_state LINK_UP netdev enp3s0f0np0 
link rocep3s0f1/1 state DOWN physical_state DISABLED netdev enp3s0f1np1 
merlijn@strixhalo:~$ ibv_devinfo -d rocep3s0f0
hca_id: rocep3s0f0
        transport:                      InfiniBand (0)
        fw_ver:                         14.32.1912
        node_guid:                      9803:9b03:0001:9952
        sys_image_guid:                 9803:9b03:0001:9952
        vendor_id:                      0x02c9
        vendor_part_id:                 4117
        hw_ver:                         0x0
        board_id:                       MT_2470112034
        phys_port_cnt:                  1
                port:   1
                        state:                  PORT_ACTIVE (4)
                        max_mtu:                4096 (5)
                        active_mtu:             1024 (3)
                        sm_lid:                 0
                        port_lid:               0
                        port_lmc:               0x00
                        link_layer:             Ethernet
```


```
merlijn@strix-halo-2:~$ rdma link show
link rocep3s0f0/1 state ACTIVE physical_state LINK_UP netdev enp3s0f0np0 
link rocep3s0f1/1 state DOWN physical_state DISABLED netdev enp3s0f1np1 
merlijn@strix-halo-2:~$ ibv_devinfo -d rocep3s0f0
hca_id: rocep3s0f0
        transport:                      InfiniBand (0)
        fw_ver:                         14.32.1912
        node_guid:                      9803:9b03:0096:49f0
        sys_image_guid:                 9803:9b03:0096:49f0
        vendor_id:                      0x02c9
        vendor_part_id:                 4117
        hw_ver:                         0x0
        board_id:                       MT_2470112034
        phys_port_cnt:                  1
                port:   1
                        state:                  PORT_ACTIVE (4)
                        max_mtu:                4096 (5)
                        active_mtu:             1024 (3)
                        sm_lid:                 0
                        port_lid:               0
                        port_lmc:               0x00
                        link_layer:             Ethernet

```

davidcanar:
Thanks — that devinfo helps, and the negative result on the timeout is itself informative: the drain loop only declares a timeout when the RDMA completion queue is completely idle, so a larger `DS4_TP_GATE_TIMEOUT_MS` can't fix chunks that are never going to arrive. Combined with your devinfo, my suspicion moves to the RDMA link/Docker path itself.

Two things stand out from the output:

1. **`active_mtu: 1024` (while `max_mtu: 4096`).** The ds4 RDMA path streams 16 KiB SEND messages in 256-chunk windows, so at MTU 1024 each window is ~4096 wire packets. MTU 1024 usually means the module/DAC or port config negotiated down; ConnectX-4 Lx is also picky about cables. Worth trying: set `ip link set enp3s0f0np0 mtu 4096` on **both** boxes and confirm `active_mtu` rises to 4096, then retry.
2. ConnectX-4 Lx on fw 14.32 is an old-ish stack; RC retransmission under a lossy or flow-control-less RoCE link can stall a QP for a long time, which matches a silent CQ.

On our side, we just ran the TCP A/B I mentioned, end-to-end on our own two Strix Halo boxes with the same model and your context scale: `--transport tcp` completes the full **47,288-token** prefill + decode cleanly (prefill 38.5 t/s, generation 6.5 t/s, zero gate failures), and a short 227-token prompt also passes (51.3 / 8.1 t/s). So the TCP transport is fully functional at your context size — if it also works for you, the RoCE/Docker path is confirmed as the culprit and you can either stay on TCP (25G is more than enough; our TCP numbers match our OdinLink TB5 numbers within noise) or debug the RDMA path with:

```
# one box:
ib_send_bw -d rocep3s0f0 -R
# other box:
ib_send_bw -d rocep3s0f0 -R <ip-of-first-box>
```

`ib_send_bw` is the standard Mellanox send-burst benchmark; if a plain 16 KiB send stream can't sustain ~line rate with zero errors there, the link (cable/MTU/switch) is the problem and ds4 is just the messenger. Also useful after a failed ds4 run: `ethtool -S enp3s0f0np0 | grep -iE "crc|drop|error|discard"` on both boxes — rising counters would directly confirm link-level drops.

Two more things that would help narrow it down:

- The failure logs from the raised-timeout run — did the failing line move (e.g. the barrier timeout now at 10 s instead of 750 ms), and at which layer?
- The CLI-worker and bare-metal (outside Docker) variants if you get a chance: the PR's validation used the CLI worker binary, and bare metal rules out the container's `/dev/infiniband` plumbing.

For transparency: our boxes have no verbs devices at all (we run the OdinLink TB5 transport from #1018), so we can't reproduce your RoCE path directly — the tests above are the fastest way to pin it down from your side.

MerlijnWajer:
Thank you for the detailed reply, I will report back later this weekend. I do have this fork (`https://github.com/wkljohn/ds4-strix-halo-tp-odinlink`) working with RDMA on the same hardware at 20t/s tg, so I think the hardware stack is capable/functional, but you mention some good points and I'll make sure to check them and report back.

MerlijnWajer:
Okay, so I got things to work with this commit created by deepseek-v4-flash-0731 using pi: 
https://github.com/MerlijnWajer/ds4/commit/6d2dc68ba0640cea79596e6de8e877e5177d821b

The tcp transport already worked as you mentioned, but the above commit makes the rdma transport work for me as well.

GLM Flash 5.3 runs at similar speed: ~50 t/s prefill, 8t/s gen. Deepseek v4 flash runs at 14 t/s gen (slower than the 20t/s from the repo I mentioned earlier). 

As you can see the link itself is stable, although I think I might be able to do some work in configuring my switch so that it can handle larger MTUs.

I don't know if the above fix belongs in this branch or not, but it made things work for me. The commit also adds some diagnostics code which may or may not be useful.

In any case, thanks for all the help, this is really great work, as it also makes GLM Flash 5.3 work in a distributed manner (it doesn't on the main branch). I might ask GLM Flash 5.3 if it can figure out why the `https://github.com/wkljohn/ds4-strix-halo-tp-odinlink` fork runs at 20 t/s gen as opposed to the 15 t/s gen on deepseek v4 flash, but I don't want to hijack the PR any further. Please let me know if I can help with anything or what other tests you'd like me to run if any.

davidcanar:
This is a great catch, and your diagnosis is exactly right — thank you. The `UC` in our transport is a fossil from its origin: it was written for the Apple thunderbolt-ibverbs driver, where only UC queue pairs exist. On a real NIC, UC's "drop a SEND that lands before the peer's recv WR is armed" behavior explains everything about your failure signature, including why no timeout setting could help (the chunks were genuinely never going to arrive, and the completion queue stays silent because UC reports nothing).

You were also right about our 0e02528 hardening being wrong: `tp_rdma_bulk_reset()` rebuilds the bulk QP while the orphaned window lives on the latency QP, and its peer-coordinated handshake can't complete on an asymmetric failure anyway. I've dropped that call — recovery from a window timeout needs a symmetric latency-QP rebuild and is noted as future work; under RC the transport-level retries make the timeout a true dead-peer signal.

I've integrated your fix into this branch as 4402b61, with one change: instead of switching to RC unconditionally, the transport now **probes RC and falls back to UC** — the Apple thunderbolt driver is UC-only, and the rdma path is shared code, so a hard switch would have broken macOS tensor parallelism. On RC it sets the RTR/RTS attributes UC doesn't consume (same values as your commit: `min_rnr_timer=12`, `timeout=14`, `retry_cnt=7`, `rnr_retry=7`, `max_rd_atomic=1`), and it includes your DS4_TP_CQ_DEBUG queue dump on the timeout path.

Since you have the only verbs hardware in the loop: could you pull `davidcanar:glm-tp-rocm` and re-run your RDMA setup once? The probe should log `ds4-tp: rdma using RC queue pairs` at startup, and everything else should behave like your commit.

And nicely done getting DeepSeek-V4 to produce a working verbs fix — the commit message reads like it genuinely understood the failure mode. Thanks for the kind words too: GLM-5.3 Flash distributed across two Strix Halos is exactly what this PR is for, so reports like yours are the payoff.

MerlijnWajer:
I tried the latest commit that you pushed and it doesn't quite work for me yet. I get this on the coordinator:

```
ds4: warming sharded tensor pages (rank 0): 96.13 of 177.76 GiB
ds4: sharded warm done in 102.6s (checksum=3149873359)
ds4: ROCm preparing model tensor mappings: 14.49 GiB
ds4: ROCm startup model preparation covered 14.49 GiB of tensor spans in 0.002s
ds4: ROCm mapped 1.05 GiB auxiliary model
ds4: rocm backend initialized for graph diagnostics
ds4: memory: KV 2.79 GiB (raw 0.00 + compressed 2.79) + buffers 3.16 GiB + resident model 96.13 GiB = 1
02.08 GiB planned
ds4: memory detail: ctx=250000 prefill_cap=2048 raw_kv_rows=0 compressed_kv_rows=250000 backend=rocm
ds4-tp: waiting for worker on 0.0.0.0:8000 ...
ds4-tp: rdma device rocep3s0f0 (port state 4)
ds4-tp: rdma device limits: max_qp_wr 8192, max_sge 30, max_cqe 4194303, max_mr_size 184467440737095516
15
ds4-tp: rdma using RC queue pairs
ds4-tp: worker connected, transport=rdma gate-timeout=10000ms
0912 21:42:26 ds4-server: tp rdma: warm-up post_send: No space left on device
```

and this on the worker:

```
ds4: warming sharded tensor pages (rank 1): 96.13 of 177.76 GiB
ds4: sharded warm done in 61.6s (checksum=3148094405)
ds4: ROCm preparing model tensor mappings: 14.49 GiB
ds4: ROCm startup model preparation covered 14.49 GiB of tensor spans in 0.000s
ds4: ROCm mapped 1.05 GiB auxiliary model
ds4: rocm backend initialized for graph diagnostics
ds4: memory: KV 2.79 GiB (raw 0.00 + compressed 2.79) + buffers 3.16 GiB + resident model 96.13 GiB = 102.08 GiB planned
ds4: memory detail: ctx=250000 prefill_cap=2048 raw_kv_rows=0 compressed_kv_rows=250000 backend=rocm
ds4-tp: connecting to 192.168.100.1:8000 ... (Connection refused)
ds4-tp: connecting to 192.168.100.1:8000 ... (Connection refused)
ds4-tp: rdma device rocep3s0f0 (port state 4)
ds4-tp: rdma device limits: max_qp_wr 8192, max_sge 30, max_cqe 4194303, max_mr_size 18446744073709551615
ds4-tp: rdma using RC queue pairs
ds4-tp: leader connected, transport=rdma gate-timeout=10000ms
tp worker: tp rdma: warm-up post_send: No space left on device
```

I am not quite sure how to interpret 'No space left on device' here.

davidcanar:
Progress — the probe works (`rdma using RC queue pairs`), so RC creation and the modified QP transitions are fine on your ConnectX-4 Lx. The new failure is at the first warm-up `post_send`, but I don't yet trust the message itself: `ibv_post_send` returns the error code directly and does not reliably set `errno`, so our print was showing a possibly-stale `errno` — "No space left on device" (ENOMEM) may not be the real reason.

I pushed cb7cb70, which:
- captures the real return code from `ibv_post_send`/`ibv_post_recv` at the warm-up and big-gate sites and prints `strerror()` on that;
- logs the driver-adjusted QP caps after creation (`ds4-tp: rdma RC QP caps: max_send_wr=... max_recv_wr=...`) — for RC the driver can adjust the requested counts, and I want to see what it actually gave us vs. what your own working commit used.

Could you pull and re-run once (both ranks, and add `DS4_TP_CQ_DEBUG=1` to both so the hello exchange details are logged too)? With the real return code and the caps we should be able to pin this quickly. Two early hypotheses, depending on what the re-run says: if it turns out to be EINVAL, something about the RC transition differs from your commit; if it is genuinely ENOMEM, the RC send queue is being sized smaller than requested and I'll size the warm-up/window posts to the negotiated depth instead.

Frustrating that this is now a three-way handshake between your hardware, my code, and both of our assumptions — but we're close: the transport already negotiates RC, connects, and fails at a well-defined point with (soon) honest error codes.

MerlijnWajer:
>Could you pull and re-run once (both ranks, and add DS4_TP_CQ_DEBUG=1 to both so the hello exchange details are logged too)? With the real return code and the caps we should be able to pin this quickly. Two early hypotheses, depending on what the re-run says: if it turns out to be EINVAL, something about the RC transition differs from your commit; if it is genuinely ENOMEM, the RC send queue is being sized smaller than requested and I'll size the warm-up/window posts to the negotiated depth instead.

I will do that later today or tomorrow at latest, thanks for working on this.

calvinmorrow:
Brief note to say I got this working across 2x Strix Halos using Intel E810 w/ irdma driver

```
ds4: ROCm startup model preparation covered 96.13 GiB of tensor spans in 0.000s
ds4: ROCm mapped 1.05 GiB auxiliary model
ds4: rocm backend initialized for graph diagnostics
ds4: memory: KV 1.46 GiB (raw 0.00 + compressed 1.46) + buffers 3.16 GiB + resident model 96.13 GiB = 100.75 GiB planned
ds4: memory detail: ctx=131072 prefill_cap=2048 raw_kv_rows=0 compressed_kv_rows=131072 backend=rocm
ds4-tp: waiting for worker on 10.255.254.5:1234 ...
ds4-tp: rdma device rocep193s0f1, RoCEv2 GID 1, RC
ds4-tp: rdma device limits: max_qp_wr 8127, max_sge 6, max_cqe 1048574, max_mr_size 35184372088832
ds4-tp: rdma using RC queue pairs
ds4-tp: rdma RC QP caps: max_send_wr=1919 max_recv_wr=2047
ds4-tp: worker connected, transport=rdma gate-timeout=750ms
ds4-tp: rdma warm-up ok (1 attempt), queue depth recv 2047 send 1919
ds4: GLM TP: KDA attention of single-token decode split by heads (32 of 64 heads per rank)
tensor parallelism bound: rank 0, 50/50 expert split, rdma transport0915 00:17:34 ds4-server: context buffers 4736.53 MiB (ctx=131072, backend=rocm, prefill_chunk=2048, raw_kv_rows=0, compressed_kv_rows=131072)
ds4: GLM session ctx=131072 (model max=1048576); prefill/work cap=4096; dense attention limit=2051
ds4: GLM graph using compact DSA KV only; expanded full-attention KV cache is skipped
ds4: GLM graph allocating compact DSA cache: rows=131072 logical_ctx=131072 kv_layers=45 indexer_layers=11 f16 1.46 GiB
ds4: GLM compact indexed prefill chunk=2048 score_rows=2048 score_scratch=256.00 MiB
0915 00:17:35 ds4-server: KV disk cache /tmp/kv-cache (budget=32768 MiB, cross-quant=accept, min=512, cold_max=30000, continued=10000, trim=32, align=2048, hit_half_life=21600s)
0915 00:17:35 ds4-server: listening on http://0.0.0.0:8000
0915 00:19:42 ds4-server: chat ctx=0..1294:1294 prompt start
ds4: ROCm GLM causal indexed prefill using fp16 hipBLAS attention GEMMs (tokens=1294 rows=1294 cache=f16)
ds4: ROCm routed MoE prefill using MMQ Q4_K tensor-core GEMMs (n_tokens=1294, owned experts 144)
0915 00:19:54 ds4-server: chat ctx=0..1294:1294 prefill chunk 1294/1294 (100.0%) chunk=0.00 t/s avg=110.66 t/s 11.693s
0915 00:19:54 ds4-server: chat ctx=0..1294:1294 prompt done 11.702s
0915 00:19:54 ds4-server: kv cache stored tokens=1294 trimmed=0 reason=cold key=token-text size=175.74 MiB save=34.9 ms
ds4: ROCm GLM one-token value projection using wave-parallel Q8 rows
0915 00:19:57 ds4-server: chat ctx=1294..1344:50 gen=50 decoding chunk=13.36 t/s avg=13.36 t/s 3.743s
0915 00:20:01 ds4-server: chat ctx=1344..1394:50 gen=100 decoding chunk=13.34 t/s avg=13.35 t/s 7.490s
0915 00:20:05 ds4-server: chat ctx=1394..1444:50 gen=150 decoding chunk=13.37 t/s avg=13.36 t/s 11.230s
0915 00:20:09 ds4-server: chat ctx=1444..1494:50 gen=200 decoding chunk=13.35 t/s avg=13.36 t/s 14.974s
0915 00:20:09 ds4-server: chat ctx=1494..1499:5 gen=205 decoding chunk=13.30 t/s avg=13.36 t/s 15.350s
0915 00:20:09 ds4-server: thinking live checkpoint remembered ctx=0..1294:1294 live=1499 visible=6052
0915 00:20:09 ds4-server: chat ctx=0..1294:1294 gen=205 finish=stop 27.142s
```

MerlijnWajer:
I just pulled in the latest changes and things seem to work for me. So whatever changed since I last posted seems to have likely fixed the problem. Seeing some nice speeds too: ~180 t/s prefill and ~18-20 t/s gen.

MerlijnWajer:
Just wanted to report back and state that `GLM-5.3-UD-IQ2_XXS_RoutedIQ2XXS_blk78Q2K.gguf` (not the flash model) now also works on strix halo machines in pipeline mode with this PR - maybe not intentional, but I'm happy that it does. I'm getting about 30-40 t/s PP and 3-4 t/s TG split over four machines (not TP).

The TP=2 implementation for the particular model that I mentioned doesn't work because it's not Q4_K in some places:

>0923 16:46:02 ds4-server: tensor parallelism on ROCm requires Q4_K routed experts (layer 3); this build has ownership-aware kernels for Q4_K only

davidcanarnt:
> Just wanted to report back and state that `GLM-5.3-UD-IQ2_XXS_RoutedIQ2XXS_blk78Q2K.gguf` (not the flash model) now also works on strix halo machines in pipeline mode with this PR - maybe not intentional, but I'm happy that it does. I'm getting about 30-40 t/s PP and 3-4 t/s TG split over four machines (not TP). (EDIT: these numbers are from the model spread over four strix halos)
> 
> The TP=2 implementation for the particular model that I mentioned doesn't work because it's not Q4_K in some places:
> 
> > 0923 16:46:02 ds4-server: tensor parallelism on ROCm requires Q4_K routed experts (layer 3); this build has ownership-aware kernels for Q4_K only

Thank you for testing. I should be able to test this on 2 strix halos and see why TP=2 is not working. Unfortunately I won't be able to run it on more than 2 since I don't have them.

PR #1027, opened 2026-09-11
Make BPE merging O(n log n) for large CJK pieces

wbllcbbg-ai:
## Problem

The merge loop in `bpe_emit_piece` rescans every adjacent pair and shifts the symbol array on every merge, so it is quadratic in the number of symbols per piece. Punctuation-free CJK text forms letter-run pieces of hundreds of thousands of symbols under both pre-tokenizers, and tokenizing a ~1M-token Chinese prompt stalled for 34+ minutes (single thread, still running) — while the same scale in Italian tokenizes in about two minutes because whitespace keeps pieces tiny.

Sampled the stuck server with `sample`: nearly 100% in `bpe_tokenize_text -> bpe_emit_piece`, with the memmove-heavy array shifts dominating.

## Change

Fast path with identical merge semantics (always merge the leftmost pair with the minimal rank):

- symbols are `(start,end)` ranges into the encoded buffer — no per-symbol allocations, no concatenation copies
- doubly linked list instead of array shifts
- pair ranks cached and recomputed only for the two pairs touching each merge point
- min-heap ordered by `(rank, original symbol index)` with lazy invalidation, which reproduces the leftmost tie-break exactly

`DS4_BPE_SLOW=1` selects the original path, kept as reference and rollback.

## Verification

- Identical token counts vs the original path on seven text categories: punctuation-free Chinese, punctuated Chinese, English, Python, mixed zh/en, needle-marker lines, JSON blobs.
- Byte-identical greedy generation for the same prompt under both paths.
- The original failure case now completes: a 1,028,610-token Chinese context prefills normally with 5/5 needle retrieval (before the fix, tokenization alone never finished).
- A 12K-symbol single piece drops from 24.7s to 5.1s end to end; the gap widens quadratically beyond that.

Tested on M3 Ultra 512GB (Metal), GLM 5.3 Flash Q4_K and DeepSeek V4 Flash MXFP4.


Flor1an-B:
Same bug, same fix shape as #873 (open since 27/08, not yet reviewed) — worth cross-linking so review effort isn't duplicated.

#873 fixes the identical root cause in `bpe_emit_piece` (quadratic rescan-per-merge on long CJK/no-space pieces) with the same architecture: doubly-linked symbol list + min-heap of candidate merges ordered `(rank asc, left-position asc)` to reproduce the original "leftmost pair wins on tied rank" behavior exactly, staleness detected via a length snapshot at push time. Verified there via `--dump-tokens` byte-diff against the pre-fix binary across several cases (ASCII, a tie-break-heavy repeated-bigram stress case, empty/single-char, ~46 KB prose) — no repro at the 1M-token CJK scale this PR adds, which is a real gap #873 doesn't cover.

This PR's addition of an actual large-scale CJK repro (the 34+ min stall) and coverage under both pre-tokenizers looks like a genuine improvement over #873's test set either way. Happy to help reconcile the two into whichever one antirez prefers to review, if that's useful — just didn't want two independent implementations of the same fix to both sit waiting for review without either side knowing about the other.

wbllcbbg-ai:
Thanks for the detailed write-up @Flor1an-B — you're right, we independently arrived at the same diagnosis and essentially the same design: min-heap of candidate pairs over a doubly-linked symbol list, `(rank asc, leftmost asc)` ordering to reproduce the original scan's tie-break, lazy staleness checks at pop time. Nice to see the architecture validated twice independently.

I compared the two implementations side by side:

**Where #873 is stronger:**
- The tie-break stress case (`"ab"` × 800), `leaks --atExit` run, and clean `-Wall -Wextra -std=c99` build are checks this PR doesn't have.
- Smaller diff overall.

**Where #1027 adds value:**
- Symbols are zero-copy ranges into the already-allocated byte-encoded buffer — no per-symbol allocations and no realloc+memcpy per merge, which keeps allocator pressure flat on multi-MB single CJK pieces.
- The original implementation is kept as `bpe_emit_piece_slow`, selectable via `DS4_BPE_SLOW=1`, so the old path stays one env-var away for bisection/A-B.
- A 1M-token pure-CJK reproduction of the worst case (the motivating scenario: tokenization alone stalled 34+ min before, completes in seconds after) plus engine-level differential verification: token counts 7/7 match, greedy generation output byte-identical, 1M-context needle tests pass.

**Consolidation:** agreed that one PR is much easier for antirez to review. My suggestion — whoever's PR becomes the base, fold in the other's missing pieces:

- If #873 is the base: take the 1M-token CJK repro from here, consider the zero-copy range symbols, and optionally the `DS4_BPE_SLOW` escape hatch.
- If #1027 is the base: take the tie-break stress case, the leaks run, and the fixture regression checks from #873.

Happy to do the folding work in either direction — just say which base you prefer. (@antirez: your call; the two end states are functionally equivalent.)

PR #1030, opened 2026-09-12
Fix CUDA long-context smoke test linking

Matthley:
## Summary
- link `ds4_image.o` into `tests/cuda_long_context_smoke`
- resolve `ds4_deepseek4_attention_bounds` for `cuda-regression`

Requested by @GiorgioOppo in #952.

## Test plan
- [x] `make clean && make CUDA_ARCH=sm_121 cuda-regression`
- [x] `make -j4 cuda-spark`
- [x] DGX Spark GB10, CUDA 13.0, sm_121a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -626,7 +626,7 @@ ds4_rocm_compat.o: ds4_rocm_compat.cu ds4_gpu.h ds4_gpu_mgpu.h ds4_gpu_args.h ds
 ds4_rocm_unavailable.o: ds4_rocm_unavailable.cu
 	$(HIPCC) $(ROCM_CFLAGS) -c -o $@ ds4_rocm_unavailable.cu
 
-tests/cuda_long_context_smoke: tests/cuda_long_context_smoke.o ds4_cuda.o $(MMQ_OBJS)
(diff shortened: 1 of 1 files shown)


adamlawi:
Independent confirmation on a second DGX Spark (GB10 / sm_121, CUDA 13.0).

Reproduced the failure first, on the current head of #952 (`7a5002de`) without this patch:

```
/usr/bin/ld: ds4_cuda.o: in function `ds4_gpu_attention_visual_mixed_batch_heads_tensor':
ds4_cuda.cu:19444: undefined reference to `ds4_deepseek4_attention_bounds'
collect2: error: ld returned 1 exit status
make: *** [Makefile:622: tests/cuda_long_context_smoke] Error 1
```

With the one-line change applied, it links and the test passes:

```
ds4: CUDA backend initialized on NVIDIA GB10 (sm_121) dev=0
cuda-regression: top-k n_comp=32768 n_tokens=32 elapsed=0.000s
```

Worth noting for anyone else building on CUDA: `make cuda-spark` is unaffected — `ds4_image.o` is already in `CORE_OBJS`, so only the `tests/cuda_long_context_smoke` link line is short of it. That is presumably why this went unnoticed; we have built and shipped several CUDA production binaries from this tree without ever linking that target.

GiorgioOppo:
merged with some conflicts resolved

Issue #809, opened 2026-08-15
Support --ssd-streaming alongside --mtp / --dspark


lele872:
### Feature Description
Currently, attempting to run `ds4` with both `--ssd-streaming` and `--mtp` (or `--dspark`) results in the following operational limitation/error:
> `ds4: --ssd-streaming is not compatible with --mtp yet`

Support for combining `--ssd-streaming` with `--mtp`/DSpark speculative decoding would be extremely beneficial for hardware configurations with memory constraints (such as Apple Silicon devices with 64 GB Unified Memory).

### Motivation & Use Case
On systems like a Mac Studio M1 Ultra (64 GB RAM), running large quantizations requires SSD streaming to keep the memory footprint manageable and prevent OOM / system swapping. However, streaming experts from SSD naturally introduces latency during token generation.

Combining `--ssd-streaming` with DSpark / MTP speculative execution would offer the best of both worlds:
1. Low RAM usage via dynamic expert streaming from fast NVMe SSDs.
2. Improved tokens-per-second generation speed via speculative decoding.

### System Configuration
- **Hardware:** Mac Studio M1 Ultra (64 GB RAM)
- **Tooling:** `ds4` with DeepSeek / DSpark models

Thank you for all the incredible work on `ds4`!

Issue #724, opened 2026-08-06
DSpark proposals systematically wrong under --ssd-streaming (DeepSeek-V4-Flash-0731, Metal): enabling patch observed, miss_first=100%


johnpippett:
# DSpark (--mtp --dspark) produces systematically incorrect proposals under --ssd-streaming on DeepSeek-V4-Flash-0731 (Metal)

**Labels:** bug? / feature-gap — filed as **issue not PR** to let you decide whether
the enabling patch should land at all.

## Environment

- ds4 @ `b0309611041655f4e45671cfd9c9886aff161406` (== origin/main as of 2026-08-06)
- macOS Tahoe 27.0, Apple M5 Max, 64 GB RAM (internal 2 TB SSD)
- Model: `DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix-0731.gguf`
  (80.77 GiB, sha256 `ca22ae2f838e14077c22bc1c1417b71b45b5e5a3687bd96c2ac6e17fdb6261c0`)
- DSpark: `DeepSeek-V4-Flash-DSpark-support-0731.gguf` (5.58 GiB, sha256
  `7e319924541db3f7a163ed7e11d7532a70d48228ab59d36cb81e1d4511885360`)
- Backend: Metal. Config: `--ssd-streaming --ssd-streaming-cache-experts 32GB --ctx 65536`.

## Problem

`--ssd-streaming` + `--mtp` is refused at engine open (ds4.c:56508-56515:
"--ssd-streaming is not compatible with --mtp yet"). We wrote a minimal enabling
patch (see below) so we could measure what happens when they run together.

With our patch, everything is structurally correct (no crashes, no Metal view
errors, verifier available, greedy identity preserved), but **the drafter's
proposals are systematically wrong for this model+config**: with the scheduler and
confidence gating forced open, DSpark emits full-length draft chains every cycle
yet its *first* drafted token mismatches the target's greedy next-token 100% of
the time. Net effect of current behavior: speculation spends time and gains zero
(in our measurements ds4-server decode is ~20-35% slower with `--dspark`).

## Reproduced steps

No patch needed to observe "incompatible not supported yet" (stock refusal).
To observe the proposal misalignment, our PoC build is required:

[...]

PR #1031, opened 2026-09-12
CUDA: use selected expert cache for IQ2 SSD prefill

Matthley:
Fixes CUDA SSD streaming for routed MoE models.

1. IQ2/Q2 batch prefill now uses the compact selected-expert table rather than resolving full expert tables.
2. `--ssd-streaming-cache-experts` is a bounded CUDA LRU, released on model/lifecycle changes.
3. Resident experts are stored in contiguous gate/up/down slabs. When the selected set is resident, kernels receive the slab bases plus remapped stable slot IDs, avoiding per-layer GPU-to-GPU copies into compact staging.
4. CUDA now supports `--expert-profile` / `DS4_EXPERT_HOTLIST` to collect routed-expert locality data.

AProjQ4 matched prefill (GB10, e9528de, ctx 1024, forced batch prefill):
- Before: 1.09 t/s; CUDA file cache 78.62 GiB.
- Selected-expert path: 2.59 t/s; CUDA file cache 6.06 GiB.
- Result: 2.38x prefill (+138%) and 92.3% less file cache.

AProjQ8 cache-size tuning smoke (GB10, ctx 1024, deterministic prompt):

| Expert slots | Expert cache | Flash generation | Vision generation |
| ---: | ---: | ---: | ---: |
| 256 | 1.69 GiB | 1.62 t/s | 1.61 t/s |
| 512 | 3.38 GiB | 2.05 t/s | 3.04 t/s |
| 1024 | 6.76 GiB | 4.02 t/s | 4.17 t/s |
| 2048 | 13.52 GiB | 6.24 t/s | 6.00 t/s |

The two locally available Flash variants show the same capacity curve: 256 is below the routed working set, 1024 is the measured throughput/memory compromise, and 2048 favors generation throughput where the extra memory is available. Flash hit/miss at 256/512/1024/2048: 755/18,056; 5,027/13,784; 11,066/7,745; 13,809/5,002. Vision: 721/18,048; 8,707/10,062; 11,546/7,223; 13,922/4,847. These are single-run GB10/model-specific tuning observations, not a general performance claim.

A follow-up can expose a non-binding runtime hint based on `4 × (local routed layers × top-k)` when the memory planner permits it; this PR intentionally does not alter the user-selected cache budget.

Validation:
- `make -j4 cuda-spark`: PASS
- `make test-cuda-q8-scratch`: PASS
- Deterministic SSD-streaming smoke: PASS, identical output.


Hazelhof:
Independent reproduction on a 3× RTX 5080 (16 GB) host — the target hardware class this PR matters for.

**Base:** `8db1d1d` (2026-09-16). **Artifact:** the ds4-native
`DeepSeek-V4-Flash-IQ2XXS-w2Q2K-AProjQ8-SExpQ8-OutQ8-chat-v2-imatrix-0731.gguf` (80.76 GiB).
**Command (identical before/after):**
```
DS4_CUDA_NO_DIRECT_IO=1 DS4_CUDA_KEEP_MODEL_PAGES=1 \
  ds4-server -m <artifact> --cuda --ctx 16384 --ssd-streaming --prefill-chunk 512
```

**Before (stock 8db1d1d):** the model loads and serves short turns, and **every ≥10 k prompt dies**:
```
ds4: CUDA SSD cache cannot stage 107 experts with system headroom
ds4: gpu layer 0 ffn batch encode failed
ds4: gpu layer-major prefill layer 0 encode failed
```
The boundary is tight and reproducible: **93 experts (0.61 GiB) stages; 107 (0.72 GiB) does not.** Reading
`ds4_cuda.cu`, staging is bounded by `available = cudaMemGetInfo().free − 8 GiB` against
`unique_experts × 6.75 MiB`; on a 16 GB card that leaves ~8.0 GiB free at staging time, so `available ≈ 0`.
`--ssd-streaming-cache-experts 8GB` prints a plan that would fit
(`8.00 GiB = 3.38 GiB prefill headroom + 4.62 GiB dynamic cache (701 experts, 6.75 MiB each)`) and **still
fails** — the cache hint is not the binding term. `--prefill-chunk 512` alone gets 93 experts staged.

**After (this PR, built for sm_120):** **the same 12,422-token prompt completes at 113.45 t/s**, and a
66-token chat turn goes from 1.34 → 3.15 t/s. G1/G2/G3 all pass. Nothing else about the run changed except
the binary.

Two things that may be worth a line in the PR text rather than the code:

1. **`--ssd-streaming` + `--mtp-model`/`--dspark` is refused upstream** (issue #809), and on small-VRAM cards
   streaming is the *only* way this model loads at all — the resident layout is refused
   (`CUDA model arena alloc failed for tensor-span:26`, 80.76 GiB vs 48 GB). So speculation is unreachable
   exactly where a drafter would help most. (Issue #724's enabling patch measured the combined path as
   `miss_first=100 %`, so the guard looks correct today; this is only about how much this PR carries for
   consumer cards.)
2. **The four commits do not cherry-pick onto current `main`**: applying them to `8db1d1d` conflicts in
   `ds4_cuda.cu` (the PR was written against base `371cb40100`, and main moved five days of CUDA work under
   it). We therefore build the branch head itself — worth knowing for anyone trying to rebase before merge.

Also on this build, for what it is worth: the explicit expert cache works at every size we tried
(256 / 512 / 1024 / 2048 slots — all four complete a 12,422-token prefill). Our curve is **not** monotonic
(512 measured faster than 1024 on single runs), unlike the GB10 table in the PR text, so it may be
workload- or staging-dependent rather than a clean capacity curve.

PR #1033, opened 2026-09-13
metal: add opt-in slab residency for DeepSeek V4.1 Flash （M2 192GB 0.25tk/s -> 13tk/s)

Dango233:
This attaches owned SSD expert-cache slabs to a queue residency set on macOS 15+. The switch is off by default. Memory-pressure relief removes the set until the cache is rebuilt. This addresses the multi-second command submission delay reproduced with large slabs on this M2 Ultra.

Enable with `DS4_METAL_STREAMING_SLAB_RESIDENCY=1`.

Mean 8-token decode: **0.240 → 9.310 tokens/s** (38.8×).

| --- | --- | ---: | ---: |
| 1 | off | 72.13 | 0.24 |
| 2 | on | 67.99 | 9.35 |
| 3 | on | 67.81 | 9.27 |
| 4 | off | 61.97 | 0.24 |

All four decoded outputs match. These are two observations per variant on one host. The short decode includes the first-token cost; candidate steady decode is 13.41, 13.31 tokens/s. Original 2026-09-13 measurements remain in the report.

### Validation and limits

Slab lifecycle and pressure-relief tests, SSD expert kernels, MoE prefill, admission/address checks, and the real-model session fixture pass under Metal API validation.

[Commands, model hash, raw CSVs and full results](https://github.com/Dango233/ds4/blob/perf/metal-streaming-slab-residency/speed-bench/v41_slab_residency_m2_ultra.md#upstream-revalidation-2026-09-14).


ds4_metal.m
```diff
@@ -13896,6 +13915,36 @@ static uint64_t ds4_gpu_stream_expert_slab_target_bytes(void) {
     }
     buffer.label = label;
     g_stream_expert_cache_buffer_allocs++;
+    if (getenv("DS4_METAL_STREAMING_SLAB_RESIDENCY") &&
+        !g_stream_expert_cache_mlock_relief_applied) {
+#if TARGET_OS_OSX
+        if (@available(macOS 15.0, *)) {
+            const BOOL fresh = g_stream_slab_residency_set == nil;
```
emilianbold, line 13922: I find the `fresh` bool logic hard to read. It's basically an if/else on `g_stream_slab_residency_set == nil` with the `else` just adding the new buffer to the existing residency set.

ds4_metal.m
```diff
@@ -13896,6 +13915,36 @@ static uint64_t ds4_gpu_stream_expert_slab_target_bytes(void) {
...
+            if (fresh) {
+                MTLResidencySetDescriptor *desc = [[MTLResidencySetDescriptor alloc] init];
+                desc.label = @"ds4_streaming_expert_slabs";
+                desc.initialCapacity = 64;
+                NSError *error = nil;
+                g_stream_slab_residency_set = [g_device newResidencySetWithDescriptor:desc error:&error];
+                if (!g_stream_slab_residency_set) {
+                    fprintf(stderr, "ds4: streaming slab residency set failed: %s\n",
+                            [[error localizedDescription] UTF8String]);
+                    return nil;
```
emilianbold, line 13932: Somewhat debatable to me why can't we just return the usable `buffer` here.

emilianbold (commented):
I like how small the actual change is. Some minor comments below.

PR #1035, opened 2026-09-13
engram: parallelize DeepSeek V4.1 Flash decode reads on macOS

Dango233:
This partitions macOS single-token Engram reads into four joined readers, preserving input order and disjoint output ranges. The switch is off by default. Inputs are validated before dispatch and worker errors propagate after joining. Upstream now parallelizes large batches on non-macOS platforms; this PR still covers the separate macOS 24-row decode path and preserves the new upstream path.

Enable with `DS4_ENGRAM_PARALLEL_DECODE=1`.

Both arms use `DS4_METAL_DISABLE_STREAMING_EXPERT_SLABS=1` to isolate this change. No companion optimization is enabled.

Mean 512-token decode: **10.205 → 10.740 tokens/s** (+5.2%).

| --- | --- | ---: | ---: |
| 1 | off | 68.43 | 10.20 |
| 2 | on | 68.65 | 10.72 |
| 3 | on | 68.63 | 10.76 |
| 4 | off | 68.66 | 10.21 |

All four decoded outputs match. These are two observations per variant on one host. Original 2026-09-13 measurements remain in the report.

### Validation and limits

Serial/parallel row equality, duplicate IDs, output guards, invalid inputs, EDOM in each worker, EIO, and upstream batch error checks pass. The real-model session fixture passes under Metal API validation with parallel reads enabled.

[Commands, model hash, raw CSVs and full results](https://github.com/Dango233/ds4/blob/perf/v41-parallel-engram-read/speed-bench/v41_engram_m2_ultra.md#upstream-revalidation-2026-09-14).


adriangalilea:
Resident data point (the PR measures SSD streaming): Mac Studio M3 Ultra 512 GB, Q4 GGUF fully resident, 2k prompt, greedy 256, `DS4_ENGRAM_PARALLEL_DECODE=1` vs off, ABBA on the same binary. Base already has the single-host layer queue from #1034 (extended to resident mode, see my comment there).

| | decode t/s |
|---|---:|
| off | 21.63 / 21.63 |
| on | 23.24 / 23.23 (**+7.4 %**) |

Generated text byte-identical in all four runs. It composes with committing each queued layer without waiting (another +6.7 % on its own): both together 24.86 / 24.85 t/s, i.e. +15 % over the queued base and +48 % over upstream's per-layer-drain default on this box (16.8 t/s this morning). Per-token GPU stage timestamps put the remaining CPU-side time outside the GPU span at ~5 ms before this change; the Engram reads are exactly that gap, so the +7 % matches. Worth defaulting on macOS in my view — the resident path has no other reader to contend with.

Dango233:
Thanks for the resident Q4 measurements and for testing this with the queue changes. That complements our M2 Ultra Q2 SSD result (+5.2% in the September 14 ABBA run).

I have merged current upstream `9139e2a`, resolved the benchmark README conflict, and rerun clean Metal/CPU builds, unit tests and the real-model session fixture under Metal API validation (`ce0c537`). The implementation remains independent of the queue PR. The recorded throughput numbers still refer to the September 14 base; this update is integration validation.

I am keeping the macOS switch opt-in for now. Your resident data makes default-on worth discussing, but I would like to retain the SSD/concurrent-reader case in that decision as well. I have also followed up on #1041 so we can converge on one queue implementation and contribute our forced-eviction/full-logit checks there.

ds4_engram.c
```diff
@@ -157,6 +178,23 @@ bool ds4_engram_read(const ds4_engram_table *t, const uint32_t *rows,
...
         }
     }
+#ifdef __APPLE__
+    if (count == DS4_ENGRAM_COLS && getenv("DS4_ENGRAM_PARALLEL_DECODE")) {
+        /* Decode reads 24 uncached rows. Partition the original order without
+         * sorting or allocating; each reader owns disjoint output rows and
+         * dispatch_apply joins all readers before returning to the caller. */
+        engram_decode_read read = {.table = t, .rows = rows, .out = out};
+        dispatch_apply_f(ENGRAM_DECODE_READERS,
+            dispatch_get_global_queue(QOS_CLASS_USER_INITIATED, 0), &read, read_decode_part);
```
emilianbold, line 188: I like how small this PR is (ignoring the benchmark) but the other block where we read in parallel (`read_batch_part`) we do it with threads on non-Apple systems.

This PR somehow breaks symmetry: non-Apple systems read sequentially.

It's still seems good based on the numbers measured, just somewhat odd.

PR #1036, opened 2026-09-13
ROCm: add DeepSeek V4.1 Flash support for Strix Halo (gfx1151)

kyuz0:
Adds **DeepSeek V4.1 Flash Q2 text and vision on ROCm 10.0 / Strix Halo `gfx1151`**: resident experts, single-machine SSD streaming, and two-machine inference over TCP or RoCE.

- V4.1 operators, CED prefill, disk-backed Engram and session state.
- FP32 split-key decode attention, optimized quantized projections, matrix-based prefill and exact batched top-512 selection.
- Two-machine inference splits the experts and vocabulary projection; shared-expert prefill overlaps peer transfers, and scalar gate waiting/reduction is fused.
- V4.1 ROCm optimizations use guarded dispatch with no required tuning environment overrides. Ordinary TCP over USB4 Ethernet uses the same TCP commands.

### Performance

Native `ds4-bench`, calibrated Q2, fresh text prefixes, 69,632 allocated context, greedy output, no DSpark or images. Values are **prefill / decode tokens/s**; one run per cell. Startup and readiness warmup are excluded.

| Prompt tokens | Generated tokens | SSD | TCP, 100 GbE | RoCE, 100 GbE |
|---:|---:|---:|---:|---:|
| 1,024 | 128 | — | 123.54 / 15.42 | 123.80 / 15.94 |
| 16,384 | 512 | 301.23 / 9.73 | 412.94 / 15.59 | 410.54 / 15.94 |
| 65,536 | 128 | 350.03 / 9.08 | 432.83 / 14.96 | 431.85 / 15.33 |

- Second RoCE 16K/512 run: **408.25 / 15.92**. Full frontier logits and continuations match between TCP and RoCE at each depth; SSD matches corresponding resident runs.
- Hardware: 128 GB Framework Desktops, 16-core Strix Halo. Coordinator: Ryzen AI Max+ 395, SK hynix PC711 1 TB (PCIe 3.0 ×4, ext4). Worker/SSD node: engineering sample `100-000001243-50_Y`, Kingston FURY Renegade 2 TB (`SFYRD2000G`, PCIe 4.0 ×4, btrfs); SSD cache/staging budget 92 GiB.
- Intel E810-C QSFP, 100 Gb/s, MTU 9000; NIC attachment PCIe 3.0 ×4 on coordinator and 4.0 ×4 on worker. RDMA device logs and hardware counters verify RoCE payload traffic.
- Linux 7.2.5, ROCm SDK `10.0.0-4` / HIP `7.15.26333`, TuneD **`accelerator-performance`**, maximum fans. No OOM; some qualification runs recorded zram swap-out. These are not cold-cache measurements.

### Quality and limits

- Official scoring: **115 cases / 3,954 targets** per resident and two-machine panel. Probability drift is mixed, including worse selected long cases; full NLL/MAE/reference tables are in QA.
- Per mode: **44 practical checks**, **69 state comparisons**, and **six image cases** pass; focused SSD image checks also pass. Coding, tool-use, cancellation, rewind and snapshot recovery are covered. Repetitive raw greedy benchmark continuations remain a caveat; practical passes do not establish distribution equivalence.
- Production-shape kernel/reference checks, frontend build/help checks, focused ROCm transport tests, RDMA completion tests and server/agent unit checks pass. No fresh physical CUDA/Metal/GLM regression run is claimed.
- Exactly two machines, each fitting its assigned experts in RAM. Cluster SSD streaming and DSpark are unsupported. RoCE needs compatible adapters/providers and uses system-RAM staging.
- Measured text contexts reach 64K. Populated 256K quality and long-running production endurance are unqualified; image workloads may need a smaller SSD cache.

Setup, exact commands and detailed results: [single-node SSD](https://github.com/kyuz0/ds4/blob/40b1f63afa3a028bed23678b013f7895140b4f02/docs/STRIX_HALO.md#deepseek-v41-flash), [TCP/RoCE cluster](https://github.com/kyuz0/ds4/blob/40b1f63afa3a028bed23678b013f7895140b4f02/docs/CLUSTERING_ROCM.md), [QA](https://github.com/kyuz0/ds4/blob/40b1f63afa3a028bed23678b013f7895140b4f02/QA_BEFORE_RELEASES.md#deepseek-v41-flash-rocmgfx1151).

### Update — 16 September 2026

- Removed Linux USB4STREAM support and its setup/kernel-patch documentation. On the tested AMD systems it required a custom kernel patch and offered no meaningful performance benefit over the retained TCP/RoCE options.
- A shared Linux TCP stress test (`tests/test_tp_tcp`) fails with tiny socket buffers. Found during validation after the rebase, it also fails on unchanged `main`; the TCP functions it exercises are unchanged by this PR. This is a separate upstream issue and remains unresolved. ROCm transport checks pass.

### Update — 17 September 2026

Merged upstream main and retested ROCm support.

| Check | Result |
|---|---|
| Official scoring | All 115 cases / 3,954 targets retain the same scores after the merge. |
| Resident and SSD serving | Completed tests with two or four concurrent requests, cancellation, session reuse and recovery. |
| TCP and RoCE | 8K and 16K tests retained identical generated output and showed similar throughput after the merge. |
| Vision | All 23 image and cache-reuse cases passed in each of resident, SSD, TCP and RoCE modes. |

I found an unresolved issue during concurrent-request testing: identical requests can return different replies, including with greedy decoding. I also reproduced this on unmodified upstream main using CUDA on NVIDIA GB10 with Q2 SSD streaming, so it is not specific to ROCm. This update does not fix that issue.


GiorgioOppo:
@kyuz0 Check out the optimizations made in #952. I think they can greatly speed up prefill, especially the use of FP16. If you have time, take a look and let me know.

Piega:
Independent verification on different hardware — works, thanks for this port.

**Hardware**: Bosgame M5 — Ryzen AI Max+ 395 (Radeon 8060S, gfx1151), 128 GB LPDDR5x, 2 TB YMTC PC41Q-2TB-B NVMe (PCIe 4.0), stock kernel boot params (no `amdgpu.gttsize` / `ttm.pages_limit` overrides).
**Software**: `rocm/dev-ubuntu-26.04:10.0.0-full` (ROCm 10.0), built from this branch with `make strix-halo ROCM_ARCH=gfx1151` — clean build, no modifications.
**Model**: `ds41f-q2` (340.6 GiB file), Engram disk-backed.

Command (following docs/STRIX_HALO.md, idle machine):

```sh
./ds4-bench --prompt-file long.txt --ctx-max 8192 --gen-tokens 64 --rocm \
  --ssd-streaming --ssd-streaming-cache-experts 92GB \
  -m gguf/DeepSeek-V4.1-Flash-Q2.gguf
```

Results:

| ctx | prefill tok/s (new interval) | steady gen tok/s |
|---:|---:|---:|
| 2048 | 48.77 | 5.11 |
| 4096 | 60.47 | 6.02 |
| 6144 | 60.61 | 5.96 |
| 8192 | 60.83 | 5.76 |

Memory plan reported at startup: 9.37 GiB static weights + 7.12 GiB prefill expert staging + **41.13 GiB dynamic expert cache** + 1.66 GiB buffers + KV 0.06 GiB = 59.34 GiB planned; Engram disk-only. No OOM, no swap, single run (not repeated, so treat as a data point rather than a stable average).

Observations:

1. **Decode matches yours**: 5.76-6.02 tok/s vs your 5.47 tok/s — consistent with a streaming/SSD-bound decode path, which is reassuring across two different machines.
2. **Prefill is lower here**: ~61 tok/s at 8K vs your 214 tok/s. Two contributing factors on this box: direct model-file read measured **2.4 GB/s** (vs your 5.16 GB/s verified on a Kingston FURY PCIe 4.0), and the expert cache that was actually admitted (41.13 GiB) is well below the 89-94 GiB on your test system.
3. Question on the cache budget: `--ssd-streaming-cache-experts 92GB` was passed while ~120 GB of RAM were free (the other local model service was stopped for the run), but autosizing still planned 41.13 GiB. Is that expected admission behavior with stock GTT/TTM settings (we have no boot-param overrides), or should the explicit 92GB have been honored/warned about? Worth documenting the interaction between the explicit flag and automatic sizing if it can be silently reduced.
4. Minor doc inconsistency: `./download_model.sh --help` still describes `ds41f-q2` as "Metal only: use SSD streaming on one 128 GB Mac, tensor parallelism on two, or a larger resident Mac", while docs/MODELS.md and docs/STRIX_HALO.md in this same PR document ROCm support on gfx1151. Might be worth updating the help text.

For reference, on this same machine DeepSeek V4 Flash Q2 (0731) + DSpark resident runs at ~273-284 tok/s prefill and ~24 tok/s decode, so V4.1 is clearly the quality/vision option rather than the interactive one — consistent with the PR description.

kyuz0:
Thank you for testing so quickly! I'll need to check out what might be causing such a slow down. Right now I am working on the distributed implementation, once I've got that working I'll come back to testing this.

Piega:
Follow-up on my own cache-admission question above, after re-reading docs/STRIX_HALO.md: this machine runs **stock kernel boot parameters** (no `amdgpu.gttsize=126976 ttm.pages_limit=32505856 ttm.page_pool_size=32505856`), so the GPU-visible pool here is the small one (~62 GiB) that the doc warns about with "Some 128 GB configurations expose only about 62 GB to the GPU". That is very likely why automatic sizing admitted 41.13 GiB of expert cache instead of honouring the 92GB request — the cache budget follows the GPU-visible/GTT headroom, not just free RAM.

So my point 3 may not need a code change at all: it may just be the documented prerequisite. Worth stating explicitly in docs/STRIX_HALO.md (or in the flag's help text) that the expert-cache budget scales with the GTT configuration, so a 92GB request on a stock-parameter host will be silently reduced to what the pool allows. If that is the intended behaviour, a one-line warning at startup when the requested cache is clamped would remove the guesswork.

kyuz0:
> Follow-up on my own cache-admission question above, after re-reading docs/STRIX_HALO.md: this machine runs **stock kernel boot parameters** (no `amdgpu.gttsize=126976 ttm.pages_limit=32505856 ttm.page_pool_size=32505856`), so the GPU-visible pool here is the small one (~62 GiB) that the doc warns about with "Some 128 GB configurations expose only about 62 GB to the GPU". That is very likely why automatic sizing admitted 41.13 GiB of expert cache instead of honouring the 92GB request — the cache budget follows the GPU-visible/GTT headroom, not just free RAM.
> 
> So my point 3 may not need a code change at all: it may just be the documented prerequisite. Worth stating explicitly in docs/STRIX_HALO.md (or in the flag's help text) that the expert-cache budget scales with the GTT configuration, so a 92GB request on a stock-parameter host will be silently reduced to what the pool allows. If that is the intended behaviour, a one-line warning at startup when the requested cache is clamped would remove the guesswork.

I was exactly going to ask that, I always assume people set up their Strix Halo to use unified memory, as it's one of the strong points of the APU:

https://strix-halo-toolboxes.com/#config

Any reason why you don't set it up like that?

kyuz0:
There’s also a benchmark-shape difference: your 8K row times a 2K append, while mine times a fresh 8K prefill. If you’re willing to isolate that difference, could you try:

```bash
./ds4-bench -m gguf/DeepSeek-V4.1-Flash-Q2.gguf --rocm \
  --ssd-streaming --ssd-streaming-cache-experts 89GB \
  --ctx-start 8192 --ctx-max 8192 --ctx-alloc 10240 \
  --gen-tokens 16
```

I’d still expect the cache request to be reduced on your current setup. Could you share the resulting timing and the startup cache-admission lines? I’ll make the GTT-dependent limit clearer in the documentation and check whether the existing “cache fitted” diagnostic clearly reports the requested and effective GiB.

Piega:
@kyuz0 thanks — good catch on the workload shape, here is the exact run you asked for.

```bash
./ds4-bench -m gguf/DeepSeek-V4.1-Flash-Q2.gguf --rocm \
  --ssd-streaming --ssd-streaming-cache-experts 89GB \
  --ctx-start 8192 --ctx-max 8192 --ctx-alloc 10240 --gen-tokens 16
```

CSV row: `8192,8192,156.95,16,3.56,403.605,15,3.67,0`

Startup cache-admission lines (with the 89GB request):

```text
ds4: rocm SSD streaming cache budget 89.00 GiB capped to 51.00 GiB to stay below the graph working-set pressure budget
ds4: V4.1 SSD expert storage 50.99 GiB = 7.12 GiB two-layer staging + 43.87 GiB dynamic (4733 experts)
ds4: V4.1 SSD cache fitted from 4733 to 4427 experts for context/runtime headroom
ds4: V4.1 SSD prepared 9.37 GiB static weights and 7.12 GiB expert staging; 41.04 GiB dynamic cache, 1.82 GiB graph, 2.00 GiB runtime/I/O allowance
ds4: memory: KV 0.07 GiB + buffers 1.75 GiB + resident model 9.37 GiB + expert cache 41.04 GiB + prefill expert reserve 7.12 GiB = 59.34 GiB planned
```

So a fresh 8K prefill here is **156.95 t/s** vs your 214 t/s, and the diagnostic accounts for most of it: the effective cache is capped at 51 GiB and fits 4,427 of 4,733 experts, so prefill streams more. My row only generates 16 tokens, so the 3.56/3.67 decode figure is not comparable to your 5.47 — I am not reading anything into it.

On the unified-memory setup: no strong reason, just inertia. Early on I tested Unsloth Studio on this box before touching any firmware settings, and without that configuration it did not see the full RAM available to the GPU, so I kept the stock setup for the production llama.cpp service and never came back to it. You are right that unified memory is the whole point of the APU — I will set it up as in https://strix-halo-toolboxes.com/#config when I get a moment, then re-run this same command and post the before/after here, so you get the GTT comparison and we can see whether the cache cap moves from 51 GiB towards the 89 GiB request.

Two small suggestions while we are here: the "capped to X GiB to stay below the graph working-set pressure budget" line is already very useful — adding requested-vs-effective and which budget bound it would remove the last bit of guesswork. And `./download_model.sh --help` still describes `ds41f-q2` as "Metal only", while this PR documents ROCm on gfx1151.

kyuz0:
I've pushed some updates for `./download_model.sh`  and better diagnostics.

jyatesdotdev:
I'm going to try this out in tandem with #861.

Piega:
Before/after, as promised — the box is now configured with unified memory the way the toolboxes recommend, and it confirms the cache-admission hypothesis.

**What changed**: BIOS `UMA Frame Buffer Size` = **512 MB** (it was previously a larger carve-out) plus the documented kernel parameters `amdgpu.gttsize=126976 ttm.pages_limit=32505856` (no `amd_iommu=off`). Result:

```text
MemTotal        128,650,972 kB  (122.7 GiB, full system RAM back)
mem_info_gtt_total  124 GiB
mem_info_vram_total  ~0 GiB     (512 MB carve-out)
```

Worth noting for anyone following this thread: setting the UMA frame buffer **large** (I first tried ~96 GB) is the wrong direction — `MemTotal` collapsed to 30.5 GiB and everything starved. Small carve-out + large GTT is what works.

Same command you asked for, before (stock parameters) vs after:

```bash
./ds4-bench -m gguf/DeepSeek-V4.1-Flash-Q2.gguf --rocm \
  --ssd-streaming --ssd-streaming-cache-experts 89GB \
  --ctx-start 8192 --ctx-max 8192 --ctx-alloc 10240 --gen-tokens 16
```

| | before (stock GTT) | after (124 GiB GTT) |
|---|---|---|
| cache budget | 89 GiB → **capped to 51.00 GiB** | **81.88 GiB dynamic** (no clamp message) |
| experts fitted | 4,427 / 4,733 | (fits the 81.88 GiB cache) |
| planned memory | 59.34 GiB | 100.19 GiB |
| fresh 8K prefill | 156.95 t/s | **160.37 t/s** |
| row | `8192,8192,156.95,16,3.56,…` | `8192,8192,160.37,16,5.27,202.724,15,5.29,0` |

So my earlier question 3 was indeed just the documented prerequisite, not a code problem: the cache budget follows the GPU-visible headroom. The startup line is already informative — a "requested X, effective Y" wording would close the loop, but nothing is broken.

Prefill is still below your 214 t/s, and my SSD is the likely reason (2.4 GB/s direct read vs your 5.16 GB/s), now with the same class of cache budget. If you want, I can re-run once more with a warmer page cache for the model file to separate SSD bandwidth from cache effects.

Two side observations from the same session, both with the new pool, for reference:
- DeepSeek V4 Flash Q2 (0731) resident on unmodified `main` (no host-memory arena patch needed anymore): prefill 287-297 t/s, steady gen 14.0-14.8 t/s (vs 273-284 / 13.2-13.7 with the small pool plus that patch).
- GLM 5.3 Flash Q2 resident on #1011: ctx 4096 prefill 130.4 t/s, gen 10.9 t/s (vs 118.7 / 10.2 before).

kyuz0:
Thank you. What I am doing now is rebasing on main, as CUDA support has been added, and I will try a few different kernels and solutions specifically to improve decode on this PR - I'll ping you when i think this can do with another test.

saxs2224:
Tested this independently on my own Strix Halo box (AMD Ryzen AI Max+ 395 / Radeon 8060S, gfx1151, 128GB unified memory, ROCm 10.0 / HIP 7.15) — `DeepSeek-V4.1-Flash-Q2.gguf` via `./download_model.sh ds41f-q2` (365.71GB, checksum matched), `--ssd-streaming --ssd-streaming-cache-experts 92GB`.

**Build**: clean, no errors.

**Speed** (`ds4-bench --gen-tokens 0`, fresh prefill, `speed-bench/promessi_sposi.txt`):

| ctx | prefill tok/s (3 runs) |
|---:|---:|
| 16384 | 239.66 / 240.97 / 241.63 |

Matches your reported 234-259 tok/s range closely.

**Decode**: quick informal check via a single `ds4-server` HTTP request (default power profile, no `tuned-adm accelerator-performance`, not your clean `ds4-bench` methodology) — 300 completion tokens in 55.8s ≈ **5.4 tok/s**. Lower than your reported 8.8-14.7 tok/s; likely explained by the different measurement setup (power profile, single HTTP request vs isolated bench) rather than a functional issue, but flagging the gap in case it's useful data — haven't dug further into it.

**Correctness**:

- `make test-deepseek41-rocm`: **PASS, 118 isolated shapes**, 0 numerical mismatches.
- Determinism: reran the same prefill 3x at ctx=16384 with `--dump-frontier-logits-dir`, diffed every pairing — **0/129292 differing logits every time**.
- End-to-end: sent the same prompt through `ds4-server` twice (greedy, two different max_tokens budgets, one truncated mid-reasoning and one completing normally) — **byte-identical output both times** in both cases.

No correctness issues found. Nice work getting V4.1 onto ROCm this fast after the CUDA/Metal ports landed.

methuselah-0:
I made another independent test and can confirm it works on dual strix halo over TCP (usb4v2 tuned to 9microsecond latency) at a915971, rocm10 Ubuntu 26.04 docker image. Scored a whopping 94% at pi-bench. Decode stable around 13t/s, prefill around 110-150t/s based on ds4 log output. I had one connection timeout issue causing "tp worker sync: tp: prefill synchronization failure" occur when pi-bench was invoking the judge at one test, it happened during a 40MiB sync, but that's all.

And thanks, amazing to be able to run this locally!

kyuz0:
I have pushed some performance updates and removed USB4STREAM support, as there was no performance benefit and it required kernel patches.

If people want to test, have a go!

Right now I am putting this through Terminal Bench Mini and comparing the initial 4 tasks against the V4 Flash run:

image

What's notable is that V4.1 uses way less tokens for these 4 tasks, so slower inference but at least on this first panel a more efficient use of tokens.

Piega:
Benchmarked the updated branch on the same box as our earlier run (Bosgame M5, Ryzen AI Max+ 395, gfx1151, ROCm 10.0, stock kernel boot params, PR #1028 host-RAM merged on top).

Model: DeepSeek-V4.1-Flash-Q2, `--ssd-streaming --ssd-streaming-cache-experts 60GB` (auto-reduced from 92GB: the production llama server shares RAM), `--chat-prompt-file`, greedy 64 tokens per frontier, ctx 2048→8192.

A/B vs the Sep 13 build of this PR, identical settings:

| ctx | prefill tok/s (old→new) | gen tok/s (old→new) |
|---|---|---|
| 2048 | 48.9 → 49.1 | 6.0 → 10.0 |
| 4096 | 72.3 → 72.4 | 6.2 → 10.4 |
| 6144 | 71.8 → 70.5 | 6.1 → 10.5 |
| 8192 | 70.9 → 71.7 | 6.1 → 10.5 |

Decode +72% and first-token ~40% faster (209→98 ms at 2048). Prefill unchanged, consistent with being SSD-streaming bound. Happy to help with Terminal Bench Mini if useful.

jyatesdotdev:
I have this running on my two nodes; awesome! Thanks for this!

kyuz0:
Identical, seeded V4.1 requests can return different replies when run concurrently. I reproduced this on **unmodified upstream `main` with CUDA**, independently of the ROCm implementation in this PR.

I checked `antirez/ds4` at `8db1d1d155cb0400a86a86b9c62d0defb3a6148b` (the current upstream head when tested), built for CUDA `sm_121`, on NVIDIA GB10 with CUDA 13.0 and driver 580.173.02. Model: `DeepSeek-V4.1-Flash-Q2.gguf`, SSD streaming with a 76 GiB expert-cache request, two resident server sessions, 8K context and disk KV checkpoints.

Both complete upstream batching suites returned all **16 responses / 8 pairs** each. With the default scheduler, JSON reported mismatches in `medium-sampled`, `long-greedy` and `very-long-sampled`; SSE reported mismatches in `medium-sampled`, `very-long-sampled` and `long-sampled`, and completed all four cancellation checks. These were duplicate-output assertion failures, not timeouts. In particular, the JSON result includes a greedy mismatch.

Repeating both complete suites with the existing `--mixed-prefill-quantum 2048` option passed: all 32 responses and four cancellation checks completed without duplicate-output mismatches. Both server processes exited normally. The engine was built from the unmodified upstream archive; the test client had only an output-capture addition, with upstream fixtures and assertions unchanged.

Reproduction:

```sh
./ds4-server --cuda -m /path/to/DeepSeek-V4.1-Flash-Q2.gguf \
  --ctx 8192 --batched-session 2 \
  --ssd-streaming --ssd-streaming-cache-experts 76GB \
  --kv-disk-dir ./v41-repro-kv --kv-disk-space-mb 256 \
  --host 127.0.0.1 --port 19560

python3 tests/test_server_batching.py \
  --url http://127.0.0.1:19560 --pairs 8 --workers 4 \
  --nonce ds41-concurrency-full-json --timeout 1800

python3 tests/test_server_batching.py \
  --url http://127.0.0.1:19560 --pairs 8 --workers 4 \
  --nonce ds41-concurrency-full-stream --timeout 1800 \
  --stream --cancel-first 4
```

For the control, restart the same build with a separate empty disk-KV directory and add `--mixed-prefill-quantum 2048`, then repeat both client commands.

ROCm also exhibits duplicate-request mismatches. An independent one-session ROCm control changes logits and greedy continuations when changing only the prefill slice size from 2048 to 128; repeated 2048 runs match. That shows chunk geometry is sufficient to introduce a difference in that control, without concurrent clients or checkpoint restore. It does not establish the complete root cause of the CUDA result, a KV-cache race, or a general model-quality regression.

The HTTP scheduler switches between 2048-token idle and 128-token busy prefills. CUDA and ROCm have different internal short-prefill implementations, but the observed issue crosses backends and also occurs without this PR. It needs investigation as a shared serving/numerical-repeatability issue. Metal has not been tested here.

This PR keeps upstream scheduling defaults. The existing `--mixed-prefill-quantum 2048` option is a diagnostic control/workaround, not a root-cause fix; larger slices can substantially delay an already-generating response. It is not being made the ROCm default.

jhohertz:
Does this work for clusters >2? I saw there was no --layers flags and wondering if it auto-configures or is wired for 2 only.

kyuz0:
> Does this work for clusters >2? I saw there was no --layers flags and wondering if it auto-configures or is wired for 2 only.

Only 2 at the moment, I am not sure if it's worth trying with more TBH.

davidcanar:
> > Does this work for clusters >2? I saw there was no --layers flags and wondering if it auto-configures or is wired for 2 only.
> 
> Only 2 at the moment, I am not sure if it's worth trying with more TBH.

I always had in mind that Q2's quality was not good but Q4's is. I don't know if that is still true? I believe 3 machines will allow running Q4?

kyuz0:
> I always had in mind that Q2's quality was not good but Q4's is. I don't know if that is still true? I believe 3 machines will allow running Q4?

I have found DeepSeek V4 Flash and now DeepSeek V4.1 Flash to be quite resistant to quantization. Also, antirez' quant is not just flat 2-bit, quite a few weights are left at higher precision, I'd definitely give 4.1 a try in Q2 and get a feeling for youtself.

methuselah-0:
More testing:
Before optimization: 20:26 was the avg task time I got on swebench mini verified using pi-bench with decode around 13t/s. 94% score.

16:54 now after optimizations at commit 40b1f63 and decode around 15t/s. Still 94% (47/50), but maybe surprisingly only 2 of the 3 failures were common in between these 2 test runs.

@davidcanar you can check https://pi-local-coding-bench.dev/ for model benchmark score comparisons with some other local-capable models at higher quant's.

kyuz0:
Thank you @methuselah-0. I should add a note to https://pi-local-coding-bench.dev/ that I am not using that benchmark any longer, i think the verifiers for some tasks are not great and can incorrectly reject working solutions. I tried to mitigate this using an LLM judge, but that is what adds too much variability.

Right now I am using Terminal Bench Mini, which is a subset of 19 tasks from Terminal Bench 2.1 that I curated to test local model engines/quants, to find inference bugs and focus on long-horizon agentic/coding tasks that are much longer than what's in that SWE Bench Mini.

DS4.1F did very well, finishes all the tasks with way less tokens wasted than say V4.

The usual disclaimer on benchmarks: these are mostly to make sure models are reasonably capable and don't lose coherence over a long context and can complete tasks, but comparing model performances is really hard, no benchmark captures that really!

hitman249:
# DeepSeek-V4.1 Flash Q2 — Strix Halo

## Hardware

| Component | Configuration |
|---|---|
| CPU | FEVM FAEX1 AI MAX+ 395 |
| GPU | Radeon 8060S |
| Unified RAM | 128 GB |
| Model SSD | 3× Apacer AS2280Q4X 1 TB, RAID-0 |
| RAID read | **13.5 GB/s** |
| OS SSD | 64 GB USB 3 NVMe |
| OS | CachyOS Server |

## DS4

```bash
DS4_SERVER_DEFAULT_TEMP=1.0 DS4_SERVER_DEFAULT_TOP_P=0.95 \
./ds4-server --host 0.0.0.0 --port 11434 --cors --power 100 --rocm \
  -m "/mnt/raid0/models/antirez/deepseek-v4.1-flash-gguf/DeepSeek-V4.1-Flash-Q2.gguf" \
  --vision "/mnt/raid0/models/antirez/deepseek-v4.1-flash-gguf/DeepSeek-V4.1-Flash-Vision.gguf" \
  --ssd-streaming --ssd-streaming-cache-experts 96GB \
  --ctx 400000 --tokens 131072 --batched-session 1
```

### Memory

| Metric | Result |
|---|---:|
| Context | 400K |
| SSD expert storage | 96 GB |
| Dynamic expert cache | 88.88 GB |
| Expert staging | 7.12 GB |
| Static weights | 9.37 GB |
| Graph | 5.66 GB |
| KV + buffers | 5.65 GB |
| Total planned | **111.02 GiB** |
| Engram | Disk-only |

## Performance

| Metric | Result |
|---|---:|
| Prefill | **~70–111 t/s** |
| Typical prefill | **~95–105 t/s** |
| Decode | **~9–10 t/s** |
| Peak observed NVMe read | **~11 GB/s** |

During prefill, RAID read reached ~11 GB/s.

## KV cache issue

Example:

```text
live   = 55572
prompt = 55332
common = 55237
reason = token-mismatch
```

Only **95 tokens** differ, but DS4 starts a full prefill from `0..55332`.

Stopping and continuing a chat can similarly trigger a full context recomputation.

## GLM-5.3 Flash comparison

Same hardware:

| Runtime | Decode | Prefill |
|---|---:|---:|
| DS4 | ~5 t/s | — |
| Unsloth llama.cpp (`glm5next/upstream`) | **15.6 t/s** | **up to ~200 t/s** |

This suggests there may be substantial performance headroom for V4.1 on llama.cpp, but this has not yet been benchmarked.

jyatesdotdev:
Have been running this for a couple days, working great. Here is my setup!

System details:
 - Fedora Linux 43 Server
 - Kernel 7.1.5-101.fc43.x86_64
 - Driver: in-tree amdgpu
 - ROCm 10.0.0 (/opt/rocm/core) - from 10.0.0 tree, not Fedora RPM
 - HIP 7.15.26333
 - LLVM: AMD clang 23.0.0git (8f497e09…)
 - hipBLASLt 1.4 (libhipblaslt.so.1.4, build 100401@8d1ae90e)
 - rocBLAS 5.6 (librocblas.so.5.6, 5.6.0.8d1ae90e)

### DS4 flags being passed:

#### Coordinator
 ```
   /opt/ds4-rocm10-0c9b62a/ds4-server \
     --rocm \
     -m /home/jryates/models/ds41-flash-q2/DeepSeek-V4.1-Flash-Q2.gguf \
     -c 393216 \
     --tensor-parallel \
     --transport tcp \
     --role coordinator \
     --listen 10.99.0.2 19475 \
     --host 0.0.0.0 --port 8080 \
     -n 65536 \
     --kv-disk-dir /var/lib/ds4-kv \
     --kv-disk-space-mb 131072 \
     --kv-cache-cold-max-tokens 393216 \
     --kv-cache-continued-interval-tokens 2048 \
     --kv-cache-reject-different-quant
 ```

#### Worker
 ```
   /opt/ds4-rocm10-0c9b62a/ds4 \
     --rocm \
     -m /home/jryates/models/ds41-flash-q2/DeepSeek-V4.1-Flash-Q2.gguf \
     -c 393216 \
     --tensor-parallel \
     --transport tcp \
     --role worker \
     --coordinator 10.99.0.2 19475
 ```

Attaching a snapshot of my grafana dashboard so you can see performance details, but the headline is that dual gfx1151 TP holds ~14 t/s decode, prefills ~150 t/s typical / 450+ peak, both GPUs ~95–100% when a request is running.

https://snapshots.raintank.io/dashboard/snapshot/a8VWBaz3K10t1zt7Ja8hX7n5zpU5GAtD

Thanks for this @kyuz0!

kyuz0:
Thank you all for testing. I am rebasing on main and I also found a way to speed up the image encoder, and I'll be testing some more speed ups, will push an update this evening after testing, but it should get faster, especially for images.

PR #1037, opened 2026-09-13
Fix missing Metal object header dependencies

IngeniousIdiocy:
Rebuild `ds4_metal.o` when `ds4.h` or `ds4_image.h` changes.

Build passes on macOS; `make -n -W <header> ds4_metal.o` confirms both headers now trigger recompilation.


emilianbold (approved):
This is correct but I wonder if we should't use https://www.gnu.org/software/make/manual/make.html#Generating-Prerequisites-Automatically or something similar so these makefile header dependencies don't drift so much.

IngeniousIdiocy:
> This is correct but I wonder if we should't use https://www.gnu.org/software/make/manual/make.html#Generating-Prerequisites-Automatically or something similar so these makefile header dependencies don't drift so much.

entirely up to you guys. I normally play in my own branch of ds4 doing things not really appropriate for the whole project but this one felt like I owed a PR. thanks for all the work in maintaining it.

IngeniousIdiocy:
Landed in main via cherry-pick as 064bc229c691f1ad40d8a39c3e5f2aa1fa9b1dba; closing this PR.

PR #1041, opened 2026-09-14
v41: queue single-box decode layers and commit each without waiting (+37% decode on M3 Ultra, bit-exact)

adriangalilea:
On a single Mac, `ds41_graph_step` commits **and waits** for the GPU after every one of the 40 layers; only the two-Mac TP path keeps the token's command buffers queued. This PR queues them on one box too (draining at layer 13, before the shared Engram input is reused, and at the last layer — exactly where the TP path drains) and additionally commits each queued layer without waiting, so the GPU starts a layer while the CPU encodes the next.

Related: #1034 (Dango233) opens the same queue as an opt-in restricted to SSD streaming, measured +11 % on an M2 Ultra Q2. The resident path gains more and nothing in the mechanism is streaming-specific, hence default-on here with rollback envs, mirroring the TP path's `DS4_METAL_DISABLE_V41_TP_DECODE_QUEUE`.

### Measurements

Mac Studio M3 Ultra 512 GB, `DeepSeek-V4.1-Flash-Q4.gguf` fully resident, `./ds4 -p … -n N --temp 0`, ABBA on the same binary (A = per-layer drain via the rollback env, B = this PR), **generated text byte-identical in every pair** (md5 of the output equal across all runs).

| prompt / tokens | per-layer drain | queued | queued + commit per layer |
|---|---:|---:|---:|
| 2k / 128 | 17.08 / 16.78 t/s | 21.73 / 21.72 | — |
| 16k / 128 | 16.67 / 16.68 | 21.47 / 21.42 | — |
| 2k / 512 | 16.68 | 21.60 | — |
| 2k / 256 | — | 21.63 / 21.63 | **23.08 / 23.09** |

Per-token GPU stage timestamps (non-blocking, from #828, extended to the V4.1 graph on that branch) explain the size of the gain: with the drains, 37.3 ms of GPU busy inside a 46.9 ms span per token — 9.6 ms of scheduling gap from the 40 waits; queued, 36.9 ms busy in a 37.7 ms span (0.75 ms gap). The remaining per-token cost is the kernels themselves (routed experts 8.6 ms, attention output 4.8, attention projections 4.3, shared expert 3.1, router 2.7 …), which this PR does not touch.

Composes with #1035 (parallel Engram decode reads): both together 24.86 / 24.85 t/s on this box, i.e. +48 % over the current default.

### Rollbacks

- `DS4_METAL_DISABLE_V41_DECODE_QUEUE=1`: per-layer drain on a single box (previous behaviour).
- `DS4_METAL_DISABLE_V41_DECODE_FLUSH=1`: keep the token in one command buffer between drains (queue only).

Not changed: the batched-session path (`ds41_graph_step_batch` has its own layer-13 drain), TP, quality/imatrix modes (`layer_resident` still drains per layer).


adriangalilea:
Added `32e72d1857fe9d51ebc4984031382d2812c6234c`: queue resident single-box Metal decode through the logits tail. The idle prefill staging buffer holds the second Engram input, removing the layer-14 drain without an extra allocation; output projection joins the queue before its final wait. Rollback: `DS4_METAL_DISABLE_V41_WHOLE_TOKEN=1`.

Measured this exact PR branch on an M3 Ultra 512 GB (60 GPU cores), DeepSeek V4.1 Flash Q4_K, context 32768, a **668-token** prompt and 256 greedy generated tokens. ABBA, old schedule versus whole-token:

| | A | B | B | A |
|---|---:|---:|---:|---:|
| Decode t/s | 22.52 | 22.87 | 22.84 | 22.55 |
| Prefill t/s | 236.60 | 237.00 | 236.90 | 236.34 |

Mean decode 22.535 → 22.855 t/s, **+1.42%**. All four outputs are byte-identical (`cmp`); MD5 `40e4c6fa5651b8ca8be4a2cb5eb1166e`. This clean branch excludes #1035, #1042 and #1043. Separately, the integration stack measured 30.45 → 31.075 t/s (+2.05%) for the same scheduling change.

Validation on the measured SHA: `make -j8 ds4 ds4-bench tests/test_deepseek41_metal`, the full `tests/test_deepseek41_metal` suite, `--hc-scaled`, and `--bf16-linear` passed. The three fusion-specific tests belong to #1042 and passed on the integration commit; they are not dependencies of this PR. CPU build also passed for the original integration commit on M4 Pro; CUDA/distributed/SSD schedules retain their existing path.

ABBA command: `./ds4 -m /path/to/DeepSeek-V4.1-Flash-Q4.gguf -p "$(head -c 2000 speed-bench/promessi_sposi.txt) Continua la storia:" -n 256 --temp 0`, setting `DS4_METAL_DISABLE_V41_WHOLE_TOKEN=1` for A. Each run fetched the commit into an isolated Studio clone. #1042 remains independent of this PR.

Dango233:
Thanks for extending this and measuring the resident path. I maintain #1034 and would prefer to converge on one queue implementation once the SSD path is covered. I can contribute the M2 Ultra 192 GB / V4.1 Q2 SSD case; I am keeping #1034 open until that comparison is complete.

Our existing forced-eviction check uses two sessions sharing a 512-expert cache and compares every vocabulary logit, Engram history and saved cache span across positions 121–144. The same-engine 512-step harness also compares full logits and selected tokens. These should be useful alongside your resident Q4 results.

One scope point from reading current head `fdbf7f2`: `whole_token` is explicitly restricted to Apple resident single-host decode, but the new `queue_layers` expression and per-layer `ds4_gpu_flush_commands()` condition are broader. They also admit SSD streaming and the flush condition does not exclude TP; the queue expression is outside an Apple guard in the shared V4.1 graph. Could we make the intended platform/TP scope explicit, or include validation for those paths before treating them as unchanged? #1034 guards its added single-host eligibility with `__APPLE__` and excludes quality/imatrix.

I will report the exact candidate and rollback settings with the M2 results. No need to fold the Engram or slab changes into this PR; they can remain independent.

Dango233:
M2 Ultra Q2 SSD results are now complete for exact head `fdbf7f27ec074251badbdbb6c21ec9e3c3f09759`. The tested production code is unchanged; I added only the forced-eviction test from #1034 and its benchmark's `--ssd-streaming` plumbing locally.

M2 Ultra 192 GB, macOS 15.7.4; calibrated V4.1 Flash IQ2_XXS/Q2_K, SHA-256 `1ce6a8f8806205c13330d7ca287bd198331dc5ca35ccc5d8a9a92a188a6f6f42`. One inference process at a time. Both arms use `DS4_METAL_DISABLE_STREAMING_EXPERT_SLABS=1`; no Engram optimization or router fix. A sets `DS4_METAL_DISABLE_V41_DECODE_QUEUE=1`; B leaves queue/flush rollback variables unset. Whole-token scheduling is ineligible in SSD mode.

Fresh-process ABBA, `ds4-bench`, 2,048 prompt tokens from `speed-bench/promessi_sposi.txt`, `--ctx-alloc 8257`, automatic expert cache, 512 generated tokens; no OS/file-cache flush:

| Variant | Prefill t/s | Decode t/s |
| --- | ---: | ---: |
| rollback | 74.30 | 10.66 |
| PR default | 69.14 | 12.11 |
| PR default | 67.72 | 12.04 |
| rollback | 68.79 | 10.61 |

Mean decode **10.635 → 12.075 t/s (+13.5%)**. All four generated texts match exactly.

Two more checks pass:

- Metal API validation, two sessions sharing a 512-expert cache to force eviction: full logits, finite values, Engram history and every saved cache span match at positions 121–144.
- Balanced same-engine harness, 16 warmup steps and 512 measured steps per variant, alternating order/session assignment: **529 logit rows, 68,389,120 floats and 528 selected IDs match exactly**. PR default is 13.6732 t/s; rollback 11.8228 t/s. In the harness output, `control` means PR default and `candidate` means the queue-disable rollback.

This supports the single-host SSD path in addition to your resident results. It does not validate CUDA, TP, quality/imatrix or resident whole-token behavior on this machine; the scope question in my earlier comment remains. I am keeping #1034 open pending that clarification and convergence, rather than maintaining two competing queue changes indefinitely.

Issue #1039, opened 2026-09-14
test_deepseek41_metal: new router check fails on M1 Ultra (pre-M5); sqrt(softplus) compared at 3e-6 but 1.0f+exp is ~0.7% imprecise


jasontitus:
**Clarification:** this report is about a failure in the model-free `tests/test_deepseek41_metal` router numerical check on our M1 Ultra. V4.1 inference works on this machine: we have successfully run server requests and inference benchmarks. We are not reporting an inference crash or inability to generate. The test failure also reproduces on pristine upstream `a04f46f`, without our KV compression PR (#186); see the fresh reproduction below in the comments.

---

`tests/test_deepseek41_metal` fails the new router check on M1 Ultra:

```
router n=256 mode=0 rows=1 row=0 expert=4 logit=-11.8886719 actual=0.00263455603 ref=0.00262947031
tests/test_deepseek41_metal.c:105: fabs(probs[t * n + e] - ref[e]) <= 3e-6 * (1 + ref[e])
ds4: Metal cleanup discarded 5 live tensor handles
make: *** [test-deepseek41-metal] Error 1
```

## Repro

```sh
make tests/test_deepseek41_metal && ./tests/test_deepseek41_metal
```

No model or GGUF needed. Fails every run, identical values.

## Environment

| | |
|---|---|
| ds4 commit | `a04f46f` (DeepSeek v4.1 Flash support for CUDA) |
| Hardware | Mac Studio, Apple M1 Ultra, 128 GB |
| OS | macOS 26.6.2 (Darwin 25.6.0) |
| Backend | Metal, `make` default target |
| Metal 4 | unavailable, engine logs `Metal 4 tensor API disabled for pre-M5/pre-A19 devices` |

## Not a regression: the check is new

`check_router` is added wholesale by `a04f46f`; every line of it is new in that commit's diff of `tests/test_deepseek41_metal.c`. Confirmed by bisect on this machine:

| commit | result |
|---|---|
| `bd66c40` DeepSeek v4.1 Flash support for Metal | PASS |
| `a04f46f` DeepSeek v4.1 Flash support for CUDA | FAIL |

So the kernel behaviour is presumably unchanged; the new check is simply tighter than pre-M5 Apple GPUs can satisfy.

## Root cause

The check compares two independent FP32 evaluations of `sqrt(softplus(x))` for a logit of about -11.9, where that expression is ill-conditioned.

[...]

PR #1044, opened 2026-09-14
Improve Metal router softplus accuracy for small logits

emilianbold:
Use a four-term log1p series for exp(x) below 1/32 in all four DeepSeek router transform sites. Clamp the polynomial input so its unselected branch stays finite, and compare Metal probabilities against the shared accurate reference without widening the tolerance.

Refs antirez/ds4#1039. Validated on M5 Max with a fresh default build, test_deepseek41_metal, test_deepseek41_graph --vision-routing, and full Metal test runs with DS4_METAL_DISABLE_PRE_M5_ROUTER_TRANSFORM_FINALIZE_FUSION=1 and DS4_METAL_MATH_SAFE=1. No model required. Pre-M5 validation and model-level speed/SSD checks remain outstanding.


jasontitus:
Tested on the pre-M5 hardware from #1039: this fixes it.

Apple M1 Ultra, 128 GB, macOS 26.6.2. No `DS4_*` environment overrides. The PR head is `7f5f1b7`, and the control is its base `9139e2a`.

```sh
make tests/test_deepseek41_metal tests/test_deepseek41_graph
./tests/test_deepseek41_metal
```

| Build | Configuration | Result |
|---|---|---|
| `9139e2a` base | default, 3 runs | FAIL 3/3, identical at `test_deepseek41_metal.c:105` (`actual=0.00263455603 ref=0.00262947031`) |
| `7f5f1b7` PR | default, 3 runs | PASS 3/3 |
| `7f5f1b7` PR | `DS4_METAL_DISABLE_PRE_M5_ROUTER_TRANSFORM_FINALIZE_FUSION=1` | PASS |
| `7f5f1b7` PR | `DS4_METAL_MATH_SAFE=1` | PASS |
| `7f5f1b7` PR | `test_deepseek41_graph --vision-routing` | PASS |

On M1 the default run takes the pre-M5 fused `router_transform_finalize_weights_one_simd` path. With fusion disabled it takes `softplus_sqrt_f32_4`. Both pass against the accurate `log1p` reference at the unchanged 3e-6 tolerance, so the fix is in the kernels and the check was not loosened.

Not run here: model-level speed or SSD checks.

Dango233:
Confirmed on M2 Ultra 192 GB, macOS 15.7.4. Tested pristine base `9139e2ae58a41503968a500f36f75895c1ba63fc` and exact PR head `7f5f1b77ae831ec0ec5bb97486efa8600c57bd82` in separate clean worktrees.

- Base: full `tests/test_deepseek41_metal` fails 3/3 at line 105 with the same `actual=0.00263455603 ref=0.00262947031` reported on M1 Ultra.
- PR: full suite passes 3/3 with no overrides. Also passes separately with pre-M5 router fusion disabled, `DS4_METAL_MATH_SAFE=1`, and `MTL_DEBUG_LAYER=1`.
- `tests/test_deepseek41_graph --vision-routing` passes. The real-weight `--session-fixture` passes under Metal API validation, covering prefix reuse, snapshots, cancellation and bounds.
- Clean Metal build, CPU build, restored Metal links, and frontend/Engram/V4.1 GGUF/quality-tool unit tests pass.

I also ran a Q2 SSD model comparison: calibrated V4.1 Flash IQ2_XXS/Q2_K, SHA-256 `1ce6a8f8806205c13330d7ca287bd198331dc5ca35ccc5d8a9a92a188a6f6f42`; `ds4-bench`, 2,048-token `speed-bench/promessi_sposi.txt` prompt, context allocation 8,257, 512 generated tokens, automatic cache, fresh processes in ABBA order. Both arms set `DS4_METAL_DISABLE_STREAMING_EXPERT_SLABS=1` to avoid the separate large-slab latency issue; no companion queue/Engram optimizations. OS/file cache was not flushed.

| Variant | Prefill t/s | Decode t/s |
| --- | ---: | ---: |
| base | 64.91 | 10.66 |
| fix | 65.44 | 10.79 |
| fix | 68.08 | 11.25 |
| base | 68.70 | 10.59 |

All four model runs complete. The two base outputs match each other, and the two fixed outputs match each other, but **base and fix generated text differ**. That is consistent with changing router arithmetic; these throughput observations follow different generated sequences and should not be interpreted as a controlled speedup or output-parity claim. I have not evaluated model quality, CUDA or ROCm.

antirez:
The problem I have with that is that once the fix is applied, the quality checked against the *official* API continuations worsens: mean matching common prefix of prompts goes from 7.03 tokens to 6.40 tokens, which is significant. I can't explain it, if not thinking that for some reason DeepSeek inference engine uses something that approximates more nearly to the old code. Or alternatively, the different routing compensates for quantization (!) somehow. Ideas?

jasontitus:
Not sure either, but here is a bit more eval from the hardware:

From Fable - 

1. **The LCP shift is within noise; the continuous metrics favour the PR.** 
Same binary, kernels swapped via the `DS4_METAL_*_SOURCE` overrides, `deepseek-v4.1-flash-20260910-general` (100 cases), V4.1 Flash Q2, ctx 4096, SSD streaming. Paired per case, bootstrap 95% CI:

| metric | main | PR | Δ | 95% CI |
|---|---|---|---|---|
| **greedy LCP** | 9.51 | 8.94 | −0.57 | [−1.47, +0.27] (79 cases tie, 7 up, 14 down) |
| **NLL (sum)** | 10.948 | 10.906 | −0.4% | [−1.8%, +1.0%] (51 better, 49 worse) |
| **api_target_mae** | 0.2216 | 0.2179 | −0.0037 | [−0.0086, +0.0011] |
| **first-token matches** | 77 | 79 | | |

Same direction and size as your 7.03 → 6.40, but LCP is bimodal here (23/100 cases at 0, tail to 32, paired SD 4.4), so a 0.6 shift on 100 cases isn't resolvable. NLL and logprob agreement with the API are flat to slightly better.

2. **The only behavioural difference is tie-breaking among flushed-to-zero experts.** 
I dumped logits/probs/bias/selection per dispatch from `main` and replayed 60,800 (token, layer) decisions against `sqrt(log1p(exp(x)))` in `float64`:
* Top-6 membership differs on 304 rows (0.50%), and every one is an exact tie (rank-6/rank-7 score gap = 0.0).
* For logits < −16.6, `1.0f + exp(x) == 1.0f`, so `main` outputs prob exactly 0 and the selection score collapses to bias alone. `exp_probs_b` has only 63 distinct fp32 values across 384 experts, so ties are guaranteed and the bitonic sort breaks them by position.
* 421 rows (0.7%) select at least one prob-0 expert; in 62 rows it's 5 of 6, so that token runs on one expert with the full 1.5 weight.
* Elsewhere the formulas agree: rel. error ≤ 4e-6 for x ∈ [−4, 0), ≤ 2e-4 in [−8, −4); the rank-6 logit is at median −1.2 / p1 −8.7, inside that range.

So there is no near-boundary sensitivity and nothing for quantization to compensate: `main` decides ~1 in 200 routings by sort order among tied experts, the PR decides them by logit. That's also why base and fix generate different text.

On "DeepSeek's engine approximates the old code": it would need fp32 `1 + exp(x)` and the same tie order; PyTorch softplus is log1p-based and bf16 keeps fp32's exponent range, so `exp(−18)` doesn't flush. `main` also isn't a single behaviour to preserve — the M1 kernel gives `0.0026346` where M5 stays within 3e-6 of `0.0026295` on the same logit (#1039).

Can run `--quality` or the long fixture if useful; the dump hook is ~50 lines if you want it as a diagnostic.

emilianbold:
> quality checked against the _official_ API continuations worsens

I just started looking at `quality-testing` but I'm not sure the `avg_lcp` is a good quality measurement. `avg_lcp` is not even stable for *unchanged* code.

I executed `score_official` on the 100 cases in `data/flash` (V4 Flash Q2, M5 Max) for 9 times with the base kernels only, changing nothing but how the prompt is prefilled (`--continued-prefill N`) or fast/safe math. (Each configuration is bit-reproducible).

The results were `avg_lcp` = `4.61, 4.82, 4.90, 4.97, 5.14, 5.17, 5.30, 5.54, 5.72` a **1.1-token spread** (more than the 0.6 tokens you noticed).

The PR kernels under the same nine conditions land inside the same band (`4.57–5.59`). Comparing fix vs base gives `−0.12, +0.19, +0.07, −0.14, −0.25, +0.28, −0.04, +0.03, +0.05`.

The reason seems mechanical according to router dumps: the router has 25% of its top-6 boundaries within 0.01 score units and 5% within 0.001, so even a small perturbation flips a few expert choices and every case's logits end up different. `avg_lcp` then moves by whole tokens per case (it is a first-mismatch counter) and the 100-prompt mean by several tenths.

If you're running this for 4.1, note the README in `deepseek-v4.1-flash-20260910/README.md` mentions:

> These are sampled continuations, not greedy continuations or full logits.
> Use teacher-forced NLL and API top-token agreement; a short greedy prefix alone is not a quality failure for this set.

... so I'm reading this as `avg_lcp` not being recommended as a metric.

I think a better way to compare outputs for 4.1 (which are sampled) would be something inspired from how we do speculative decoding: keep track of residuals and have a metric based on them.

PS: I haven't verified this thoroughly but I did have AI implement this overlap metric (sums the minimum of the two models' probabilities, token by token) to `score_official.c`, then collected the values for deepseek-v4-flash-0731 (bf16, temperature 1 with top-20 logprobs).

This metric seems much more stable: for unchanged kernels overlap spans `0.8283–0.8314` (a 0.4% band, vs 11% for `avg_lcp` on the same runs), and the PR moves it by `−0.0001 ± 0.0002`.

See https://github.com/emilianbold/ds4/commit/f3e16a19d37f9c17d2f42869474742e9993c7098#diff-704d7f59f343ec67141ad40f86d6e322e1d11936a1be1ed85125ab83c168c2d1R1055-R1066  (branch https://github.com/emilianbold/ds4/tree/metric/api-overlap ) Which is to say, this metric shows no measureable loss due to the PR.

antirez:
I used greedy Vision-Exp continuations from the Novita open router provider. This is not the same of what you did. One thing is the argument that LCP is not the only metric of quality, but here you reproduced something different.

antirez:
P.S. I'm collecting a larger set of official greedy outputs to test if the same result can be reproduced for new datasets.

antirez:
Ok, on a larger corpus I no longer see any regression, nor any improvement to be honest. I will merge the PR based on the fact that it is conceptually more correct, even if I failed to prove myself that in the real world the inference is higher quality in this way. Thanks.

emilianbold:
Closing, cherry-picked in main https://github.com/antirez/ds4/commit/0719a0baf8d9f1128fcd91a6cd5588f56362b01e

PR #1048, opened 2026-09-14
Remove unsupported Hugging Face download resume promise

pminervini:
`./download_model.sh ds41f-q2` uses `hf download` but tells users that rerunning an interrupted download will resume it. That promise is incorrect with `huggingface_hub 1.31.0`: a new attempt creates a fresh temporary file instead of reusing the previous partial file.

Remove the unconditional resume message from `download_one_hf()`. The curl path retains its message because it explicitly uses `curl -C -`.

Evidence from the Hugging Face v1.31.0 source (links pinned to its commit):

- [`_download_to_tmp_and_move`, lines 1961–1970](https://github.com/huggingface/huggingface_hub/blob/495b17c8529614759ae0f1ccf1ebe9a61c148b7c/src/huggingface_hub/file_download.py#L1961-L1970) creates a UUID-suffixed temporary filename and opens it with `wb`. The preceding comment explains that shared partial files could be corrupted by concurrent writers on filesystems whose locks are not honored.
- [Lines 1978–2009](https://github.com/huggingface/huggingface_hub/blob/495b17c8529614759ae0f1ccf1ebe9a61c148b7c/src/huggingface_hub/file_download.py#L1978-L2009) use that temporary file for the Xet or HTTP transfer, then delete it in `finally` on failure. The cleanup comment explicitly explains why that partial file cannot be reused.

This removes a guarantee the wrapper cannot make across installed Hugging Face client versions. It does not change download behavior.

Validation: `sh -n download_model.sh`; `python3 tests/test_model_download.py` (9 tests passed); `git diff --check`.


emilianbold:
Yep, I just downloaded 30GB and it failed midway. On restart, it did not resume... 😮‍💨

download_model.sh
```diff
@@ -533,7 +533,6 @@ download_one_hf() {
     echo "Downloading $file"
     echo "from https://huggingface.co/$REPO"
     echo "using $HF_CMD download"
-    echo "If the download stops, run the same command again to resume it."
```
emilianbold, line 536: I mean, this is a step forward but I'd rather see actual resume work.

pminervini: I did a thing but that would bloat the codebase

emilianbold: I mean, PR get merged very slowly so you might as well open a PR for resume too, if you have it working.

pminervini: here you go:

```bash
#!/bin/sh
set -eu
cd -- "$(dirname -- "$0")"
mkdir -p "${DS4_GGUF_DIR:-gguf}"
cd -- "${DS4_GGUF_DIR:-gguf}"
file=DeepSeek-V4.1-Flash-Q2.gguf
[ ! -f "$file" ] || exit 0
touch "$file.part"
while [ "$(wc -c < "$file.part")" -lt 365713686528 ]; do
    start=$(( $(wc -c < "$file.part") ))
    end=$((start + 67108864 - 1))
    [ "$end" -lt 365713686528 ] || end=365713686527
    curl -fL -C - -H "Range: bytes=$start-$end" -o "$file.part" \
        "https://huggingface.co/antirez/deepseek-v4.1-flash-gguf/resolve/main/$file"
done
[ "$(wc -c < "$file.part")" -eq 365713686528 ]
mv "$file.part" "$file"
```

pminervini: > I mean, PR get merged very slowly so you might as well open a PR for resume too, if you have it working.

`hf download` explicitly doesn't support resume (and I kind of trust them on this) but let's see what we can do

PR #1051, opened 2026-09-14
Fix GLM-5.3 Metal decode masking for padded selections

IngeniousIdiocy:
GLM-5.3’s pooled selection pads unused tail slots with 0xffffffff. The serial decode call currently passes selected_rows_valid = true, causing Metal’s split-group8 kernel to skip bounds checks and read those slots as cache rows, diluting the attention output.

This addresses the decode counterpart of the prefill masking fix in 0e9cc2d.


IngeniousIdiocy:
Landed in main via cherry-pick as 582d809fd300b0d0284911e4f351dce26ef095b5; closing this PR.

PR #1056, opened 2026-09-15
Metal: optimize Qwen3.8 kernels, MTP state and SSD MoE scheduling (Up tp 40%)

GiorgioOppo:
Qwen3.8 Flash Next spends substantial time preparing routed experts, waiting for SSD reads, and repeating small projection work during speculative decoding. This series reduces those costs on Metal, prepares the MTP predictor's historical cache correctly, and overlaps selected-expert reads with useful GPU work during prefill and decode.

### Relationship to #1047

This builds on #1047, **Metal: add SSD expert streaming for Qwen3.8 Flash Next**. That PR introduces bounded routed-expert streaming, selective staging for cache overflow/MTP, and the initial decode gate/up overlap.

#1047 is still open. Consequently, this PR against `antirez:main` currently includes its three commits (`e37f185`, `85d37ae`, `d6cbc77`) plus eight follow-up commits. Review the streaming foundation in #1047 and the [incremental optimization diff](https://github.com/GiorgioOppo/ds4/compare/d6cbc77ca6b9b67c004af7f8a0e41b90ba3bf9ca...5cbdb6ef81d6b079ddb5ba53657360257125c651) here. The requested head is `GiorgioOppo:qwen-kernel-opt`, currently `5cbdb6e`.

### Changes beyond the streaming foundation

- **Reuse inputs in Metal kernels.** Reuse IQ2_XXS gate/up activations across two rows, share F16 hyper-connection inputs and sigmoid values, and reuse Q2_K down activations across two output rows. Preserve the logical 640-column width independently of the physical 768-column Q2_K row stride. Keep generic fallbacks and the original per-row arithmetic/reductions.
- **Reduce padded prefill work.** Select narrower MoE matrix tiles from the existing routing frequencies for sparse expert batches, and use type-specialized IQ2_XXS/Q2_K pipelines. Automatic hardware-specific choices are restricted to the measured M1 Max shapes and batch ranges.
- **Prepare and preserve MTP state.** Populate historical predictor K/V and indexer keys during prompt prefill, retain the last trunk row until its actual successor is known, and preserve it across append, verification, rejection and checkpoint restore. Cache-only preparation skips unused query/output projections and predictor expert reads. Catch-up runs the predictor MoE only for the consumed final row; a Q8 projection specializes the two active halves of the MTP input.
- **Recover expert-cache capacity and reduce CPU preparation.** Bound transient MTP staging from the actual compact/fallback lifetimes. On the measured Q2/4096-context configuration, the automatic cache grows from 7,371 to 8,212 entries within the same planned memory budget. Plan batch eviction victims once while preserving cache ordering, counters and inflight protection.
- **Overlap SSD reads and GPU computation.** Extend overlap through cached down projections and T2/T3 MTP verification. For supported batched prefill, run resident experts first, submit missing gate/up work as soon as those payloads arrive, and overlap it with down-weight reads. Immutable address snapshots, pending-buffer retention and failure drains protect command-buffer lifetimes; cache entries become valid only after complete reads.

No GGUF conversion or weight-precision change is required. These changes do not load all experts speculatively: routed reads remain limited to selected missing experts, with the existing full-layer fallback where required.

### Performance

Measurements were recorded on **Apple M1 Max, 32 GiB**, with `Qwen3.8-Flash-Next-Q2.gguf`, Metal SSD streaming, automatic expert cache, `--ctx 4096 --prefill-chunk 128 --temp 0 --nothink`.

The final prefill-pipeline comparison used **32 runs in ABBA/BAAB order**, four observations per arm and configuration. The short prompt was `narrami la storia di roma` (29 input tokens, 100-token output limit); the long prompt had 575 input tokens and a 50-token output limit. MTP was tested both on and off.

**This is an incremental pipeline comparison.** Its baseline already contains resident-prefill overlap and Q2_K down-row reuse on top of `81b9ebb`; it is not a clean comparison against that commit, #1047, or upstream main. There is no measured whole-PR speedup, and results from separate optimization stages must not be added together.

| Workload | Prefill t/s, baseline → pipeline | Change | Decode t/s, baseline → pipeline | Change |
| Short, MTP off | 9.820 → 10.710 | +9.06% | 7.950 → 7.820 | -1.64% |
| Short, MTP on | 9.605 → 10.650 | +10.88% | 9.765 → 9.650 | -1.18% |
| Long, MTP off | 46.900 → 51.380 | +9.55% | 9.125 → 8.720 | -4.44% |
| Long, MTP on | 45.735 → 48.665 | +6.41% | 10.130 → 10.440 | +3.06% |

Values are medians. All 32 runs completed successfully and produced byte-identical output for the same prompt. Expert-cache budgets, entry counts, hits, misses and read volumes matched within each configuration. Recorded source hashes match the final committed implementation.

**The prefill improvement has a decode tradeoff:** long-prompt ordinary decode decreased by 4.44% in this comparison. Four observations per arm do not rule out a regression, even where the ranges overlap. Token rates exclude initial model loading; filesystem caches were not purged and desktop/background activity was not controlled. These are local measurements, not a uniform speed guarantee.

A separate earlier 32-run comparison measured the staging-bound, eviction-planning and decode-overlap series (`9e1429b` through `81b9ebb`) against `b86c8ae`. Short/long median decode improved by 23.27%/31.05% with MTP and 7.28%/9.57% without it, with identical output. Those results cover that three-commit series only; baseline MTP decode was variable (7.71–10.82 t/s short, 8.25–10.78 t/s long), and they are not cumulative with the table above.

### Validation

Executed during development on the M1 Max/Q2 configuration:

- Metal application builds, Qwen memory-estimation tests, full Qwen kernel and MoE specialization suites, including T8191/T8192/T8193 boundaries and guarded/odd-width Q2_K rows.
- SSD cold/mixed/warm cache tests with ordinary and unretained command buffers, separate expert buffers, format fallbacks, selected-read assertions, and ownership/batch restoration. Eighteen additional down-read failure cases span three quantization pairs and T29/T33/T128; each checks four consecutive failures, an exact retry and a warm pass.
- Cache-planning tests against an independent eviction oracle, including a full 40,960-entry cache, ties, protected/inflight entries, stale plans and interrupted consumption.
- Real-model MTP cache/projection/residual/full-logit comparisons, recursive drafts, append, checkpoint/restore/rewind, exact-sampling continuation and n-gram read-failure recovery. Direct kernel/predictor comparisons were bitwise exact; independently batched target continuation used the existing tolerance, with a maximum observed absolute logit difference of `0.000974178`.
- Final performance series: 32 successful runs with identical output, following 64 exploratory runs. Experimental Q8 row variants and FP16 activation staging were removed after full-model measurements.

`git diff --check` passes for the complete PR diff. CUDA/ROCm, M5 NAX execution and vision were not validated by this series; performance claims are limited to the tested Metal/M1 Max workloads.

**Checkpoint compatibility:** Qwen payload version 3 stores the prepared predictor prefix and pending trunk row. Older Qwen disk checkpoints must be rebuilt.

Example with built-in MTP:

```sh
./ds4 --metal -m /path/to/Qwen3.8-Flash-Next-Q2.gguf \
  --ssd-streaming --ctx 4096 --prefill-chunk 128 \
  --mtp-timing --temp 0 --nothink -n 100 \
  -p "narrami la storia di roma"
```

Omit `--mtp-timing` to compare ordinary decoding with the same model and settings.

<details>
<summary>Raw benchmark results — final 32-run pipeline comparison</summary>

Rows are in recorded execution order. Baseline and pipeline refer only to the incremental comparison described above. Process time includes startup; prefill/decode rates do not. Every row exited successfully and matched the expected output bytes.

| Run | Prompt | MTP | Arm | Prefill t/s | Decode t/s | Process time (s) |
| ---: | --- | --- | --- | ---: | ---: | ---: |
| 1 | Short | On | Baseline | 9.52 | 9.21 | 15.126 |
| 2 | Short | On | Pipeline | 10.43 | 9.61 | 18.329 |
| 3 | Short | On | Pipeline | 10.52 | 9.29 | 14.744 |
| 4 | Short | On | Baseline | 9.54 | 9.55 | 14.814 |
| 5 | Long | On | Baseline | 44.44 | 10.05 | 19.113 |
| 6 | Long | On | Pipeline | 45.82 | 10.34 | 18.669 |
| 7 | Long | On | Pipeline | 49.92 | 10.62 | 17.434 |
| 8 | Long | On | Baseline | 44.55 | 9.95 | 19.261 |
| 9 | Short | On | Pipeline | 10.80 | 9.69 | 14.225 |
| 10 | Short | On | Baseline | 9.81 | 10.07 | 14.088 |
| 11 | Short | On | Baseline | 9.67 | 9.98 | 14.261 |
| 12 | Short | On | Pipeline | 10.78 | 10.07 | 13.880 |
| 13 | Long | On | Pipeline | 47.93 | 10.54 | 17.939 |
| 14 | Long | On | Baseline | 47.30 | 10.24 | 18.260 |
| 15 | Long | On | Baseline | 46.92 | 10.21 | 18.459 |
| 16 | Long | On | Pipeline | 49.40 | 10.20 | 17.836 |
| 17 | Short | Off | Baseline | 9.91 | 7.93 | 16.684 |
| 18 | Short | Off | Pipeline | 10.74 | 7.54 | 17.142 |
| 19 | Short | Off | Pipeline | 10.68 | 7.71 | 17.028 |
| 20 | Short | Off | Baseline | 9.83 | 7.62 | 17.207 |
| 21 | Long | Off | Baseline | 46.85 | 8.54 | 19.328 |
| 22 | Long | Off | Pipeline | 54.89 | 8.70 | 17.574 |
| 23 | Long | Off | Pipeline | 51.65 | 8.68 | 18.101 |
| 24 | Long | Off | Baseline | 45.96 | 9.16 | 19.262 |
| 25 | Short | Off | Pipeline | 10.38 | 7.93 | 16.555 |
| 26 | Short | Off | Baseline | 9.79 | 7.97 | 16.738 |
| 27 | Short | Off | Baseline | 9.81 | 8.22 | 16.252 |
| 28 | Short | Off | Pipeline | 10.78 | 8.06 | 16.362 |
| 29 | Long | Off | Pipeline | 51.11 | 9.43 | 17.734 |
| 30 | Long | Off | Baseline | 48.60 | 9.59 | 18.240 |
| 31 | Long | Off | Baseline | 46.95 | 9.09 | 18.888 |
| 32 | Long | Off | Pipeline | 50.06 | 8.74 | 18.478 |

</details>


GiorgioOppo:
@evanwtf could you test this pr without ssd streaming?

GiorgioOppo:
## Direct comparison: #1047 versus #1056

This is a new end-to-end comparison of the complete optimization series, replacing the need to infer a cumulative gain from the incremental benchmarks in the PR description.

- **Baseline:** #1047 at `d6cbc77ca6b9b67c004af7f8a0e41b90ba3bf9ca`.
- **Optimized:** #1056 / `qwen-kernel-opt` at `5cbdb6ef81d6b079ddb5ba53657360257125c651`.
- **Hardware/model:** Apple M1 Max, 32 GiB; the same `Qwen3.8-Flash-Next-Q2.gguf` for both versions.
- **Settings:** Metal, SSD streaming, automatic expert cache, context 4096, prefill chunk 128, temperature 0, no thinking; MTP on and off. Both arms used `DS4_METAL_STREAMING_EXPERT_TIMING_SUMMARY=1`; other inherited `DS4_*` overrides were removed.
- **Method:** clean source snapshots and separate binaries, identical `make -j4 ds4` compiler flags, each run loads its own Metal sources. Two excluded long/MTP warmups, then 32 serial runs in ABBA/BAAB order (four observations per version/workload). No concurrent ds4 server or model instance. Filesystem caches were not purged.

The short prompt was `narrami la storia di roma`: 29 input tokens and a 100-token output limit. The long prompt was a historical analysis request: 575 input tokens and a 50-token output limit. The same exact prompt files were used on both versions.

### Median throughput

| Workload | Prefill #1047 → #1056 (t/s) | Change | Decode #1047 → #1056 (t/s) | Change |
| Short, MTP off | 9.170 → 10.725 | +16.96% | 7.560 → 8.095 | +7.08% |
| Short, MTP on | 9.200 → 10.815 | +17.55% | 7.480 → 9.935 | +32.82% |
| Long, MTP off | 35.550 → 53.850 | +51.48% | 9.415 → 9.860 | +4.73% |
| Long, MTP on | 34.650 → 51.245 | +47.89% | 7.480 → 10.670 | +42.65% |

### Total process time

These times include model startup and cleanup as well as prefill/generation. A negative change means less elapsed time.

| Workload | Median #1047 → #1056 (s) | Time change |
| --- | ---: | ---: |
| Short, MTP off | 20.496 → 19.349 | -5.59% |
| Short, MTP on | 20.711 → 16.832 | -18.73% |
| Long, MTP off | 24.307 → 18.313 | -24.66% |
| Long, MTP on | 26.077 → 18.645 | -28.50% |

### What changed

- Repeated IQ2 gate/up, Q2_K down and HC input loads become reuse across rows/streams; sparse prefill batches use routing-aware narrower tiles and type-specialized kernels on supported M1 Max shapes.
- The initial decode gate/up overlap in #1047 extends through cached down projections, batched MTP verification and the prefill gate/up-to-down read pipeline.
- Expert-buffer reuse plans batch evictions instead of repeatedly scanning the complete cache.
- MTP prepares its historical predictor cache, retains the pending hidden row across rollback/restore, skips unused projections, and computes only the consumed final predictor MoE row.
- Compact MTP staging reserves more of the same memory budget for the automatic expert cache. Therefore these results measure the complete PR, including memory/cache policy and predictor-state corrections; they do not isolate kernel arithmetic.

### Cache and SSD work

Medians from the per-process streaming counters. `pread` bytes include filesystem-cache reads and are not physical SSD traffic.

| Workload | Expert-cache slots #1047 → #1056 | Expert cache misses #1047 → #1056 | Expert reads #1047 → #1056 (GiB) |
| --- | ---: | ---: | ---: |
| Short, MTP off | 8212 → 8212 | 9221 → 9221 | 12.80 → 12.80 |
| Short, MTP on | 7371 → 8212 | 9719 → 9593 | 15.75 → 14.85 |
| Long, MTP off | 8212 → 8212 | 19655 → 19655 | 27.27 → 27.27 |
| Long, MTP on | 7371 → 8212 | 22282 → 20135 | 32.06 → 28.77 |

With MTP enabled, staging falls from 2.59 to 1.42 GiB while the total planned memory budget remains 19.84 GiB in both versions. Without MTP, both versions have the same 8,212 cache slots and the same expert-read volumes: the ordinary-prefill speedup does not depend on a larger expert cache.

MTP counters were repeatable within each version. Short: 57 → 57 verifier cycles and 42/57 → 43/57 first drafts accepted (73.7% → 75.4%). Long: 33 → 32 verifier cycles and 15/33 → 18/32 accepted (45.5% → 56.2%). These changes are included in the end-to-end result; they are not an isolated measure of the predictor-state fix.

### Output checks and limits

All 32 measured runs exited successfully. Output was byte-identical across versions, repetitions and MTP settings for each prompt.
This checks generated-text reproducibility for these workloads; it does not establish bitwise equality of every intermediate tensor or broader model quality. No code or precision change was made for this measurement.

Reported values are medians of four observations; raw results are below. Ordinary-decode ranges overlap: short 7.10–7.69 → 7.53–8.36 t/s, long 9.01–9.66 → 9.64–10.69 t/s. These ranges are minima/maxima, not confidence intervals. The sample is small, filesystem caches were warm/uncontrolled, and desktop/background activity was not controlled. Generation rates are the CLI end-to-end rates, not isolated kernel timings. CUDA, ROCm, other Apple chips, cold-disk operation and other context lengths are not covered.

<details>
<summary>Raw benchmark results — all 32 measured runs</summary>

Rows are in execution order; rates exclude model startup while process time includes it.

| Run | Prompt | MTP | Version | Prefill t/s | Decode t/s | Process time (s) |
| ---: | --- | --- | --- | ---: | ---: | ---: |
| 1 | Short | On | #1047 | 9.07 | 7.44 | 21.271 |
| 2 | Short | On | #1056 | 10.84 | 9.66 | 17.343 |
| 3 | Short | On | #1056 | 10.79 | 9.79 | 13.998 |
| 4 | Short | On | #1047 | 9.21 | 7.45 | 20.881 |
| 5 | Long | On | #1047 | 34.02 | 7.21 | 24.895 |
| 6 | Long | On | #1056 | 50.85 | 10.57 | 20.360 |
| 7 | Long | On | #1056 | 52.66 | 10.47 | 16.831 |
| 8 | Long | On | #1047 | 35.28 | 7.48 | 27.260 |
| 9 | Short | On | #1056 | 10.84 | 10.16 | 16.784 |
| 10 | Short | On | #1047 | 9.19 | 7.68 | 20.542 |
| 11 | Short | On | #1047 | 9.54 | 7.51 | 17.503 |
| 12 | Short | On | #1056 | 10.76 | 10.08 | 16.881 |
| 13 | Long | On | #1056 | 50.52 | 10.77 | 17.220 |
| 14 | Long | On | #1047 | 34.01 | 7.48 | 27.891 |
| 15 | Long | On | #1047 | 35.63 | 7.57 | 24.015 |
| 16 | Long | On | #1056 | 51.64 | 11.07 | 20.070 |
| 17 | Short | Off | #1047 | 8.98 | 7.44 | 21.020 |
| 18 | Short | Off | #1056 | 10.68 | 8.36 | 19.198 |
| 19 | Short | Off | #1056 | 10.88 | 8.19 | 16.009 |
| 20 | Short | Off | #1047 | 9.10 | 7.68 | 20.674 |
| 21 | Long | Off | #1047 | 35.82 | 9.39 | 22.541 |
| 22 | Long | Off | #1056 | 56.01 | 10.69 | 19.404 |
| 23 | Long | Off | #1056 | 51.71 | 10.00 | 17.223 |
| 24 | Long | Off | #1047 | 35.28 | 9.44 | 26.073 |
| 25 | Short | Off | #1056 | 10.77 | 8.00 | 19.501 |
| 26 | Short | Off | #1047 | 9.34 | 7.69 | 20.318 |
| 27 | Short | Off | #1047 | 9.24 | 7.10 | 18.421 |
| 28 | Short | Off | #1056 | 10.63 | 7.53 | 20.558 |
| 29 | Long | Off | #1056 | 54.35 | 9.64 | 16.856 |
| 30 | Long | Off | #1047 | 33.77 | 9.01 | 26.928 |
| 31 | Long | Off | #1047 | 36.01 | 9.66 | 22.418 |
| 32 | Long | Off | #1056 | 53.35 | 9.72 | 20.430 |

</details>

<details>
<summary>Reproduction commands and exact long prompt</summary>

Build each pinned revision in a separate checkout and run from its own directory, so its Metal sources match its executable. For the short prompt:

```sh
DS4_METAL_STREAMING_EXPERT_TIMING_SUMMARY=1 ./ds4 --metal \
  -m /path/to/Qwen3.8-Flash-Next-Q2.gguf \
  --ssd-streaming --ctx 4096 --prefill-chunk 128 \
  --temp 0 --nothink --mtp-timing -n 100 \
  -p "narrami la storia di roma"
```

Omit `--mtp-timing` for ordinary decoding. For the long case, replace `-p ...` with `--prompt-file prompt-long.txt` and use `-n 50`. The prompt below is kept in its original Italian so the benchmark input is unchanged.

```text
Vorrei capire come Roma sia riuscita a governare territori molto diversi per un periodo così lungo, senza ridurre la spiegazione alla forza delle legioni o al talento di pochi imperatori. Scrivi un saggio in italiano che colleghi istituzioni, economia e organizzazione militare, seguendo il passaggio dalla tarda repubblica al principato e arrivando alle trasformazioni del terzo secolo.

Parti dai rapporti tra senato, magistrature, assemblee e comandanti. Spiega come le conquiste abbiano modificato gli equilibri politici e perché strumenti pensati per una città potessero incontrare difficoltà amministrando un dominio mediterraneo. Usa le vicende dei Gracchi, di Cesare e di Augusto come esempi da discutere, evitando di presentare la nascita dell'impero come un esito inevitabile.

Passa poi alla vita materiale. Mi interessano il ruolo dei contadini, il lavoro servile, la proprietà fondiaria, le imposte provinciali e gli approvvigionamenti delle città. Collega il commercio marittimo alle strade, ai porti e alla circolazione monetaria. Chiarisci quali gruppi potevano beneficiare dell'espansione e quali sostenerne maggiormente i costi. Distingui, quando serve, la situazione dell'Italia da quella delle province orientali o delle regioni di frontiera.

Dedica una parte all'esercito come istituzione sociale: reclutamento, paga, pensionamento, insediamento dei veterani e relazioni con le comunità locali. Vorrei capire anche come la fedeltà dei soldati ai comandanti potesse entrare in tensione con l'autorità politica e quanto il mantenimento delle truppe dipendesse dalla capacità fiscale dello Stato. Evita spiegazioni che attribuiscano ogni crisi soltanto alle invasioni.

Confronta inoltre il sistema romano con quello delle città greche e, con prudenza, con uno Stato contemporaneo. Indica sia analogie utili sia differenze che rendono ingannevoli certi paragoni, soprattutto per cittadinanza, rappresentanza, burocrazia e distribuzione della ricchezza. Mostra come l'integrazione delle élite locali potesse convivere con forti disuguaglianze e con forme diverse di resistenza.

Nel ragionamento considera la qualità delle fonti. Che cosa possiamo ricavare dagli storici antichi, dalle iscrizioni, dalle monete e dai reperti archeologici? Segnala il punto di vista sociale degli autori e distingui i dati relativamente solidi dalle ricostruzioni controverse. Non inventare citazioni letterali o statistiche precise. Concludi valutando due fattori che favorirono la stabilità e due tensioni strutturali che resero più difficile conservarla, collegandoli agli esempi già discussi.
```

Prompt file SHA-256: `5da9ad53318d2cd4d5eaa7dcc644cf47b8567ceb8a7384a2b6ac918cba10270f`.

</details>

GiorgioOppo:
Update: [2d6a207](https://github.com/GiorgioOppo/ds4/commit/2d6a207bbb6e3e18810865988e63797fb98842a9) separates **prefill projection policy from decode/batch projection policy** to substantially reduce the numerical drift observed against [e37f185](https://github.com/antirez/ds4/pull/1056/commits/e37f18576cd0c06217b0f1745021f1fe16f665cc).

### What was corrected

Short prompt chunks and tails could select projection paths intended for decode batches. The newer float-operand F16/Q8 tiles and split-K FP32 reductions changed the prefill arithmetic; changing the prefill chunk size could therefore change subsequent logits.

The graph now selects the phase explicitly. Contiguous prefill, including short tails, retains the reference operand precision and unsplit dense reductions. Ordinary decode, independent session batches, MTP verification and predictor operations retain their optimized dispatch. Phase state is restored on both success and failure. This is a dispatch correction; it does not rewrite the decode kernels. CUDA projection dispatch and the separate vision encoder retain their existing behavior.

### Logit comparison against e37f185

Before = `0024ca0`; after = `2d6a207`. Apple M1 Max, 32 GiB, `Qwen3.8-Flash-Next-Q2.gguf`, Metal SSD streaming, **1,024 cached experts**, context 8,192, MTP disabled, Metal API validation enabled.

Separate binaries use their own revision's shader sources. The reference records exact prompt IDs and 16 greedy continuation IDs; both versions consume those same IDs. Every comparison covers all 248,320 FP32 vocabulary logits, rather than generated text alone.

Errors below are maximum absolute differences from **e37f185**, in raw logit units. Decode takes the worst of 16 teacher-forced steps and includes any prefill-state difference.

| Prompt | Tokens / chunk | Prefill before | Prefill after | Decode before | Decode after |
|---|---:|---:|---:|---:|---:|
| Rome, CLI system prompt | 29 / 128 | 0.127451301 | 0.000001907 | 0.275259018 | 0.000367641 |
| Rome, no system prompt | 19 / 128 | 0.017433405 | 0.000001907 | 0.031987667 | 0.000383437 |
| I promessi sposi | 575 / 128 | 0.422063470 | 0.000001907 | 1.376746650 | 0.000319958 |
| I promessi sposi | 5,760 / 128 | 0.487301350 | 0.000001431 | 1.001465320 | 0.000828743 |
| I promessi sposi | 5,760 / 2,048 | 0.179281831 | 0.000001431 | 0.433784008 | 0.000828743 |

**85 frontiers and 21,107,200 logit pairs checked.** Both the before and after default configurations select the reference argmax at every measured frontier. The change substantially reduces numerical drift; it does **not** establish bit-exact parity with e37f185 or prove that broader quality regressions are resolved.

For the 5,760-token fixture, the patched chunk-128 and chunk-2,048 runs produce **bit-identical logits at all 17 frontiers**, including subsequent decode. Before this correction, those two chunk settings differed by up to 0.45611 after prefill and 1.07523 during decode.

### Regression coverage

- Builds, memory-planner tests and the Metal Qwen kernel suite pass.
- New `--prefill-reference` generation test: automatic session logits agree bit-for-bit with diagnostic legacy/unsplit direct prefill followed by automatic decode. One-shot generation also emits the same tokens. Three cases cover 29 tokens, a 575-token prompt with a 63-token tail, and two complete 128-token chunks; all 27 vocabulary frontiers pass.
- Negative control: the same test compiled against original `0024ca0` fails at its first frontier, confirming that it catches the regression.
- MTP cache, predictor, verification, checkpoint and continuation tests pass under their existing exactness/tolerance checks.

To run the new real-model regression test from the checkout:

```sh
make -j4 tests/test_qwen4_generation
./tests/test_qwen4_generation /path/to/Qwen3.8-Flash-Next-Q2.gguf \
  --ssd-streaming --prefill-reference
```

### Performance impact of this correction

This comparison is **0024ca0 → 2d6a207**, not #1047 → #1056. It uses different prompts/context/cache settings from the earlier comment, so the absolute rates should not be compared with that earlier benchmark.

Sixteen sequential runs: four repetitions per version/workload in ABBA/BAAB order, identical teacher-forced tokens, MTP off, fixed 1,024-expert cache, and Metal validation disabled. Model loading and extra logit analysis are excluded; normal synchronous public-API work remains included. Filesystem caches were not purged.

| Prompt / chunk | Median prefill t/s | Change | Median decode t/s | Change |
|---|---:|---:|---:|---:|
| 575 / 128 | 34.931 → 34.682 | −0.71% | 5.402 → 5.487 | +1.59% |
| 5,760 / 2,048 | 141.779 → 145.376 | +2.54% | 5.013 → 5.420 | +8.12% |

The sampled medians show no substantial throughput penalty. Repetition ranges overlap, and decode covers only 16 steps per run. In particular, **+8.12% is a preliminary whole-model observation, not evidence of a faster decode kernel**: corrected prefill state can affect expert routing/cache behavior, and SSD/system variability remains. For that case, decode ranges are 4.650–5.470 before and 5.197–5.921 t/s after.

<details>
<summary>Raw timing results — all 16 runs</summary>

Rows are in execution order. A = 0024ca0; B = 2d6a207.

| Run | Prompt / chunk | Version | Prefill t/s | Decode t/s |
|---:|---|---|---:|---:|
| 1 | 575 / 128 | A | 35.520373 | 5.506170 |
| 2 | 575 / 128 | B | 35.085339 | 5.419560 |
| 3 | 575 / 128 | B | 34.517489 | 5.126633 |
| 4 | 575 / 128 | A | 34.832407 | 5.579655 |
| 5 | 575 / 128 | B | 34.845904 | 5.554857 |
| 6 | 575 / 128 | A | 35.029069 | 5.296926 |
| 7 | 575 / 128 | A | 31.799975 | 4.906440 |
| 8 | 575 / 128 | B | 33.262266 | 5.732392 |
| 9 | 5,760 / 2,048 | A | 148.465994 | 5.010698 |
| 10 | 5,760 / 2,048 | B | 147.327766 | 5.260550 |
| 11 | 5,760 / 2,048 | B | 143.424677 | 5.197421 |
| 12 | 5,760 / 2,048 | A | 131.998468 | 5.015202 |
| 13 | 5,760 / 2,048 | B | 129.355412 | 5.920792 |
| 14 | 5,760 / 2,048 | A | 138.479145 | 5.469539 |
| 15 | 5,760 / 2,048 | A | 145.079537 | 4.650052 |
| 16 | 5,760 / 2,048 | B | 150.099421 | 5.579721 |

</details>

These checks cover the local Metal SSD configuration. They do not establish resident-mode parity, CUDA/ROCm GPU behavior, long-generation stability, or perplexity/quality improvements.

GiorgioOppo:
@slycrel, I should have fixed the drift in the tokens, could you try to benchmark again?

GiorgioOppo:
Follow-up throughput benchmark of [2d6a207](https://github.com/GiorgioOppo/ds4/commit/2d6a207bbb6e3e18810865988e63797fb98842a9), using the repository's **`ds4-bench` with 128 decode evaluations per run**.

This supplements the [previous numerical validation and short timing comparison](https://github.com/antirez/ds4/pull/1056#issuecomment-5729873052). The measurements below cover the current commit only; they are not a before/after speedup comparison.

### Configuration

- Apple M1 Max, 32 GiB; `Qwen3.8-Flash-Next-Q2.gguf`.
- Metal SSD streaming, **1,024 cached experts**, allocated context **8,192**, prefill chunk **128**, MTP disabled.
- Input: `speed-bench/promessi_sposi.txt`. Each run evaluates the following **128 fixed continuation tokens** with `--teacher-forced-decode`; this is not free-running chat generation.
- Six serial runs, two per prefix, in order **512, 2048, 4096, 4096, 2048, 512**. Each run starts a new process/session and measures one complete prefix from scratch, using equal `--ctx-start` and `--ctx-max` values. Prefill is therefore not an incremental suffix measurement.
- No concurrent model/server process at run startup. Inherited `DS4_*` overrides and Metal API validation were cleared. Normal SSD startup behavior was retained; filesystem caches were not purged. Model startup is excluded from throughput timings.

### Results

Medians of two observations, with observed minimum–maximum in parentheses:

| Full prefix tokens | Prefill t/s | Decode t/s |
|---:|---:|---:|
| 512 | **38.455** (37.55–39.36) | **6.815** (6.55–7.08) |
| 2,048 | **39.275** (39.21–39.34) | **6.205** (6.09–6.32) |
| 4,096 | **37.935** (37.82–38.05) | **6.460** (6.26–6.66) |

All six runs exited successfully and completed exactly 128 decode evaluations: **768 evaluations total**. Logs confirm a 1.42 GiB expert cache and approximately 10.00 GiB planned memory for this configuration.

<details>
<summary>Raw results — all six runs, in execution order</summary>

| Run | Prefix tokens | Prefill t/s | Decode t/s |
|---:|---:|---:|---:|
| 1 | 512 | 37.55 | 6.55 |
| 2 | 2,048 | 39.34 | 6.09 |
| 3 | 4,096 | 37.82 | 6.66 |
| 4 | 4,096 | 38.05 | 6.26 |
| 5 | 2,048 | 39.21 | 6.32 |
| 6 | 512 | 39.36 | 7.08 |

</details>

### Reproduction

From the checkout at `2d6a207`, for the 2,048-token case:

```sh
make -j4 ds4-bench
./ds4-bench --metal -m /path/to/Qwen3.8-Flash-Next-Q2.gguf \
  --ssd-streaming --ssd-streaming-cache-experts 1024 \
  --ctx-alloc 8192 --prefill-chunk 128 \
  --ctx-start 2048 --ctx-max 2048 --gen-tokens 128 \
  --teacher-forced-decode \
  --csv benchmark.csv
```

Set both frontier arguments to 512 or 4096 for the other cases, keeping the allocated context and other settings fixed. Prompt-file SHA-256: `f53e0d80cb2d4492d24ebd63c7000c397b16ae70f9bf09b3763e5d8323ec209f`.

Two repetitions per configuration provide indicative local rates and ranges, not confidence intervals. These tests do not establish a speedup, model quality, or logit parity. Their prompts/continuations differ from the previous 16-step tests, and chunk 128 is used throughout, so their absolute rates are not directly comparable with those earlier results. Resident mode, automatic/larger expert caches, CUDA and ROCm are outside this measurement.

slycrel:
Re-ran everything on the M1 Max **64 GB** against `04c0867`. Base is still `antirez/main` at `8db1d1d`, separate worktrees and binaries, `--temp 0 --nothink`, ABBA/BAAB order, 4 observations per arm per cell (2 for the deep and streaming cells), medians.

**First: I have to retract both caveats from my previous comment. Both were my measurement error, and `0024ca0` is what let me find it.**

Your note that upstream one-shot ignores the CLI prefill chunk is exactly right, and it invalidated my two negative findings. `ds4_engine_generate_argmax` on `8db1d1d` calls `qwen4_prefill_chunk_tokens()`, which reads `DS4_QWEN4_PREFILL_CHUNK` or defaults to 8192, so my `--prefill-chunk 128` runs on base were really running at 8192 while this PR honoured 128. Control, same base binary, same CLI flag, long prompt:

| base `8db1d1d`, `--prefill-chunk 128` | Prefill t/s | output sha |
|---|---:|---|
| `DS4_QWEN4_PREFILL_CHUNK` unset (really 8192) | 267.5 | `6bd92fcf` |
| `DS4_QWEN4_PREFILL_CHUNK=128` (really 128) | 125.7 | `ef58e1d1` |

So "prefill −52% at chunk 128" was 8192-vs-128, not base-vs-PR. **With the chunk pinned in both arms, this PR is faster at chunk 128: 127.5 → 136.6 t/s (+7%).** And the output hash I reported as "PR differs from main" at chunk 128, `ef58e1d1`, is exactly what base produces once it actually runs at 128. My apologies for the noise.

Everything below pins `DS4_QWEN4_PREFILL_CHUNK` to the same value as `--prefill-chunk` in **both** arms.

### Resident mode, base → PR, ctx 16384, chunk 2048

| Workload | MTP | Prefill t/s | Decode t/s | Decode range base / PR | MTP accepted |
|---|---|---:|---:|---|---|
| short (29 tok) | off | 66 → 75 | 26.14 → 30.55 **+17%** | 25.1–26.6 / 28.5–30.7 | |
| short | on | 57 → 72 | 31.16 → 34.36 **+10%** | 31.1–31.3 / 34.1–34.5 | 42/57 → 43/57 |
| roma575 | off | 230.9 → 245.2 **+6%** | 26.54 → 30.59 **+15%** | 26.4–26.6 / 30.3–30.7 | |
| roma575 | on | 229.9 → 230.3 +0.2% | 26.17 → 30.49 **+17%** | 25.3–26.3 / 30.2–30.9 | 16/33 → 18/32 |
| long (5,760) | off | 274.4 → 288.0 **+5%** | 24.96 → 28.57 **+14%** | 24.3–25.0 / 28.2–28.7 | |
| long | on | 273.5 → 259.0 **−5.3%** | 27.24 → 31.75 **+17%** | 26.3–27.5 / 31.2–32.1 | 52/72 → 53/67 |
| deep (17,408), ctx 32768 | off | 271.4 → 285.0 **+5%** | 23.39 → 26.80 **+15%** | 22.3–24.5 / 26.5–27.1 | |
| deep | on | 272.6 → 265.9 −2.5% | 24.76 → 28.41 **+15%** | 24.1–25.4 / 28.2–28.6 | 47/79 → 51/72 |

Decode ranges don't overlap in any cell. Prefill is now **positive** on every ordinary-decode cell (+5 to +6% on the real prompts) where I previously measured it flat or negative; the chunk mismatch had been masking that too. The one consistent cost is **prefill with MTP on: −5.3% long, −2.5% deep**, which looks like the reference projection path in `2d6a207`. Short-prompt prefill numbers (29 tokens) are noise, ignore them. MTP acceptance improves in every cell (72.2% → 79.1% long, 59.5% → 70.8% deep).

### The drift fix: measured against an unchunked reference

Text output still differs from `8db1d1d` on the 5,760-token prompt, so I tried to characterise it rather than just diff bytes. Using `--dump-logits` on the first token after the full prompt, and taking **prefill in a single unchunked pass as the reference** (`--prefill-chunk 8192` ≥ 5,760 tokens, so there are no chunk boundaries at all):

| Configuration | max abs logit delta vs unchunked reference |
|---|---:|
| base `8db1d1d` @ 8192 (1 chunk) | 0 (this is the reference) |
| **PR `04c0867` @ 8192 (1 chunk)** | **0.000000 — bit-identical to base** |
| PR @ 2048 (3 chunks) | 0.424518 |
| PR @ 128 (45 chunks) | 0.424518 |
| base @ 2048 (3 chunks) | 0.594551 |
| base @ 128 (45 chunks) | 1.132671 |

Two things fall out of this:

1. **Unchunked, the PR and main agree bit-for-bit.** So the series introduces no arithmetic change in the unchunked path; the whole remaining difference is chunk-boundary behaviour.
2. **`8db1d1d`'s own prefill is chunk-dependent, by more than the PR differs from it.** Main spans 1.13 logits between its own chunk 8192 and chunk 128 results. The PR spans **0.000000** between chunk 2048 and chunk 128 — `2d6a207` makes chunked prefill chunk-invariant, which main is not — and sits 0.42 from unchunked at every chunk size, i.e. closer to the unchunked reference than main is at either chunk size.

So against the reference that has no chunk boundaries, this PR is both closer and self-consistent. I could not get a CPU ground truth to arbitrate absolutely; `--metal-graph-prompt-test` doesn't finish in any reasonable time on the 41.72 GiB Q2 model here, so "unchunked GPU prefill" is the best reference I have. The residual 0.42 between your chunked and unchunked paths is the one thing I'd still flag as open.

For completeness, with both arms at chunk 2048 over 128 greedy steps the two continuations separate at step 33 on a 0.044-logit near-tie, after which the contexts differ and per-step logit comparison stops meaning anything. Short and roma575 outputs are byte-identical between arms at every setting.

### Tests
All pass here: `test-qwen4-memory`, `test-qwen4-kernels`, `test-qwen4-q2`, `test-metal-q8-reduction`, **`test-metal-q8-gemv-reference`** (new), `test-metal-qwen4-hc`, `test-metal-qwen4-moe-mid`, `test-metal-qwen4-moe-down`, `test-qwen4-ssd-experts`, `test-metal-ssd-reuse`, `test-qwen4-moe-mm-compact`; `tests/test_qwen4_mtp_prefill` with the real GGUF resident and `--ssd-streaming`; and **`tests/test_qwen4_generation`** both normally and with `--prefill-reference` — 9 complete vocabulary frontiers bit-exact in each of the six configurations, including the 575-token/63-token-tail and two-complete-chunk cases. Build is warning-free.

### Streaming on 64 GB (PR only, main rejects `--ssd-streaming` for Qwen)
ctx 4096, chunk 128: roma575 decode 20.0 t/s with MTP (18/32 accepted), 18.0 without; short 13.9 / 11.2 t/s; prefill 97–119 t/s on roma575. Auto cache sizes to 24,576 experts (34.10 GiB) here vs 8,212 on your 32 GiB box. Resident is still much faster on this machine, as expected.

GiorgioOppo:
@slycrel  Thanks for trying again, I'm glad you had a chance to try again. Thanks to your testing, I was able to find another small bug that I'll fix in the next commit.

GiorgioOppo:
@slycrel could you try one last time with the latest optimizations?

slycrel:
Re-ran against `11811b9` on the M1 Max **64 GB**. Base is still `antirez/main` at `8db1d1d`, separate worktrees and binaries, `--temp 0 --nothink`, `DS4_QWEN4_PREFILL_CHUNK` pinned to the same value as `--prefill-chunk` in **both** arms, ABBA/BAAB, 4 observations per arm per cell (2 for deep/streaming/chunk-128), medians.

## `89986c3` closes the chunked-vs-unchunked gap completely

This was the one open item I left last time (0.42 logits between your chunked and unchunked paths), and it is gone. Using `--dump-logits` on the first token after the whole prompt, with **prefill in a single unchunked pass as the reference** — 5,760-token prompt, `--prefill-chunk 8192`, so no chunk boundaries exist:

| Configuration | max abs logit delta vs unchunked reference |
|---|---:|
| base `8db1d1d` @ 8192 (1 pass) | 0 — this is the reference |
| base @ 2048 (3 chunks) | 0.594551 |
| base @ 128 (45 chunks) | 1.132671 |
| **PR `11811b9` @ 8192 (1 pass)** | **0.000000** |
| **PR @ 2048 (3 chunks)** | **0.000000** |
| **PR @ 128 (45 chunks)** | **0.000000** |

Every chunk size on this PR is now bit-identical to the unchunked reference. Main still spans 1.13 logits across its own chunk sizes. The same thing shows up at text level without any logit tooling: on the long prompt the PR emits main's **single-pass** output byte-for-byte at both chunk 2048 and chunk 128 (`6bd92fcf`), while base at chunk 128 emits something different (`ef58e1d1`). So what I previously reported as "PR differs from main" is now unambiguously main's own chunk-dependence, and this PR is the arm that is self-consistent.

The 17,408-token prompt confirms invariance but cannot confirm the absolute value: **PR spread across chunks 8192 / 4096 / 2048 / 128 = 0.000000**, base spread = 1.099256, PR sits 0.756354 from base@8192. I could not get a single-pass reference at that length to arbitrate — `--prefill-chunk` ≥ 17,408 dies with `kIOGPUCommandBufferCallbackErrorOutOfMemory` on **base as well**, so that is a 64 GB limit here, not something in your series.

SSD streaming is chunk-invariant too: identical output at chunk 2048 and 8192.

## Performance, resident, base → PR

| Workload | MTP | Prefill t/s | Decode t/s | Decode range base / PR | MTP accepted |
|---|---|---:|---:|---|---|
| roma575 | off | 232.1 → 246.3 **+6.1%** | 26.04 → 30.00 **+15.2%** | 25.9–26.5 / 29.8–30.2 | |
| roma575 | on | 232.0 → 231.4 −0.2% | 26.20 → 30.25 **+15.5%** | 26.1–26.4 / 30.2–30.8 | 16/33 → 18/32 |
| long (5,760) | off | 274.8 → 289.2 **+5.2%** | 24.48 → 27.94 **+14.1%** | 24.5–24.6 / 27.9–28.0 | |
| long | on | 274.8 → 269.3 **−2.0%** | 28.12 → 32.51 **+15.6%** | 28.1–28.2 / 32.4–32.7 | 52/72 → **56/67** |
| long, chunk 128 | off | 128.5 → 137.3 **+6.9%** | 24.55 → 27.92 **+13.7%** | 24.5–24.7 / 27.9–28.0 | |
| deep (17,408), ctx 32768 | off | 273.9 → 287.4 **+4.9%** | 23.93 → 27.16 **+13.5%** | 23.9–24.0 / 27.2 | |
| deep | on | 273.6 → 267.6 −2.2% | 25.09 → 27.02 **+7.7%** | 25.0–25.2 / 27.0–27.1 | 47/79 → 46/78 |
| short (29 tok) | off | noise | 26.10 → 29.91 **+14.6%** | 26.0–26.4 / 29.9–30.0 | |
| short | on | noise | 30.87 → 34.02 **+10.2%** | 30.8–31.2 / 33.9–34.4 | 42/57 → 43/57 |

Decode ranges don't overlap anywhere. The MTP-on prefill cost I flagged last round has roughly halved: long went **−5.3% → −2.0%**, deep −2.5% → −2.2%.

### One thing to look at: deep-prompt MTP acceptance moved backwards since `04c0867`

Deterministic counters, identical in both runs of each arm, same pinned chunk:

| 17,408-token prompt, MTP on | acceptance | decode t/s |
|---|---:|---:|
| base `8db1d1d` | 47/79 (59.5%) | 25.09 |
| PR `04c0867` | 51/72 (**70.8%**) | 28.41 |
| PR `11811b9` | 46/78 (**59.0%**) | 27.02 |

So the deep MTP-on decode gain fell from +14.8% to +7.7%, and acceptance is back at base's level. The 5,760-token prompt moved the other way over the same commits (53/67 → **56/67**, 79.1% → 83.6%). Since `89986c3` legitimately changes the attention arithmetic at the sparse boundary, drafts and verifications now agree differently, so this may simply be the honest number rather than a regression — but deep + MTP is now the weakest cell in the matrix and you may want to look at whether the predictor path benefits from the same treatment.

## `make test-qwen4-moe-mm-compact` now fails

```
AssertionError: kernel void kernel_qwen4_moe_mm_mid(
  tests/test_qwen4_moe_mm_compact.py:88
  assert body.count("qwen4_moe_mm_expert(args, tgpig.y)") == 1
```

`11811b9` refactored `kernel_qwen4_moe_mm_mid` into a thin wrapper that delegates to the new `qwen4_moe_mm_mid_impl<NT, float, false>`, so the call the scanner looks for is no longer inside the kernel body. **I checked the logic is intact** — `qwen4_moe_mm_expert(args, tgpig.y)` and its `counts[e]` ordering guard both live in `qwen4_moe_mm_mid_impl` now, and the other three markers (`_down`, `_mid_nax_t`, `_down_nax_t`) still match — so this is a stale assertion rather than a real compaction regression. It passed at `04c0867`; `11811b9` is the only one of the three new commits that touches `metal/qwen4.metal`. The check probably needs to follow the wrapper into the impl.

## Everything else passes

`test-qwen4-memory`, `test-qwen4-kernels`, `test-qwen4-q2`, `test-metal-q8-reduction`, `test-metal-q8-gemv-reference`, **`test-metal-qwen4-moe-half`** (new), `test-metal-qwen4-hc`, `test-metal-qwen4-moe-mid`, `test-metal-qwen4-moe-down`, `test-qwen4-ssd-experts`, `test-metal-ssd-reuse`; `tests/test_qwen4_mtp_prefill` with the real GGUF resident and `--ssd-streaming`; `tests/test_qwen4_generation` both normally and with `--prefill-reference`. Build is warning-free.

## The `11811b9` half-operand path, on 64 GB

It only engages for SSD streaming at chunks ≥ 8192 and there is no opt-in flag, so I could not isolate it from the chunk change. Streaming, 5,760-token prompt, 3 observations each: prefill **238.7 t/s at chunk 8192 vs 224.3 at chunk 2048 (+6.4%)**, decode 16.2 vs 16.8 t/s with fully overlapping ranges — consistent with your own note that the isolated kernel gain does not carry into aggregate decode throughput. Output identical at both chunks.

GiorgioOppo:
Hi @antirez , this pr is ready for merge

slycrel:
Two results for `a6ad636`: a merge-readiness check on the M1 Max 64 GB, and the first run of this PR on M5+-class hardware, a **Mac mini M6 with 32 GB**.

Main had moved 25 commits since my last round (`8db1d1d` → `0aaea5a`), so both machines test **the actual merge result**, not the PR branch. `git merge a6ad636` onto `0aaea5a` is clean: no conflicts, a warning-free build, and the same tree hash on both machines. Settings throughout: `--temp 0 --nothink`, with `DS4_QWEN4_PREFILL_CHUNK` pinned to the same value as `--prefill-chunk`. Runs are in ABBA order, 2 observations per cell, and I report means.

## Merge readiness (M1 Max 64 GB): looks ready from here

- **Numerics:** chunk invariance still holds after the merge. At chunks 8192, 2048 and 128, the merge is **0.000000** from main's unchunked reference, while main's own spread across chunk sizes is 1.132671.
- **Tests:** all 12 Qwen/Metal targets pass, including `test-qwen4-moe-mm-compact`, which `fd0c24a` fixed. So do `test_qwen4_mtp_prefill` (resident and `--ssd-streaming`) and `test_qwen4_generation` (plain and `--prefill-reference`).
- **Full `make test`:** I ran it on both base and merge, with Qwen symlinked as `ds4flash.gguf`. The suite expects DeepSeek V4 Flash, so many sections fail on **both** trees. Every one of the 19 sections has **identical OK/ERR status** on both. The failure count moves 52 → 53, and that decomposes into:
  - −1 because main can't open a Qwen engine with SSD streaming, which this PR adds
  - +2 golden-vector token mismatches inside that newly running section
- **The merge kept both sides:** all 35 of main's added `ds4.c` lines are present, the PR's symbols are present, and there are no conflict markers.

Performance, resident, main `0aaea5a` → merge:

| Workload | MTP | Prefill t/s | Decode t/s |
|---|---|---:|---:|
| roma575 | off | 225.1 → 240.2 **+6.7%** | 26.26 → 30.03 **+14.4%** |
| long (5,760) | off | 263.1 → 276.7 **+5.2%** | 24.06 → 28.16 **+17.0%** |
| long | on | 257.7 → 257.8 **+0.0%** | 28.28 → 32.42 **+14.7%** |
| deep (17,408), ctx 32768 | on | 261.5 → 256.6 −1.9% | 24.59 → 26.87 **+9.3%** |

The MTP-on prefill cost on the long prompt is gone: it went −5.3% → −2.0% → +0.0% across rounds. Deep + MTP is still the weakest cell, with acceptance at 46/78 against main's 47/79. See the M6 note below.

**What I could not cover:**
- I don't have DeepSeek V4 or GLM weights, so the changes to shared files (`ds4.c`, `ds4_metal.m`, `metal/dense.metal`) are unverified at model level for those families.
- I can't test CUDA or ROCm.

## M6 32 GB: this PR is the only way to run Qwen3.8 on it

- **Resident** needs 41.72 GiB, so on 32 GB it fails with `kIOGPUCommandBufferCallbackErrorOutOfMemory`.
- **Main** rejects `--ssd-streaming` for Qwen (`… SSD streaming … are not supported`, exit 1).
- **With this PR** it runs in streaming mode with a 19.84 GiB plan: 6.32 GiB static weights plus 6,665–6,981 cached experts (about 9.5 GiB, against about 24K on 64 GB). Peak RSS is 18.3 GB, and a 17K-token prompt at `--ctx 32768` completes.

| Prompt | MTP | M6 streaming | M1 Max streaming | M1 Max resident |
|---|---|---:|---:|---:|
| | | prefill / decode t/s | prefill / decode t/s | prefill / decode t/s |
| roma575 | off | 63.3 / 13.20 | 100.9 / 18.84 | 247.3 / 30.16 |
| roma575 | on | 62.5 / 15.09 | 116.8 / 19.78 | 233.1 / 30.36 |
| long (5,760) | off | 150.2 / 11.41 | 231.3 / 18.02 | 290.0 / 27.87 |
| long | on | 144.5 / 12.37 | 218.0 / 21.80 | 270.5 / 32.76 |
| deep (~17K), ctx 32768 | off | 142.4 / 10.62 | 253.5 / 18.18 | 259.2 / 26.59 |
| deep | on | 149.3 / 13.32 | 236.7 / 18.68 | 268.2 / 27.02 |

- **The Metal 4 tensor path** (auto-enabled on M5+) is worth about **+20% prefill** on long prompts: 150 vs 122 t/s at 5,760 tokens and 142 vs 117 on deep, against `DS4_METAL_DISABLE_METAL4=1`. Decode is the same within noise.
- **On this 256 GB model, prefill is limited by SSD bandwidth, not cache size.**
  - At the M6's exact cache size (`--ssd-streaming-cache-experts 6665`), the M1 Max still prefills at 245–265 t/s.
  - `fio` random reads at 256 KiB–1 MiB: 2.9–3.0 GB/s on the M6 vs 6.1–6.8 GB/s on the M1 Max.
  - So M5/M6 machines with larger internal SSDs should do noticeably better.

**Deep-prompt MTP acceptance is higher on the M6:** 52/75 (69%) with the tensor path and 53/74 (72%) without, against 46/78 (59%) on the M1 Max. That's the same code, and the counters are deterministic and repeat identically. Since you added deep-prompt MTP diagnostics in `a6ad636`, this may be a useful lead: M6 arithmetic accepts noticeably more drafts on exactly the cell that lags.

**Chunk invariance holds on the tensor path.** First-token `--dump-logits` after 5,760 tokens, chunk 2048 vs 128: **0.000000** with the tensor path on, and 0.000000 with it off. This is the first check of the PR's numerics on M5+.

**Cross-device output differs in some configurations, but it's numerics, not a regression.** Greedy text on the M6 sometimes differs from the M1 Max, while every configuration repeats exactly on its own machine.
- Which M6 path matches the M1 Max depends on ctx and chunk. With the tensor path off, the M6 reproduces the M1 Max byte-for-byte at ctx 4096 / chunk 128. With it on, it matches on roma575 at ctx 16384 / chunk 2048. Neither path matches on the long and deep prompts.
- At 5,760 tokens, first-token logits differ from the M1 Max by up to 0.98–1.26, with the same argmax and a margin of about 1.7.
- It isn't cache-size dependent: the M1 Max gives identical output at 6,665, 12,000 and 23,616 cached experts.
- I haven't pinned down the mechanism. I'm noting it so M6 output differences aren't mistaken for a bug.

**Tests:** the same 12 targets pass on the M6 with the tensor path both on and off. So do `test_qwen4_mtp_prefill --ssd-streaming` and `test_qwen4_generation --ssd-streaming` (plain and `--prefill-reference`), which reports "9 complete vocabulary frontiers bit-exact". The resident variants can't run on 32 GB.

GiorgioOppo:
@slycrel Thanks for the test on m6; it was very useful for identifying a further improvement. It should run better now, as I’ve enabled expert calculations to run while the SSD loads the data. If you could try testing on m6 again, that would be very helpful.

GiorgioOppo:
@slycrel I also enabled the optimization for m5 on m6 as well.

Flor1an-B:
Tested this on Apple M5 Max, 128GB — resident mode (no SSD streaming, our actual deployment: `--ctx 8192 --prefill-chunk 4096 --batched-session 4`, real `Qwen3.8-Flash-Next-Q2.gguf`), a different regime from the M1 Max SSD-streaming numbers in the PR body.

- Full unit suite green, `make test-qwen4-kernels` all passed (extensive byte-exact assertions across HC norm reuse, GDN prefill, MTP stage/combine).
- Greedy correctness check: coherent, correct output, matches expectations.
- Speed (800-token code-gen completion, prod sampling temp=1.0/top_p=0.95/top_k=20/min_p=0), two runs for stability: **52.0 tok/s** both times, vs **50.6 tok/s** on current `main` without this PR — a modest but real and repeatable **~2.8% gain** in resident mode.

Makes sense given the PR's own framing: the kernel-reuse work (IQ2_XXS/Q2_K activation reuse) helps regardless of streaming vs resident, while the larger "up to 40%" comes from the SSD read/compute overlap, which doesn't apply when the model is fully resident. Just wanted to add a resident-mode data point on different hardware, since the existing numbers are all SSD-streaming/M1 Max — useful context for anyone on a larger Mac running this resident, like we do.

No correctness concerns found from our side.

PR #1058, opened 2026-09-15
Key tool-conversation KV checkpoints by the client-visible transcript

### LINKED ISSUE #1053: Shutdown checkpoint is never found on restart when the conversation began after another live conversation in the same process
**Build:** main `9139e2a`, Metal, M4 Max 64 GB, `Qwen3.8-Flash-Next-Q2.gguf` (self-contained), `--mtp -c 163840 --prefill-chunk 1024 --kv-disk-dir DIR --kv-disk-space-mb 16384`. Same result on ivanfioravanti/ds4-metal `6c1e836` (identical `ds4_kvstore.c`), so it predates the Qwen merge and is not specific to this model's checkpoint payload.

## Repro (10 minutes, two small OpenAI-compatible clients)

1. Conversation **A**: 4 chat turns, ~35–40K tokens.
2. Conversation **B**: 8 turns with tools, ~80K tokens. B's first request logs
   `live kv cache miss live=38403 prompt=11129 common=3 vision=match reason=token-mismatch`
   then `kv cache stored tokens=38403 ... reason=evict` (A goes to disk), and B is prefilled.
3. `SIGTERM` the server. Log: `persisting resident KV cache before shutdown slot=0 tokens=79188` → `kv cache stored tokens=79188 ... reason=shutdown key=token-text`. The file is on disk; its header SHA matches its text; the text starts with B's system prompt.
4. Restart with the same flags and `--trace`; resend B's exact history plus one user turn (~90K tokens).

**Result: full re-prefill.** Trace for that request:

```
memory_miss_reason: no-live-checkpoint
cache_source: none
cached_tokens: 0
disk_cached_tokens: 0
```

No `kv cache hit` and no `kv cache skipped` line; the lookup found nothing, not even B's smaller continued checkpoints.

## What does and does not matter (all tested)

| Variant | Restores after restart? |
|---|---|
| B alone in the process (no A), empty cache dir | **yes** (`cache_source: disk-text`, 80,091 tokens, ~0.3 s) |
| B alone, cache dir at budget (`--kv-disk-space-mb 3072`) | yes |
| A (with tools) then B (with tools) | **no** |
| A **without** tools, then B with tools | no |
| A with tools, then B **without** tools | no |

So tool calls, cache pressure and the eviction order of #444 are ruled out. The one condition that flips the result is whether B started **cold** (fresh process, live position 0) or **warm** (another conversation was live; B's first request matched a few common tokens of it and the rest was re-prefilled). In every warm case B's first continued checkpoint lands at 8,192 tokens instead of the cold case's 10,240, so B's session bookkeeping differs from the first turn on.

## Where I got to in the code (pointers, not a diagnosis)

- `ds4_kvstore_find_text_prefix` hashes the prompt prefix against every index entry (longest SHA1 match), so a miss means no stored text is a byte prefix of the replayed prompt.
- The stored text comes from `slot->thinking_live.visible_text` / `responses_live.visible_text` when that record is valid and `live_tokens == tokens->len` (`kv_cache_store_current`, ds4_server.c ~11275), otherwise from `ds4_kvstore_render_tokens_text` (detokenised). `token_text_disk_key` is set from `prompt_preserves_reasoning && !visible_continuation`.
- My guess is that after a warm start the stored text for B is produced on a different path than after a cold start (detokenised vs the client's rendered text, or with/without the thinking-visible handling), so its SHA1 can never equal the hash of the client's bytes. I did not verify which branch was taken; a log line naming the text source at store time would settle it in one run.

## Why it matters

A daily server always has other conversations behind it (cron jobs, side requests), so in practice the live conversation is never restored after a graceful restart; at 160K that is an ~8-minute first turn every time the service restarts.

Clients used: two ~150-line Python scripts (streaming, tools); happy to attach them or run a variant with extra logging.

emilianbold:
Fixes #1053.

## Problem

After a graceful shutdown, the disk KV checkpoint of a tool-using Chat
Completions conversation was never found on restart (`kv cache stored ...
reason=shutdown key=token-text`, then a full re-prefill). The report was on
Qwen, but the bug is generic: it reproduces on DeepSeek Flash and the same
code path serves GLM.

Root cause, in `ds4_server.c`:

1. `should_remember_thinking_checkpoint` refused to remember a client-visible
   key for any request with tools (or a prompt that preserves reasoning)
   unless the model syntax was Qwen. On DeepSeek/GLM no thinking-visible key
   ever existed for a tool conversation, so evict, shutdown and continued
   stores fell back to the detokenized session as key. That text embeds the
   sampled hidden reasoning, which OpenAI-compatible clients never send back
   as `reasoning_content`, so it can never byte-prefix-match a replayed
   prompt. The same mismatch made every later request in the conversation
   miss live KV at the previous assistant turn's `<think>` and re-prefill from
   the last disk checkpoint (up to 10K tokens behind).
2. Even with the gate opened, `build_thinking_visible_text` produced
   `<｜Assistant｜></think>content` for DeepSeek, while in tool context the
   renderer echoes earlier assistant turns as `<think>{reasoning}</think>content`,
   i.e. `<think></think>content` when the client omits reasoning. Qwen renders
   the same shape in both contexts, which is why the Qwen-only exception
   worked and the bug looked model-specific.
3. Tool-call turns had a visible checkpoint only for Qwen; DeepSeek/GLM
   tool-call turns cleared it, so the request carrying the tool result missed
   as well.
4. `continued` checkpoints written while prefilling turn N+1 of a
   live-continued session were keyed by tokens that include turn N's hidden
   reasoning, so they were dead after a restart too (the reporter's "not even
   B's smaller continued checkpoints").

## Fix

- Remember the visible key for tool / preserved-reasoning prompts on every
  syntax on Chat Completions (Responses and Anthropic keep their own live
  records). The existing `token_text_disk_key` behaviour is unchanged: the
  first disk key stays exact and switches to the visible key once a request
  actually continues without reasoning.
- DeepSeek and DeepSeek 4.1 build the visible key as `<think></think>content`
  when the prompt is in tool context (same predicate the renderer uses).
- Add a DSML tool-call turn visible checkpoint (DeepSeek, DeepSeek 4.1, GLM)
  next to the Qwen one. The key ends at the raw tool block, like Qwen, and a
  live-tail check refuses the shortcut if the sampled frontier carries extra
  bytes past the end marker.
- Key continued stores that cover a visible frontier by the visible text plus
  the detokenized prefill suffix, and re-arm the frontier after a
  thinking-visible disk restore so later continued stores stay matchable.

Behaviour note: tool conversations on DeepSeek/GLM now keep their own hidden
reasoning in KV across turns, as tool-less chats (and Qwen with tools) already
did. Outputs of tool conversations can therefore differ slightly from before,
where the model was silently re-conditioned each turn on `<think></think>`.

## Impact

Measured on DeepSeek Flash (Metal, M5 Max), a 12K-token tool conversation:

| Request | before | after |
|---|---|---|
| short follow-up / tool-result request | 815-877 tokens re-prefilled, 2.1-2.2 s | 12-17 tokens, 0.18-0.23 s |
| restart replay | 1905 tokens re-prefilled, 3.8 s (partial checkpoint hit) | 14 tokens, 0.22 s |
| disk files for the conversation | 9 files / 1270 MB, 5 unmatchable | 6 files / 827 MB, all matchable |

In-session, the fix removes an average of ~5K tokens of re-prefill per request
(half the 10240-token continued spacing); at 80-160K contexts that is tens of
seconds per request in an agent loop. On restart with no matchable checkpoint
(the issue's case) it replaces a full re-prefill, ~8 min at 160K per the
report, with the new turn only.

## Tests

- Unit (`./ds4_test --server`): `test_thinking_checkpoint_remember_gate`
  updated (fails on main); new `test_thinking_visible_text_deepseek_tool_context`
  checks the DeepSeek/DS41 key against the renderer in tool context (fails on
  main); new `test_dsml_tool_visible_checkpoint_boundary` covers the tool-turn
  key for DeepSeek, DeepSeek 4.1 and GLM.
- End-to-end (second commit): `make test-kv-restart-restore DS4_TEST_MODEL=...`
  runs `tests/kv_restart_restore_smoke.sh`: warm-up conversation, tool
  conversation with a client that omits `reasoning_content`, SIGTERM, restart,
  replay. Model-gated like the DSpark/GLM smoke scripts, skips without a model,
  ~6 min. On `9139e2a` it fails with the issue signature
  (`reason=shutdown key=token-text`); with this PR it restores 12235 of 12249
  tokens.

Not verified at runtime: Qwen (no local model; the continued-key change is
the part expected to matter there) and GLM (covered by the unit tests only).


aaa2015:
I ran your smoke test against GLM-5.3-Flash-Q2, since the PR description flags GLM as verified by unit tests only. Two things to report: the key logic holds up, and a separate bug in the GLM renderer was hiding it.

**Result with #1058 + #1059 on GLM-5.3-Flash-Q2**

```
run 1: turns hit the live visible checkpoint as expected
run 2 (after restart): FAIL — restored 8192 of 11992 tokens, 3800 re-prefilled
```

The shutdown checkpoint is written correctly:

```
kv cache stored tokens=11998 reason=shutdown key=thinking-visible
```

but the replay misses it and falls back to an older 8192-token `key=token-text` store. So the observable symptom matches the GLM caveat you noted.

**The cause is not in your code.** Dumping both sides of the lookup shows the visible key and the replayed prompt diverge by three bytes:

```
visiblekey: ...<|assistant|><think></think><tool_call>get_weather<arg_key>city...
reqprompt : ...<|assistant|><think></think>\n<tool_call>get_weather<arg_key>city...
                                    ^^
```

`append_glm_tool_calls_text()` adds a newline before and after the reconstructed tool block. GLM's `tokenizer.chat_template` emits it inline with no surrounding whitespace, and the model samples the inline form. When the raw sampled bytes are available (`raw_tool_text`) they are emitted verbatim, so the two renders only disagree on a history replayed by a client — which is precisely the path your checkpoint has to match. Your key is correct; the text it is being compared against is three bytes off, and everything after that offset shifts.

**With that render fixed (#1065):**

```
after:  OK,   restored 11998 of 12012 tokens (14 re-prefilled)
kv cache hit text tokens=11998 text=64224 key=thinking-visible
```

So on GLM the visible-key approach works as designed once the prompt it is matched against is faithful.

**Two notes on the smoke test itself**, which I had to work around before it would run on GLM at all: `start_server` hardcodes `--prefill-chunk 1024`, which GLM rejects outright (`--prefill-chunk is not supported for GLM 5.2`), and it does not set a cwd, so the server comes up with `Metal source metal/flash_attn.metal not found` and no Metal backend.

For reference on the other model: DeepSeek-4.1-Flash-Q2 goes from a baseline that stores no shutdown checkpoint at all (`ABORT: no shutdown checkpoint stored`) to `restored 12257 of 12271`, so the change is a clear win there regardless of the above.

cropduster:
Ran your smoke test here against a Qwen3.8 model and it fails **identically on the PR head and on its merge base**, so this looks like a second gap rather than a regression.

Setup: M4 Max 128 GB, macOS 27.0, `Qwen3.8-Flash-Next-Q4.gguf` (self-contained n-grams, i.e. the shipped Qwen3.8 release layout), `DS4_KV_RESTORE_CTX=32768 --prefill-chunk 1024 --kv-disk-space-mb 4096 --mtp`, running `tests/kv_restart_restore_smoke.sh` from your branch **unmodified**.

```
PR head    3b615fe: FAIL — restored 8192 of 12343 prompt tokens (4151 re-prefilled)
                    replay hit: kv cache hit text tokens=8192 ... key=token-text
merge base 9139e2a: FAIL — restored 8192 of 12343 prompt tokens (4151 re-prefilled)
                    replay hit: kv cache hit text tokens=8192 ... key=token-text
```

Same prompt/cached numbers and the same store file hash (`0a666a48…`) on both builds.

Your change is demonstrably active in the run, so this is not a build artifact:

- the tool turn now logs `tool-turn visible checkpoint remembered`, where the base logs `qwen tool-turn visible checkpoint remembered`;
- the graceful shutdown stores `reason=shutdown key=thinking-visible` (520.11 MiB) **on both builds** — so for Qwen the store side already carries a visible key.

What still misses is the restore/lookup side: the replay resumes from the 8192-token `token-text` grid store and re-prefills the remaining 4151 tokens instead of using the `thinking-visible` checkpoint that was just written.

Two questions:

1. Does the smoke test pass for you on a Qwen3.8 build at all, or is its expectation intentionally DeepSeek/GLM-shaped (the header says "DeepSeek Flash class model")?
2. If the lookup needs the same treatment as the store, would a diagnostic line naming the key each lookup compared — and why it was rejected — be welcome? I can add that instrumentation on a branch and re-run this exact test, plus a second scenario I have logs for (a Qwen tool conversation with a 30–44K-token hidden-reasoning turn that misses live KV in-process, where 42,541 tokens / 70.6 s were re-prefilled).

Raw logs from both runs are available — happy to attach them or paste the full `server1.log`/`server2.log`.

aaa2015:
Following up on my earlier note with the actual patch, since both issues block anyone pointing this harness at a non-DeepSeek model. It applies to `3b615fe`.

```diff
diff --git a/tests/kv_restart_restore_smoke.sh b/tests/kv_restart_restore_smoke.sh
index 076d7db..a34155a 100755
--- a/tests/kv_restart_restore_smoke.sh
+++ b/tests/kv_restart_restore_smoke.sh
@@ -25,6 +25,9 @@ Environment:
   DS4_TEST_MODEL=ds4flash.gguf
   DS4_KV_RESTORE_PORT=8199
   DS4_KV_RESTORE_CTX=32768
+  DS4_KV_RESTORE_PREFILL_CHUNK=1024 set empty to omit --prefill-chunk, needed
+                                    for syntaxes that reject it (GLM selects its
+                                    own graph prefill chunks)
   DS4_KV_RESTORE_EXTRA_ARGS=""      (extra ds4-server args, e.g. "--gpu-devices 0")
   DS4_KV_RESTORE_KEEP=1             keep the temp dir (logs, kv files)
 USAGE
@@ -36,7 +39,11 @@ model=${1:-${DS4_TEST_MODEL:-ds4flash.gguf}}
 port=${DS4_KV_RESTORE_PORT:-8199}
 ctx=${DS4_KV_RESTORE_CTX:-32768}
 extra=${DS4_KV_RESTORE_EXTRA_ARGS:-}
+# GLM 5.x rejects --prefill-chunk, so let callers omit it for those models.
+chunk=${DS4_KV_RESTORE_PREFILL_CHUNK-1024}
+if [ -n "$chunk" ]; then chunk_arg="--prefill-chunk $chunk"; else chunk_arg=""; fi
 here=$(cd "$(dirname "$0")" && pwd)
+root=$(cd "$here/.." && pwd)
 client="$here/kv_restart_restore_client.py"
 
 if [ ! -f "$model" ]; then
@@ -48,6 +55,19 @@ if [ ! -x "$bin" ]; then
     exit 1
 fi
 
+# The Metal backend resolves its kernels as metal/<name>.metal relative to the
+# working directory, so the server has to start from the repository root.
+# Make both paths absolute first: either may be relative to the caller, and the
+# checks above deliberately ran against the caller's view of them.
+case "$bin" in
+    /*) ;;
+    *) bin="$(pwd)/${bin#./}" ;;
+esac
+case "$model" in
+    /*) ;;
+    *) model="$(pwd)/$model" ;;
+esac
+
 tmpdir=$(mktemp -d "${TMPDIR:-/tmp}/ds4-kv-restart.XXXXXX")
 server_pid=""
 cleanup() {
@@ -82,9 +102,9 @@ abort() {
 start_server() {
     log=$1
     # shellcheck disable=SC2086
-    DS4_LOCK_FILE="$tmpdir/ds4.lock" "$bin" -m "$model" -c "$ctx" --port "$port" \
-        --prefill-chunk 1024 --kv-disk-dir "$tmpdir/kv" --kv-disk-space-mb 4096 \
-        $extra >"$log" 2>&1 &
+    ( cd "$root" && DS4_LOCK_FILE="$tmpdir/ds4.lock" exec "$bin" -m "$model" -c "$ctx" --port "$port" \
+        $chunk_arg --kv-disk-dir "$tmpdir/kv" --kv-disk-space-mb 4096 \
+        $extra ) >"$log" 2>&1 &
     server_pid=$!
     i=0
     while ! grep -q "listening on" "$log"; do
```

Notes on the shape:

* **`--prefill-chunk` stays at 1024 by default**, so DeepSeek runs and @cropduster's Qwen3.8 run are byte-for-byte unchanged. Only `DS4_KV_RESTORE_PREFILL_CHUNK=` empty drops the flag, which is what GLM needs.
* The `cd` goes in a subshell with `exec`, so `$!` still names the server and the existing `kill -TERM "$server_pid"` in `stop_server`/`cleanup` keeps working. No PID tracking has to change.
* Both paths are made absolute *after* the existing existence checks, so a bad `DS4_SERVER_BIN` still reports the path the caller typed.

Both behaviours are verified with a stub binary that echoes its cwd and argv into the run log — default gives `cwd=<repo root>` and `--prefill-chunk 1024`; with the variable set empty the flag is absent entirely. The cwd requirement itself I confirmed separately the way @cropduster describes it: `ds4_gpu_full_source()` tries `metal/<name>.metal` and then `./metal/<name>.metal`, both relative to the working directory with no fallback to the executable's own location, and running `tests/test_session_state_gpu` from `/tmp` prints `ds4: Metal source metal/flash_attn.metal not found` where the same binary from the repo root passes — so it reproduces without loading a model.

Unrelated to the above, but @cropduster's Qwen3.8 result is worth a follow-up from you rather than me: for Qwen the shutdown store appears to carry the visible key on **both** the PR head and the merge base, so that case is on the lookup side and this patch would not touch it. I do not have a Qwen3.8 model here to look further.

PR #1065, opened 2026-09-16
GLM: render the tool call block inline, as the chat template does

aaa2015:
GLM's `append_glm_tool_calls_text()` wraps the reconstructed tool block in a newline on each side (`"\n" + "<tool_call>..." + "\n"`). `tokenizer.chat_template` does not.

Rendering the template straight from the GGUF (Jinja, `trim_blocks` + `lstrip_blocks`, as HF does) gives the inline form in every variant — empty content, non-empty content, reasoning present, followed by an observation or by a user turn:

```
<|assistant|><think></think><tool_call>get_weather<arg_key>city</arg_key><arg_value>Paris</arg_value></tool_call><|observation|>
```

The model samples that same inline form, so the reconstructed turn disagrees with the bytes it is supposed to reproduce.

## Why it stays hidden inside one process

When the raw sampled bytes are available, `calls->raw_tool_text` is emitted verbatim and both paths agree. Only a history replayed by a client goes through reconstruction, so the two forms only meet across a restart — or on a cold prefill, which renders the replayed history before anything is cached.

## Symptom

With visible-transcript KV checkpoints (the approach #1058 takes), a GLM conversation writes its shutdown checkpoint under `key=thinking-visible`, and the replayed prompt then misses it by exactly three bytes:

```
visiblekey: ...<|assistant|><think></think><tool_call>get_weather<arg_key>city...
reqprompt : ...<|assistant|><think></think>\n<tool_call>get_weather<arg_key>city...
                                    ^^
```

Everything after that point shifts, so the checkpoint can never be a byte prefix of a replay.

## Reproduction

`tests/kv_restart_restore_smoke.sh` with GLM-5.3-Flash-Q2, on top of #1058 + #1059:

```
before: FAIL, restored 8192 of 11992 tokens (3800 re-prefilled)
after:  OK,   restored 11998 of 12012 tokens (14 re-prefilled)
```

The shutdown store becomes the one actually hit:

```
kv cache stored tokens=11998 reason=shutdown key=thinking-visible
kv cache hit text tokens=11998 text=64224 key=thinking-visible
```

## Scope beyond caching

Cold prefill renders through the same path, so GLM tool conversations have been rendered in a shape the model was never trained on since this landed in `005afed`.

## Tests

* `test_glm_tool_calls_render_matches_template` (new): asserts the template-true block, and that the raw and reconstructed renders are byte-identical.
* `test_render_glm_preserves_reasoning_with_tools` had pinned the old bytes; corrected.

`ds4_test --server` passes on current main with this applied.

## Two caveats about the smoke test itself

Unrelated to this change, but they block anyone trying to run it against GLM: `start_server` hardcodes `--prefill-chunk 1024`, which GLM rejects (`--prefill-chunk is not supported for GLM 5.2`), and it does not set a cwd, so the server logs `ds4: Metal source metal/flash_attn.metal not found` with the Metal backend unavailable. Working around both locally is what made the numbers above reproducible.


aaa2015:
**On the two caveats at the end of the description** — I hit both again while reproducing the numbers, so here is the resolution for anyone who wants to point the harness at GLM. The harness itself lives in #1058, not on this branch.

### `--prefill-chunk 1024`

Reproduced verbatim:

```
ds4: --prefill-chunk is not supported for GLM 5.2; GLM uses graph-selected prefill chunks
```

The flag only exists to keep prefill chunks bounded on DeepSeek, so make it optional and leave the default alone:

```diff
+# GLM 5.x rejects --prefill-chunk, so let callers omit it for those models.
+chunk=${DS4_KV_RESTORE_PREFILL_CHUNK-1024}
+if [ -n "$chunk" ]; then chunk_arg="--prefill-chunk $chunk"; else chunk_arg=""; fi
@@
-        --prefill-chunk 1024 --kv-disk-dir "$tmpdir/kv" --kv-disk-space-mb 4096 \
+        $chunk_arg --kv-disk-dir "$tmpdir/kv" --kv-disk-space-mb 4096 \
```

With `DS4_KV_RESTORE_PREFILL_CHUNK=` set empty, GLM-5.3-Flash-Q2 gives

```
kv-restart-restore: OK (restored 11990 of 12004 tokens after restart)
```

— the same 14 re-prefilled as the fixed case above (11998 of 12012). The totals differ by 8 because that tree also carries tools-section fixes, which change the rendered byte length; the re-prefilled count is the number that matters here. The run went through `tool_calls`, a tool result and two further tool-bearing turns.

### The missing cwd

Also real, and it does not need a model to demonstrate. `ds4_gpu_full_source()` resolves each kernel as `metal/<name>.metal` then `./metal/<name>.metal` — relative to the working directory, with no fallback to the executable's own location — so anything that stages the Metal backend must start from the repository root:

```
$ cd /tmp && /path/to/ds4/tests/test_session_state_gpu

$ cd /path/to/ds4 && ./tests/test_session_state_gpu
session state tests: ok
```

In the harness the fix is to launch the server from the repository root, resolving `$bin` to an absolute path first since it defaults to `./ds4-server`. Worth a comment there explaining the requirement, since the failure surfaces as an unrelated-sounding missing-file error.

Neither workaround is needed once those two land, so the sentence about working around them locally can go.

Say the word and I will send both as a patch against #1058, where the file actually lives.

aaa2015:
Closing — this landed upstream as fff391e6c7f2137f3f62eb54162b40acc7d048b7 (cherry-picked by @antirez, 2026-09-16).

Verified equivalent to this PR's head: both touch only `ds4_server.c`, 52 insertions / 4 deletions, and the changed lines are byte-identical.

PR #1067, opened 2026-09-17
Metal V4.1 decode: one command-buffer wait per token, Engram reads on threads

notatestuser:
+45% speedup (26.07 vs 17.94 tok/s) with identical output, V4.1 on M3 Ultra 512

The changes:
  1. Flush every `DS4_METAL_V41_DECODE_FLUSH_LAYERS` layers, wait once at the logits; one Engram upload buffer per table. 17.7 → 22.8 tok/s
  2. Both Engram tables read on reader threads, joined before the first commit. 23.0 → 25.8
  3. Allocation-free pipeline lookup for the plain getter - neutral, cheaper encode (4.0 → 0.7 ms/token)
  4. Small word aligned copies as a compute kernel instead of a blit. neutral, zero blit passes per token.
  5. Same treatment for ds41_graph_step_batch - 239 → 187 ms/step at 8 sessions.

Gated on `ds4_gpu_device_is_pre_m5_apple_silicon()`


notatestuser:
Use `DS4_METAL_V41_DECODE_QUEUE_ALL_DEVICES=1` for M5 Max testing.

PR #1071, opened 2026-09-17
server: decode an open tool call past a stop token

alexph-dev:
A model that samples a stop token inside an open `<tool_call>` fails the request.

Seen with Qwen3.8-Flash-Next 125B-A6B Q2: after 351 generated tokens the model emitted `<|im_end|>` while the block was still open.

    ds4-server: incomplete DeepSeek tool call: stop=stop token token=248046 generated=351 limit=32768 room=220010
    ds4-server: chat ... DSML_START finish=error error="unterminated tool call (stop token; token=248046, generated=351, limit=32768)"

`try_repair_dsml()` only restores a missing outer wrapper, and the tool-error continuation is skipped for streaming clients, so a streaming client loses the turn and has to retry the whole request.

A tool block carries no executable call until it closes, so a stop token there is invalid output rather than the end of the turn. Decoding already constrains that region (greedy outside payload sampling); this keeps decoding with the best non-stop token instead of accepting the stop, using the existing `ds4_session_argmax_ignoring_eos()`. Drafted stop tokens drop through the existing rewind path so the next sample sees fresh logits.

The override is bounded by `DS4_TOOL_CALL_STOP_SUPPRESS_LIMIT` (24) so a model that insists on ending inside the block still terminates into the existing error path instead of spending the whole output budget. When a block never closes, the warning now carries the tail of the generation, which is what was missing while diagnosing this.

Testing: macOS 26.5.2 on an M5 Max (128 GB), Metal backend, Qwen3.8-Flash-Next 125B-A6B Q2.

    make ds4_test && ./ds4_test --server   # ok
    make ds4-server                        # clean, no new warnings

Live checks against the running server: a Responses tool call, a Responses tool-output continuation, an OpenAI chat stream and an Anthropic stream all behave as before, and a Codex session completes a full reasoning -> tool call -> answer turn.

This is independent of the two neighbouring server issues already open: #899 (classify generated text by the prompt thinking state) and #774 (return calls to custom tools as custom_tool_call). I hit all three on the same run; only this one looked unreported.

Happy to add a unit-level regression if you want one: the branch is decode-loop only, so covering it needs an engine-level harness rather than a `--server` case.


ds4_server.c
```diff
@@ -682,6 +682,12 @@ typedef enum {
 #define DS41_PARAM_START "<｜DSML｜ parameter"
 #define DS41_PARAM_END "</｜DSML｜ parameter>"
 
+/* A stop token inside an open tool call is invalid output: the block has to
+ * close before the turn can end. Decoding overrides it, but the override is
+ * bounded so a model that keeps insisting on ending there still terminates
+ * instead of spending the whole output budget. */
+#define DS4_TOOL_CALL_STOP_SUPPRESS_LIMIT 24
```
emilianbold, line 689: The comment here explains the problem but not this constant itself.

ds4_server.c
```diff
@@ -14129,10 +14169,15 @@ static void generate_job_inner(server *s, server_slot *slot, job *j) {
...
+        const size_t tail_off = text.len > tail_keep ? text.len - tail_keep : 0;
         server_log(DS4_LOG_WARNING,
-                   "ds4-server: incomplete %s tool call: stop=%s token=%d generated=%d limit=%d room=%d",
+                   "ds4-server: incomplete %s tool call: stop=%s token=%d generated=%d limit=%d room=%d suppressed=%d tail: %.*s",
                    j->req.model_syntax == SERVER_MODEL_SYNTAX_GLM ? "GLM" : "DeepSeek",
-                   stop_detail, stop_token, completion, max_tokens, room);
+                   stop_detail, stop_token, completion, max_tokens, room,
+                   tool_stop_suppressed,
+                   (int)(text.ptr ? text.len - tail_off : 0),
+                   text.ptr ? text.ptr + tail_off : "");
```
emilianbold, line 14180: This seems useful for debugging but I'm not sure we log context text like this.

ds4_server.c
```diff
@@ -13820,10 +13831,31 @@ static void generate_job_inner(server *s, server_slot *slot, job *j) {
         if (ds4_token_is_stop_for_think_mode(s->engine,
                                              token,
                                              j->req.think_mode)) {
-            finish = "stop";
-            stop_detail = "stop token";
-            stop_token = token;
-            break;
+            int forced = -1;
```
emilianbold, line 13834: Nit: so this is a `forced_token` or some `replacement`? (Just talking about the variable name).

ds4_server.c
```diff
@@ -13820,10 +13831,31 @@ static void generate_job_inner(server *s, server_slot *slot, job *j) {
...
+            if (open_tool_call &&
+                tool_stop_suppressed < DS4_TOOL_CALL_STOP_SUPPRESS_LIMIT) {
+                forced = ds4_session_argmax_ignoring_eos(slot->session,
+                                                         j->req.think_mode);
+            }
+            if (forced < 0) {
+                finish = "stop";
+                stop_detail = "stop token";
+                stop_token = token;
+                break;
```
emilianbold, line 13844: Nit: I know this `break` is from the old code but the new code is quite hard to read. It's 3 `if`s and one of them breaks too! Might as well put what follows into an `else`.

emilianbold (commented):
LGTM.

PS: Just my feedback, I can't merges PRs.

PR #1072, opened 2026-09-18
Let V4.1 single-token decode use the concurrent Engram reader

jjakemaness:
`ds4_engram_read_batch()` already fans row reads across `ENGRAM_READERS` workers — its comment says *"Fixed concurrency hides random-read latency without caching the table"* — but only above a 256-row threshold.

`ds41_graph_step()` reads `DS4_ENGRAM_COLS` rows per table (48 for the two tables) through the serial `ds4_engram_read()`, so **ordinary decode never reaches that path**. Those 48 preads run in series, and they run before `ds4_gpu_begin_commands()`, so the GPU is idle for all of them. Only bulk prefill ever clears 256.

This routes the decode read through the batched entry point and lowers the threshold to 8. Same rows, same order, same values — only the fetch is concurrent.

### Measured

Mac Studio M3 Ultra (80 GPU cores, 256 GB, macOS 26.6.2), DeepSeek-V4.1-Flash-Q2, resident (no `--ssd-streaming`), `--ctx 131072`, temperature 0, 200 generated tokens, this branch vs its merge-base:

| build | runs | median |
| --- | --- | ---: |
| `main` | 19.18, 19.14 | **19.16 tok/s** |
| `main` + patch | 20.42, 21.00, 21.11, 21.18 | **21.06 tok/s** |

**+9.9%**, about 4.7 ms/token. Generated text was **byte-identical** between the two builds. (The 20.42 was a cold first run after rebuild.)

The cost is storage *latency*, not bandwidth — a standalone pread benchmark against the same GGUF, 48 random 264-byte rows, `F_NOCACHE`, readahead disabled, 150 batches:

```
serial    5.177 ms/batch  (107.9 us/row)
parallel  0.614 ms/batch  ( 12.8 us/row)
```

The saving is roughly constant per token, so it is worth more on a faster tree: on @ivanfioravanti's `ds41f-optimizations` branch, where I originally profiled this, the same change measured **26.8 → 30.5 tok/s** (Engram wall 5.55 → 1.12 ms/token) across three alternating pairs, also byte-identical.

### Credit

- @ivanfioravanti — his V4.1 optimization work is where this was profiled. His `ds41f-dspark` branch already carries a better version of this fix, `ds41_engram_parallel()`, reading both tables in one dispatch — but gated to speculative sessions, so ordinary decode falls through to the serial loop there too. Ungating that is probably the better long-term shape of this change, and I'd defer to it.
- The oMLX / drowzeys V4.1-on-Mac work, which is what pointed me at Engram as a decode bottleneck at all. Their approach was a RAM cache of hot rows; measured row reuse here was only 22.5% over a 600-token coding generation, worth roughly +1 tok/s — so the win turned out to be concurrency rather than caching. Wrong idea, right place to look.

I've only tested Metal on M3 Ultra. The change is platform-neutral, but the non-Apple path in `read_batch_part` deserves a look from someone with that hardware.


jjakemaness:
Closing in favor of #1073 — it includes this same fix (single-token decode routed through the concurrent Engram reader, with the threshold lowered to 2) plus asynchronous prefetch of both tables, which is the better version of this change. Thanks @kernelpool.

PR #1074, opened 2026-09-18
Enhance reasoning effort parsing logic

AlessandroChecco:
With this fix the server with respond to e.g.
"reasoning_effort": "50"

so it becomes compatible with pi.dev that doens't accept numerical value (and cannot send think_level by default)


emilianbold:
This PR was cherry picked in main https://github.com/antirez/ds4/commit/ecf7a93148dba25a9fe951f834c1f20735f2b5b1

PR #1082, opened 2026-09-18
CUDA SSD streaming: run the cached experts while the miss reads are in flight (+6 to +12 % decode)

triplesparkle:
**The limitation.** With `--ssd-streaming`, a decode token's six routed experts split into hits
(already in the GPU expert cache) and misses (read from the SSD, about 9.5 MB each at Q2). Today
the layer waits for every miss read to land before it runs any expert, so each token pays for
its slowest read while the GPU sits idle.

**The change.**
- Run the hits' gate/up/down as soon as the routing is known, while the miss reads are in flight.
- Run the misses when they land, then sum the six down partials.
- A threaded pread pool ships with it, because the miss reads must leave the calling thread for
  the GPU to run anything: 8 workers, pinned staging and a CUDA stream each, gate and up of every
  miss read before the downs, no worker ever touches the O_DIRECT globals.

**Output is byte-identical.**
- The partials are summed in slot order from +0.0f, the float sequence the fused kernel executes.
- Greedy `--temp 0 --dump-logprobs` on a 45-token and a 4,392-token prompt gives the same sha256
  as the unpatched tree.

**Switches.**
- `DS4_CUDA_HITS_FIRST=0` keeps the pool and restores the read-then-launch order.
- `DS4_CUDA_STREAMING_EXPERT_PREAD_POOL=0` as well restores the serial staged reads the tree has today.

**Measured.** DGX Spark GB10, DeepSeek V4.1 Flash Q2, `ds4-bench --cuda --ssd-streaming
--ssd-streaming-cache-experts 90GB`, ctx 4096 and 6144, 128 generated tokens, decode tokens/s at
the slower context, each patched run interleaved between two unpatched runs, 4 repeats:

    round   patched t/s   unpatched t/s   delta    repeats above the tip
    A         10.37          9.65         +6.5 %   4 of 4
    B         10.34          8.96        +12.3 %   4 of 4
    C         10.29          9.54         +9.1 %   4 of 4

- Prefill is unchanged within noise.
- Identical binaries swing about 8 % run to run on this box (the model is streamed, not
  page-cached), which is why every number is a median of interleaved repeats.

**Where it does not help.**
- A fully resident model (no misses) does nothing different.
- An io_uring read pool on its own was faster on a quiet box and less stable under CPU load, so
  this PR carries the launch-order change and the plain pthread pool it needs, not the ring.

**Files.** `ds4_cuda.cu` (the pread pool, the split down kernels, the hits-first launch in
`routed_moe_launch`, the task build in the expert cache begin-load path) and a README card with
the numbers and the branches tried alongside. No new dependencies.


triplesparkle:
Opened from the wrong branch by mistake; the intended PR follows from triple-hitsfirst.

PR #1084, opened 2026-09-18
Stream assistant content incrementally when tools are enabled

zonebond:
## Problem

With `stream: true` and `tools` present (`tool_choice` other than `"none"`),
the assistant `content` is emitted as a **single** SSE chunk at the end of
generation. `reasoning_content` is unaffected and still streams token by token.
Live-rendering clients therefore show reasoning streaming and then the whole
answer appearing at once.

Reproduction (`deepseek-v4.1-flash`, `max_tokens: 150`, one short prompt):

| request | reasoning chunks | content chunks (before) |
|---|---|---|
| no `tools` | 84 | 20 (streams) |
| `tools` only (default auto) | 21 | **1** |
| `tools` + `tool_choice: "none"` | 85 | 26 (streams) |
| `tools` + `tool_choice: "auto"` | 27 | **1** |

## Root cause

`fe2d3b0` ("Keep repeated reasoning out of streamed answers") added
`guard_second_reasoning`. After the first `</think>` the streamer enters
`TEXT` mode but keeps the guard armed, and:

```c
} else if (!tool && !final) {
    return true;   /* hold everything until generation ends */
}
```

holds *all* content until a tool call or a second `</think>` appears. In the
common case neither appears, so nothing is flushed until `final`.
`tool_choice: "none"` sets `has_tools = false` and bypasses the guard, which
is why it streams.

## Fix

Keep the reroute intent but bound the hold to a short lookahead window
(`SECOND_REASONING_GUARD_BYTES = 32`): a short stray draft, or a `</think>`
split across generated tokens, is still rerouted to `reasoning_content`; past
the window the text is treated as the answer and streamed. Applied to both
`openai_sse_stream_update` and `anthropic_sse_stream_update`.

Note this matches the non-streaming behaviour, where `split_reasoning_content`
splits on the **first** `</think>`; previously the streaming guard made the two
paths disagree.

## Verification

Machine: Apple M3 Ultra, 512 GB, Metal. Model: V4.1 Flash Q4.

```sh
make clean && make
./ds4_test --server
```

Both pass. A focused regression test,
`test_openai_tools_stream_content_is_incremental`, was added. It fails on the
previous code:

```
ds4_server.c:16801: assertion failed: content_deltas >= 2
```

Live end-to-end, same request as above (content chunks after the fix):

| request | content chunks (after) |
|---|---|
| no tools | 29 |
| tools + `none` | 27 |
| tools + `auto` | 10 |

## Tradeoff

A repeated reasoning pass longer than the window would now leak into
`content`. That is already the case on the non-streaming path, so this makes
the two consistent. If a larger window is preferred, the constant is in one
place.


Flor1an-B:
Independently verified this fix live against our own production DeepSeek V4 Flash and GLM 5.3 Flash servers (real checkpoints, not synthetic) — content chunks went from 1 to 101 (DeepSeek) and 1 to 137 (GLM) on OpenAI-compatible streaming, plus 130 incremental chunks on the Anthropic API. This is the right fix, thank you.

Opened #1099 on top of this with full credit — it adds the Anthropic-path regression test your own test suite was still missing (only the OpenAI path was covered), plus a confirmation that /v1/responses is a separate code path unaffected by this bug.

PR #1086, opened 2026-09-18
server: in-process Prometheus metrics via GET /metrics

jyatesdotdev:
Adds a `/metrics` endpoint to ds4-server with the same series the journald-parsing exporter historically exposed, so existing dashboards keep working when the scrape target moves in-process.

This is the focused successor to #861's metrics portion, rebuilt to minimize merge-conflict surface:

- **All metrics state and exposition rendering live in new files** (`ds4_metrics.c` / `ds4_metrics.h`) that upstream never needs to modify.
- `ds4_server.c` carries only one-line hooks at existing log/lifecycle sites: decode progress (tokens by phase + rate gauges), the prefill progress callback (per-chunk token accounting, so long prompts never render as `rate()` spikes), prompt-done (remainder + cache-hit source), live-KV miss reason, request finish (+duration histograms), job begin/end for `ds4_active_requests`, and the route itself.
- Server-state gauges (`ds4_context_limit_tokens`, `ds4_context_tokens{slot}`) are passed into the renderer by value — the module never includes server types.
- `ds4_kvstore` gains only a five-line optional completion-hook typedef + setter, registered at startup for the stored/trimmed/bytes counters.
- Series kept identical: token counters, prefill/decode tps, requests by finish, request/prompt duration histograms, reused tokens (plus by-source), KV hit/miss by source/reason, KV stored/trimmed/bytes, active requests, context gauges. 861-specific machinery series (rewind refusals, capture skips, dispatch holds) are intentionally not carried.
- Observation only: no engine, scheduler, or KV behavior changes.

Tests: `tests/test_metrics` (wired into `make test`) is model-free and asserts exact exposition lines, histogram bucket placement, and the never-negative active gauge. Verified on macOS (Metal): `make`, `make cpu`, `ds4_test --server`, `tests/test_metrics` all pass. A live ROCm scrape from the production deployment follows as a comment.


jyatesdotdev:
Live ROCm evidence: the identical patch cherry-picked onto the production V4.1 (PR #1036-era) tree, deployed on the two gfx1151 Strix Halo nodes, serving one 95-token prompt + 256-token completion:

```
$ curl -s http://max2:8080/metrics | grep -E '^ds4_(prompt|generated|requests|active|decode_tps|prefill_tps|request_duration|context)'
ds4_prompt_tokens_total 95
ds4_generated_tokens_total{phase="thinking"} 256
ds4_generated_tokens_total{phase="response"} 0
ds4_prefill_tps{window="chunk"} 0.00
ds4_prefill_tps{window="avg"} 44.55
ds4_decode_tps{window="chunk"} 13.89
ds4_decode_tps{window="avg"} 14.03
ds4_requests_total{finish="length"} 1
ds4_request_duration_seconds_sum 20.388
ds4_request_duration_seconds_count 1
ds4_active_requests 0
ds4_context_limit_tokens 393216
ds4_context_tokens{slot="0"} 351
```

Request wall time 20.40s vs histogram sum 20.388s; slot residency 351 = 95 prompt + 256 generated. (chunk prefill rate is 0.00 here because a 95-token prompt completes in a single chunk — the chunk gauge only gets a rate from the second callback onward; the avg gauge is correct.)

jyatesdotdev:
@kyuz0 here is the branch focused on the metrics we discussed. Thanks!

PR #1088, opened 2026-09-18
server: recover unterminated reasoning as content when the turn completed

### LINKED ISSUE #1087: server: content is emptied when a thinking model ends a turn without closing </think> (finish=stop, no tools involved)
## Summary

With thinking mode enabled, when the sampled turn ends with `finish_reason: "stop"`
**without ever emitting `</think>`** and **no tool call is involved**, `ds4-server`
routes the whole buffer to `reasoning_content` and returns `content: ""`.

Two related behaviours are deliberate and stay as they are:

- **#509 (truncation, `finish_reason: "length"`)** — a cut-off reasoning chain must
  not reach clients looking like a finished reply. That is exactly what the
  comment above `ds4_local_unterminated_reasoning()` describes, and it should
  remain.
- **#318 (tool call inside unclosed thinking)** — recovery landed as a forced
  `</think>` so the tool call is delivered ("display whatever the model sampled").

Neither covers the plain case: no `tools` in the request, generation completed
normally, and the model's output is dropped.

## Measured impact

Two thinking models on the same server, same request shape, 50 scored reps each
(long-form analytical prompts, no tools, `reasoning_effort: high`):

| model | reps returning `content: ""` | `reasoning_content` in those reps |
|---|---|---|
| GLM-5.3-Flash-Q2 (GGUF, resident) | 4 / 50 | 3.060 – 18.827 characters |
| Qwen3.8-Flash-Next-Q4 (GGUF) | 5 / 50 | 20.015 – 45.734 characters |

All nine have `finish_reason: "stop"`. The buffer is not a cut-off chain — it is
whatever the model chose to end with, **and it contains the finished answer**.
From the four GLM reps, the first 400 characters of the dumped
`reasoning_content` (the benchmark task asks for a stage-1 base-rate
calculation with a fixed expected solution):

```
rep1: Base rate 2% fraudulent, 200 per 10000. Stage 1: sens 90% → 180 true alarms;
      FPR 5% → 0.05 × 9800 = 490 false alarms. Total alarms = 670.
      PPV = 180/670 = 0.26865... = 26.9%. Alarme per 10000 = 670. Echte Alarme = 180.
rep2: - True alarms: 0.9 × 200 = 180. - False alarms: 0.05 × 9800 = 490.
      - Total alarms: 670. - PPV = 180/670 = 0.26865... = 26.9%. alarme_je_10000 = 670.
rep6: PPV Stufe 1: Sens 0.9 → TP = 0.9*200 = 180. FP = 0.05*9800 = 490.
      PPV = 180/670 = 0.26865... = 26.9% (1 decimal).
rep8: Stufe 1: sensitivity 90% → detects 180 fraud. FPR 5% → alarms 490 legit.
      Total alarms = 670. PPV = 180/670 = 0.26866... = 26.9%.
```

Those are the correct expected values (670 / 26.9% / 180), computed correctly,
in all four reps that the server returned as empty. The work is not missing; it
is discarded.

Reproduced again in a controlled rerun (same prompt, same server, thinking
mode, `reasoning_effort: high`) — 1 hit in 23 reps, and that rep's dumped buffer
is 13,412 characters long. Its first 300 characters already carry the enumerated
solution:

```
Let me compute carefully. Base: 2% fraudulent, 98% legitimate per 10,000
transactions: 200 fraud, 9800 legit. Stage 1: sensitivity 90%, FPR 5%.
- True alarms: 0.9 × 200 = 180
- False alarms: 0.05 × 9800 = 490
- Total alarms: 670 → alarme_je_10000 = 670
- PPV = 180/670 = 0.26865... = 26.9% (1 decimal: 26.9)
- echte_alarme = 180
1. ppv_prozent = 26.9
2. alarme_je_10000 = 670
3. echte_alarme_je_10000 = 180
4. max_fpr: PPV ≥ 90%: 180/(180 + fpr×9800) ≥ 0.9 → ...
```

All five expected values of that task (26.9 / 670 / 29 / 15,025.00 / 78.6) are
present in the buffer. In this rep the model was still weighing two readings of
the last field when it emitted EOS, so the buffer is not always a finished
answer — but it never is nothing, and "show what the model sampled" is the
behaviour the rest of the server follows.

Note on reproducibility: intermittent. Over the original 50-rep runs it hit 9
reps (~10%); a first 12-rep rerun produced no hit, a 23-rep rerun produced one.
Every hit is accompanied by exactly one `thinking not closed` log line.

The only diagnostic on the server side is:

```
ds4-server: thinking not closed, ignoring incomplete GLM tool calls in reasoning
```

The server log line matches the affected reps 1:1 — same timestamps, same `gen=`
token counts, and the next line shows the buffer being emptied:

```
0918 19:42:58 ds4-server: chat ctx=6993..7043:50 gen=5600 THINKING decoding chunk=24.18 t/s avg=22.82 t/s 245.448s
0918 19:43:00 ds4-server: chat ctx=7043..7082:39 gen=5639 THINKING decoding chunk=19.50 t/s avg=22.79 t/s 247.449s
ds4-server: thinking not closed, ignoring incomplete GLM tool calls in reasoning
0918 19:43:00 ds4-server: chat ctx=1443..1443:0 gen=5639 THINKING finish=stop 247.449s
```

For a benchmarking or batch client these reps are indistinguishable from "the
model produced nothing" — they count as total failures, which is a ~10% silent
loss on long prompts with no way for the client to tell it apart from a genuine
empty answer.

## A/B confirmation (same server, same prompt, both binaries)

30 reps each, KV cache cleared before each run, identical request parameters:

| | original binary | with the change below |
|---|---|---|
| `thinking not closed` log lines | 1 | 1 |
| reps with `content: ""` | **1** (`unterminated_reasoning`, scored 0 %) | **0** |
| scored reps | 29 / 30 | 30 / 30 |
| reliability | 96.7 % | **100 %** |
| what the client received | `content: ""` | 3,411 characters of model text |

The rep rescued by the change carries the model's own buffer (identical to
`reasoning_content`) instead of an empty string. It holds the correct
intermediate values of this task — 180 true alarms, 670 total alarms,
PPV 26.9%, max FPR 0.204% — but as prose, not as the JSON the task asks for, so
a strict-schema client still scores it 0.

The same server path fired in both runs — once as a total loss, once delivering
the text. Making the model close `</think>`, and thus produce a properly
formatted answer, is a separate (model/template-side) problem.

*Correction:* an earlier version of this post claimed the rescued rep came back
as the expected JSON with every value correct. That was a mis-attribution on my
side — the JSON belonged to a normal rep from the same run (its `reasoning`
was 6,510 characters ending in a proper `</think>`), not to the rep the server
emptied. The log correlation is unambiguous: the affected rep is the one whose
`gen=` count and duration match the `thinking not closed` line (here
`gen=1504`, 57.5 s).

## Code path

`parse_glm_generated_message_ex()` (`ds4_server.c:6341-6350`, same shape in the
Qwen and DeepSeek variants at `:6545` and `:6131`): no `</think>` and no complete
`<tool_call>...</tool_call>` → `ds4_local_unterminated_reasoning()` (`:6092`),
which sets `content_out = ""` and moves everything into `reasoning_out`. The
decision there cannot see `finish_reason`; the truncation argument in its comment
is the `length` case.

## Suggested fix

`parse_generated_message_for_response_for_syntax()` already has `finish_io` in
scope. After a successful parse, if `content` is empty, `reasoning_content` is
non-empty and `finish != "length"`, re-emit the text as `content`:

```c
const char *finish_final = (finish_io && *finish_io) ? *finish_io : "stop";
if (strcmp(finish_final, "length") != 0
        && content_out && *content_out && !**content_out
        && reasoning_out && *reasoning_out && **reasoning_out) {
    free(*content_out);
    *content_out = xstrdup(*reasoning_out);
}
```

This keeps #509 semantics untouched (finite: `length` still returns empty
content), leaves the #318 forced-`</think>` path alone (it never produces an empty
content), keeps `reasoning_content` populated for diagnostics, and sits in the
shared wrapper so all three syntax parsers benefit from one change.

Branch with the change and four server unit tests (recovery on `stop`, unchanged
behaviour on `length`, Qwen syntax, closed-`<think>` regression guard — `ds4
tests: ok`):
https://github.com/cropduster/ds4/tree/fix/unterminated-reasoning-stop

Happy to send it as a PR if the semantics look right to you — or to adjust: if
an unterminated buffer should stay out of `content` even on `stop`, the
alternative would be an explicit marker the client can detect instead of a
silent empty string.

cropduster:
## What this fixes

With thinking mode on, a turn that ends normally (`finish_reason: "stop"`) without
ever emitting `</think>` had its entire buffer routed to `reasoning_content` with
`content` set to `""`. For a truncated turn that behaviour is intended (#509) and
for a tool call inside unclosed thinking #318 added the forced-`</think>` recovery
— but a plain completed turn with no tools in the request lost its answer
entirely.

## Evidence

Same server, two thinking models, 50 scored reps each, long-form analytical
prompts, `reasoning_effort: high`, no tools:

| model | reps with `content: ""` | `reasoning_content` in those reps |
|---|---|---|
| GLM-5.3-Flash-Q2 | 4 / 50 | 3.060 – 18.827 characters |
| Qwen3.8-Flash-Next-Q4 | 5 / 50 | 20.015 – 45.734 characters |

All nine report `finish_reason: "stop"`. The log line matches the affected reps
1:1 (timestamps and `gen=` counts), and the following line shows the buffer being
emptied by `ds4_local_unterminated_reasoning()`:

```
0918 19:42:58 ds4-server: chat ctx=6993..7043:50 gen=5600 THINKING decoding chunk=24.18 t/s avg=22.82 t/s 245.448s
0918 19:43:00 ds4-server: chat ctx=7043..7082:39 gen=5639 THINKING decoding chunk=19.50 t/s avg=22.79 t/s 247.449s
ds4-server: thinking not closed, ignoring incomplete GLM tool calls in reasoning
0918 19:43:00 ds4-server: chat ctx=1443..1443:0 gen=5639 THINKING finish=stop 247.449s
```

The discarded text contains the correct answer. Those four GLM reps were given a
base-rate task with a fixed expected solution, and the first 400 characters of
each discarded buffer already hold it:

```
rep6: PPV Stufe 1: Sens 0.9 → TP = 0.9*200 = 180. FP = 0.05*9800 = 490.
      PPV = 180/670 = 0.26865... = 26.9% (1 decimal).
```

`670` alarms, `PPV 26.9%`, `180` true alarms — correct in all four reps the
client received as empty. The effect is intermittent (~10% over the measured
runs; a 12-rep rerun of the same task produced no hit, so it is not a
deterministic repro).

(An earlier version of this description claimed the affected rep came back as
the expected JSON. That was a mis-attribution on my side: the JSON belonged to a
normal rep from the same run. See the A/B section below for what the rescued
buffer actually contains.)

## A/B confirmation

Same server, same prompt, KV cache cleared before each run, 30 reps each:

| | original binary | this branch |
|---|---|---|
| `thinking not closed` log lines | 1 | 1 |
| reps returned as `content: ""` | **1** (scored 0 %) | **0** |
| scored reps | 29 / 30 | 30 / 30 |
| reliability | 96.7 % | **100 %** |

What the rescued rep actually contains: 3,411 characters of the model's own
buffer (the reasoning text), instead of the empty string the original binary
returned. Those characters hold the correct intermediate values of this task —
`180` true alarms, `670` total alarms, `PPV 26.9%`, `max FPR 0.204%` — but as
prose, not as the JSON the task asks for:

```
Compute. Per 10,000 transactions: 200 fraudulent, 9800 legitimate.
Stufe 1: sensitivity 90% → 180 true alarms. FPR 5% → 490 false alarms.
Total alarms = 670. PPV = 180/670 = 0.26866 → 26.9%.
alarme_je_10000 = 670. echte_alarme = 180.
max_fpr: PPV ≥ 90%: ... f ≤ 0.002040816 → 0.2040816% → rounds to 0.204.
```

So a client with a strict response schema still scores that rep 0 — what the
change guarantees is that the buffer is no longer discarded, which is the
difference between "the model produced this" and "the model produced nothing".
Whether the buffer is a finished answer depends on where the model stopped;
making the model close `</think>` is the model/template side of the problem and
is not what this change does.

Fixes #1087.

## The change

Recovery happens in `parse_generated_message_for_response_for_syntax()`, which
already has `finish_io` in scope, so no parser signature changes and all three
syntax parsers benefit:

```c
const char *finish_final = (finish_io && *finish_io) ? *finish_io : "stop";
if (strcmp(finish_final, "length") != 0
        && content_out && *content_out && !**content_out
        && reasoning_out && *reasoning_out && **reasoning_out) {
    free(*content_out);
    *content_out = xstrdup(*reasoning_out);
}
```

- `finish_reason: "length"` keeps the #509 behaviour untouched (content stays
  empty, so a cut-off chain never looks like a finished reply).
- The #318 forced-`</think>` path never produces an empty content, so it is
  unaffected.
- `reasoning_content` stays populated, so the diagnostic information is not lost.
- One change site instead of three parser branches.

## Tests

Four new server unit tests in `ds4_server.c`:

- `test_unterminated_reasoning_content_recovered_on_stop` — the recovery case
  (content filled, `<think>` stripped, `finish_reason` untouched)
- `test_unterminated_reasoning_stays_empty_on_length` — regression guard for the
  #509 semantics
- `test_unterminated_reasoning_recovered_on_qwen_syntax` — same rule for the Qwen
  parser branch
- `test_closed_thinking_split_is_unchanged` — a properly closed block still
  splits, with no duplication into `reasoning_content`

```
$ make ds4_test ds4-server -j8
$ ./ds4_test --server
server: OK
ds4 tests: ok
```

Happy to restructure this if you would rather see the recovery localised in
`ds4_local_unterminated_reasoning()` (it would need `finish_reason` threaded
through the parser signatures), or if the semantics should be gated behind a flag.


cropduster:
Rebased onto current `main` (8db1d1d) so the diff is just the one file (+113/-1). Rebuilt and re-ran the server tests on that base:

```
$ make ds4_test ds4-server -j8
$ ./ds4_test --server
server: OK
ds4 tests: ok
```

All four added tests run through the affected path (the `thinking not closed` lines above `server: OK` are the new cases). The A/B measurement in the description was taken on a fork build; the change itself and the tests are unchanged by the rebase.

cropduster:
Same correction as filed on #1087: the description earlier claimed the rescued rep returned the expected JSON. That was my mis-attribution — the JSON came from a normal rep of the same run. The rescued rep (`gen=1504`, 57.5 s, log-correlated) returns 3,411 characters of the model's reasoning buffer: correct intermediate values as prose, not as the task's JSON. PR description corrected. The change still does exactly one thing: it stops discarding that buffer.

Issue #1092, opened 2026-09-19
GLM tool turns clear the live checkpoint, so every follow-up turn rebuilds its prefix


cropduster:
Hi Salvatore,

while running a long agent session against `glm5_next` (GLM-5.3-Flash, Q2 resident, ctx 393216, MTP on)
I noticed that every follow-up turn after a tool call rebuilds its prompt prefix from an older disk
block instead of continuing the live KV. On this machine that is ~30k tokens of prefill (~100 s) per
follow-up turn.

## What the log shows (one server, one client session)

```
12:32:48 chat ctx=50681..50747:66  gen=92  TOOLS DSML_START DSML_END finish=tool_calls 7.054s
12:57:50 GLM live prefix rewind from 50839 to 50746 requires rebuild
12:57:50 kv cache skipped tokens=50746 reason=evict because KV payload staging failed: session has no valid checkpoint to stage
12:57:51 kv cache hit text tokens=20536 ... load=563.0 ms
12:57:51 chat ctx=20536..50747:30211 TOOLS prompt start          <- 30211 tokens re-prefilled
```

Three observations:

1. The follow-up request is 92 tokens *shorter* than the live frontier (50839 -> 50747): the client
   (Hermes desktop) does not send the generated tool-call text back as reasoning. That is normal for
   an OpenAI-style chat client — it is exactly the case the Qwen path was built for
   (`build_qwen_tool_turn_visible_text`, "Match clients that omit reasoning").
2. The rewind cannot rescue it: `ds4_session_glm_mtp_rewind()` only covers the two-token MTP cycle, so
   `ds4_session_rewind()` finishes with `state_ok == false`, clears `checkpoint_valid` and the caller
   rebuilds. That looks correct for the compressed DSA KV — truncation is not a valid rollback.
3. Therefore the only thing that can keep the live prefix usable is a *visible* checkpoint — and for
   GLM it is never set.

## Root cause

`should_remember_thinking_checkpoint()`:

```c
const bool qwen_chat = r->model_syntax == SERVER_MODEL_SYNTAX_QWEN &&
                       r->api != API_RESPONSES && r->api != API_ANTHROPIC;
if ((r->has_tools || r->prompt_preserves_reasoning) && !qwen_chat) return false;
```

Qwen is exempted here, GLM is not. At the call site in the tool-call branch:

```c
} else if (parsed_calls.len) {
[...]

PR #1093, opened 2026-09-19
server: keep the GLM live checkpoint across tool turns

cropduster:
Keeps the GLM live checkpoint across tool turns — addresses #1092.

## Problem

`should_remember_thinking_checkpoint()` rejects `has_tools` for every non-Qwen syntax, and the
tool-call branch then calls `thinking_live_clear()`. For a GLM chat request with tools and known raw
DSML, `should_canonicalize_tool_checkpoint()` also returns false (exact DSML replay preferred), so no
checkpoint is written on any path — the live checkpoint is cleared instead. The next request then logs

```
live kv cache miss live=8471 prompt=6662 common=6041 vision=match reason=token-mismatch
kv cache hit text tokens=5897 ...              <- falls back to an older disk block
chat ctx=5897..6662:765 TOOLS prompt start     <- 765 tokens re-prefilled
```

`common` stops at the prompt end (6041): the generated reasoning block lives in the live KV but not in
the follow-up prompt, so there is no plain token match, and the compressed DSA KV cannot be rolled
back by truncation (`ds4_session_glm_mtp_rewind()` only covers the two-token MTP cycle).

## Change

Mirror the Qwen path for GLM: remember the visible key (prompt + closed think block + tool-call text)
through the existing `build_thinking_visible_text()` GLM branch, gated like the Qwen helper (chat
completions, GLM syntax, not Responses/Anthropic, decode `finish == "tool_calls"`, not inside
thinking, non-empty tool text, prompt still ending with the open `<think>` marker). The next request
renders the same visible surface, `thinking_live_visible_prefix_prompt()` matches it and the live KV
is reused instead of falling back to an older block.

The tool-call text comes from `parsed_calls.raw_tool_text`: GLM consumes the DSML call out of the
generated content, so `parsed_content` is **empty** on a tool turn (verified with a temporary
diagnostic: `finish=tool_calls inside_thinking=0 content_len=0` before the correction,
`content_len=721` after).

## Measurement

Same probe, same host (M4 Max, GLM-5.3-Flash Q2, ctx 393216, MTP on), server restarted with the KV
cache cleared before each arm. Request 1 is a tool-schema turn that generates a reasoning block;
request 2 replays the turn without the reasoning plus a tool result.

| | without the patch | with the patch |
|---|---|---|
| after the tool turn | `live kv cache miss … common=6041` | `thinking live checkpoint remembered … live=6997` |
| follow-up prefill | **765 tokens** (4.2 s), from disk block 5897 | **84 tokens** (1.2 s) |
| cache source | older disk block | live KV (visible prefix) |

The saving scales with the reasoning the client omits: here 681 tokens of prefill. In a production
agent run on this host the same mechanism showed up as a fallback from a 50 747-token frontier to a
20 536-token block — i.e. ~30k tokens of re-prefill (~100 s) — when no block near the frontier
existed. Caveat: the two arms generated different amounts of reasoning (7497 vs 1537 characters), so
this is one sample per arm rather than a controlled n; the mechanism, not the exact factor, is what
the numbers show.

## Tests

Adds `test_glm_tool_turn_checkpoint_eligible()` (positive case + 8 rejections).
`./ds4_test --server` passes.


emilianbold:
This seems related to #1058 which has more work on keys for the KV store. But it is a smaller fix, which is always nice.

I had AI look into this PR and found an issue:

> One regression found (confirmed with a probe test): the new tool_turn_content fallback feeds raw_tool_text into the Qwen helper when parsed_content is empty. build_qwen_assistant_suffix() appends the trimmed content and append_qwen_tool_calls_text() appends the raw tool text again, so the Qwen visible key contains the tool text twice and is no longer a prefix of the next prompt — Qwen pure tool turns (common in agent flows) lose the live-KV hit they had pre-patch and fall back to rebuilds. Existing tests miss it because they call the builder directly, bypassing the call-site substitution. Suggested fix: pass parsed_content ? parsed_content : "" to the Qwen helper and use the raw_tool_text fallback only for the GLM call.

Flor1an-B:
Applied the fix @emilianbold's review suggested and verified it live on both models (real weights, not synthetic).

## Fix

The regression is real: `build_qwen_assistant_suffix()` already appends `calls->raw_tool_text` via `append_qwen_tool_calls_text()`, so feeding the same `raw_tool_text` in as `content` (the new fallback) doubles it for Qwen. GLM has no equivalent built-in tool-call rendering in `remember_thinking_checkpoint()`'s path, so it needs the fallback. Split the two:

```c
const char *qwen_tool_turn_content = parsed_content ? parsed_content : "";
const char *glm_tool_turn_content = parsed_content && parsed_content[0] ?
                                    parsed_content :
                                    (parsed_calls.raw_tool_text ?
                                     parsed_calls.raw_tool_text : "");
bool checkpoint_remembered = remember_qwen_tool_turn_visible_checkpoint(
        s, slot, j, ctx_span, finish, thinking.inside,
        qwen_tool_turn_content,
        &parsed_calls);
if (!checkpoint_remembered &&
    glm_tool_turn_checkpoint_eligible(&j->req, finish, thinking.inside,
                                      glm_tool_turn_content))
{
    remember_thinking_checkpoint(s, slot, j, ctx_span, trace_id,
                                 glm_tool_turn_content, false);
    checkpoint_remembered = true;
}
```

(Qwen keeps exactly its pre-patch input; GLM keeps the new fallback.)

## Verification

Real hardware (M5 Max), real weights, full unit suite green (including the PR's own new `test_glm_tool_turn_checkpoint_eligible`).

**GLM 5.3 Flash Q2** — tool-turn then a follow-up:
```
turn1: finish=tool_calls, cached_tokens=0/151
turn2: finish=stop, cached_tokens=162/183   <- live checkpoint hit, matches the PR's own intent
```
Log: `thinking live checkpoint remembered ctx=0..151:151 live=162 visible=795`

**Qwen3.8-Flash-Next Q2** — pure tool-turn (`content_len=0`, exactly the case the review flagged) then a follow-up:
```
turn1: finish=tool_calls, content_len=0, cached_tokens=0/308
turn2: finish=stop, cached_tokens=356/393   <- fully reused, no more rebuild
```
Log: `qwen tool-turn visible checkpoint remembered ctx=0..308:308 live=356 visible=1539` then
`thinking live checkpoint remembered ctx=356..393:37 live=436 visible=1777`

Before the fix (reproduced): the doubled `raw_tool_text` broke the Qwen visible-key prefix match, so `cached_tokens` was 0 on turn 2 — confirmed the regression is real, and this patch resolves it without touching the GLM behavior at all.

Happy to open this as a small follow-up PR against `main` if that's easier than a patch here — whichever @cropduster prefers.

Issue #1076, opened 2026-09-18
Tool-enabled chat completions buffer content until generation ends


Jiffywu:
## Summary

Tool-enabled chat completions buffer normal answer text until generation finishes. Reasoning streams incrementally, but the complete answer arrives in one final content delta.

This affects Qwen, DeepSeek V4 Flash, and GLM 5.3 Flash through the OpenAI-compatible endpoint used by pi-web.

## Reproduction

Send a thinking-enabled request with tools and `stream: true` to `/v1/chat/completions`.

Observed before the fix:

| Model | Thinking deltas | Text deltas | Text delivery |
|---|---:|---:|---|
| Qwen | not recorded | 1 | complete answer at finish |
| DeepSeek V4 Flash | 163 | 1 | 1,994 characters at finish |
| GLM 5.3 Flash | 132 | 1 | 3,341 characters at finish |

## Cause

`openai_stream_start()` and `anthropic_sse_start_live()` enable `guard_second_reasoning` for every tool-enabled thinking request. The safeguard was added for the escaped second reasoning pass reported in #678, which was measured on Anthropic streaming with legacy DeepSeek formatting.

After the first `</think>`, the guard cannot distinguish a real answer from an unmarked second reasoning pass. It therefore holds all text until another close marker, a tool marker, or the final update appears. For normal generations, that turns a live stream into one final chunk.

## Proposed fix

Limit the guard to the combination it protects: Anthropic streaming with legacy DeepSeek syntax. OpenAI-compatible streams can then emit answer text immediately for every current model family. Anthropic retains the #678 protection for legacy DeepSeek while modern DeepSeek 4.1, GLM, and Qwen stream normally.

Regression coverage should verify incremental answer delivery before finish for:

- OpenAI-compatible DeepSeek, DeepSeek 4.1, GLM, and Qwen streams
- Anthropic DeepSeek 4.1, GLM, and Qwen streams
- Anthropic legacy DeepSeek second-pass rerouting

Live pi-web verification after the fix:

| Model | Thinking deltas | Text deltas | Text stream duration |
|---|---:|---:|---:|
| Qwen | 143 | 1,082 | 18.5s |
| DeepSeek V4 Flash | 248 | 506 | 13.4s |
| GLM 5.3 Flash | 656 | 762 | 26.6s |

Jiffywu:
I ran the same end-to-end pi-web SSE probe against all three installed model families. Each request used the same prompt, pi-web tool set, and high reasoning.

[...]

PR #1099, opened 2026-09-20
server: cover Anthropic + verify Responses API for the tools-streaming guard fix (#1076)

Flor1an-B:
## Context

#1076 reported that a tool-enabled thinking answer buffers entirely until
generation ends. The merged fix (d31089d / 39bc628) only exempted Qwen
from `stream_needs_second_reasoning_guard()`; DeepSeek V4 Flash and GLM
5.3 Flash still hit the original unbounded wait, confirmed live on our
own production servers (OpenAI-compatible streaming, real checkpoints,
tools + thinking on):

| Model | Reasoning chunks | Content chunks |
|---|---:|---:|
| DeepSeek V4 Flash | 85 | 1 |
| GLM 5.3 Flash Q2 | 68 | 1 |

## What this PR does

Applies **@zonebond's fix from #1084** (full credit — the bounded
`SECOND_REASONING_GUARD_BYTES = 32` lookahead window in both
`openai_sse_stream_update` and `anthropic_sse_stream_update`, model-syntax-
agnostic rather than gated per protocol/syntax) and adds the coverage that
PR was still missing:

- `test_anthropic_tools_stream_content_is_incremental`: #1084's own
  regression test only covers the OpenAI-compatible path. The guard fires
  identically in `anthropic_sse_stream_update`, and our manual live testing
  needed the same fix there too — this codifies that as an automated test
  instead of a one-off manual check.

## Verification

- Full unit suite green, including the pre-existing
  `test_openai_stream_reroutes_second_reasoning_pass` (its synthetic
  "escaped draft" is 13 bytes, safely under the 32-byte window — the
  original #678 protection still fires).
- Live, real production models, OpenAI-compatible streaming:
  DeepSeek V4 Flash 1 → **101** content chunks (word-by-word), GLM 5.3
  Flash Q2 1 → **137**. Reasoning and tool-call payloads unaffected.
- Live, real Anthropic-format streaming (`/v1/messages`): DeepSeek, 130
  incremental `text_delta` chunks.
- Checked the third client-facing surface, `/v1/responses`: its
  `responses_stream` struct has no `guard_second_reasoning` field at all,
  and a live request against a real DeepSeek server streamed 128
  incremental `response.output_text.delta` events with no buffering — this
  API is unaffected, no fix needed there.
- Edge case: an immediate tool call with no preceding free text correctly
  yields zero content chunks (nothing to stream).

## Known, deliberate tradeoff (inherited from #1084, stating explicitly)

The 32-byte window protects a *short* stray second-reasoning draft right
after the first `</think>` (the original #678 scenario). A hypothetical
escaped draft longer than 32 bytes would stream as if it were the real
answer. This is a reasonable tradeoff given the historical case it guards
against, but worth stating rather than leaving implicit.

https://claude.ai/code/session_01AwepoHH4uV6KQR6JkFK1sQ


emilianbold (commented):
This makes sense.

I mildly wish the comment would underline it is an heuristic. We do use `guard` already so this new constant `SECOND_REASONING_GUARD_BYTES` does make sense but the technique itself is just an heuristic.

Flor1an-B:
Good point, thank you. Pushed c231bea — spells out that this is a heuristic (not a guaranteed detection), why 32 bytes was picked (covers the #678 escaped-draft case with margin), and what it explicitly doesn't catch (a genuine second-reasoning draft longer than the window would stream as content).

emilianbold (commented):
LGTM.

PS: This is just my review, I don't have merge rights.

